Issue #8, opened 2022-06-17
Implement draft-ietf-ppm-dap-01


cjpatton:
The next DAP draft will be 01. This issue tracks the work that needs to be done for Daphne to be in compliance.

- [x] Upgrade to prio 0.8.0 (implements draft-irtf-cfrg-vdaf-01)
- [x] Remove HMAC from agg flow requests/responses
- [x] Remove agg flow message variants
- [x] Add bearer token to agg flow requests
- [x] Add bearer token to collect flow requests
- [x] Update info/aad strings
- [x] Update media types
- [x] Add unrecognziedAggregationJob abort type (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/285)
- [ ] WON'T FIX (see #44) Make timestamps more granular (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/281)
- [ ] WON'T FIX (see #45) Conservatively constrain collect requests (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/277)
- [x] Increase Nonce.rand from 8 bytes to 16
- [x] Update Janus integration tests

PR #14, opened 2022-06-23
Update agg request serialization

cjpatton:
Partially addresses #8.
Based on #13 (merge that first).

Aligns agg request serialization with draft-ietf-ppm-dap-01. The main
thing that has changed is that the media type is used to distinguish agg
init requests from agg cont requests, rather than a byte encoded in the
message payload itself.

While at it, pick up some clippy fixes.

### DIFF
--- a/daphne/src/messages.rs
+++ b/daphne/src/messages.rs
@@ -300,24 +300,23 @@ impl std::fmt::Display for TransitionFailure {
     }
 }
 
-/// An aggregate response sent from the Helper to the Leader in response to an (initial) aggregate
-/// request. The contents of this structure pertain to a single task and batch.
+/// An aggregate response sent from the Helper to the Leader.
 #[derive(Debug, PartialEq, Default)]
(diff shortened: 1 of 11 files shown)


daphne/src/messages.rs
```diff
@@ -129,106 +129,66 @@ impl Decode for ReportShare {
...
             agg_job_id: Id::decode(bytes)?,
-            var: AggregateReqVar::decode(bytes)?,
+            agg_param: decode_u16_bytes(bytes)?,
+            report_shares: decode_u16_items(&(), bytes)?,
         })
     }
 }
 
-/// Aggregate request variant.
+/// Aggregate initialization request.
```
armfazh, line 161: initialization? continuation?

armfazh (approved):

daphne/src/messages.rs
cjpatton: Yup, nice catch :)

PR #19, opened 2022-06-23
Triggers CI in each pull request

armfazh:
This allows run CI test from all pull request, not only those targeting the main branch.

PR #21, opened 2022-06-23
Enables cache in testing job

armfazh:
Based on #19, merge that first.
Verified, it works. The job reduced from 7m to 2m


cjpatton (changes requested):
Similar comment as in #19: Commit messages should begin with a command (i.e., "Use") and not have a period at the end.

cjpatton (approved):

PR #42, opened 2022-07-13
Ensure all durable object state is persistent

cjpatton:
Based on #41 (merge that first).

Use storage API to store the state of `LeaderStateStore`,
`HelperStateStore`, and `AggregateStore`.

### DIFF
--- a/daphne_worker/src/durable/aggregate_store.rs
+++ b/daphne_worker/src/durable/aggregate_store.rs
@@ -1,7 +1,7 @@
 // Copyright (c) 2022 Cloudflare, Inc. All rights reserved.
 // SPDX-License-Identifier: BSD-3-Clause
 
-use crate::int_err;
+use crate::{durable::state_get_or_default, int_err};
 use daphne::DapAggregateShare;
 use worker::*;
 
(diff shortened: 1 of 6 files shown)


daphne_worker/src/durable/aggregate_store.rs
```diff
@@ -26,36 +24,37 @@ const OK: &str = "Ok";
...
+        Self { state }
     }
 
     async fn fetch(&mut self, mut req: Request) -> Result<Response> {
         match (req.path().as_ref(), req.method()) {
             (DURABLE_AGGREGATE_STORE_DELETE_ALL, Method::Post) => {
-                self.agg_share.reset();
-                Response::ok(OK)
+                self.state.storage().delete_all().await?;
+                Response::empty()
```
armfazh, line 41: here and below, does `::empty` has the same meaning as `::Ok`?

armfazh (approved):

daphne_worker/src/durable/aggregate_store.rs
cjpatton: In both cases the status code is 200. This just changes the payload, which the caller ignores anyway :)

PR #43, opened 2022-07-13
Add ReportStore and HelperStateStore structs

nakatsuka-y:
This PR implements ReportStore and HelperStateStore structs that are necessary for storing report shares and helper state.

### DIFF
--- a/daphne/src/lib.rs
+++ b/daphne/src/lib.rs
@@ -313,7 +313,7 @@ pub struct DapLeaderUncommitted {
 }
 
 /// The Helper's state during the aggregation flow.
-#[derive(Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct DapHelperState {
     pub(crate) seq: Vec<(VdafState, Nonce)>,
 }
--- a/daphne/src/roles.rs
+++ b/daphne/src/roles.rs
@@ -444,6 +444,18 @@ pub trait DapHelper<S>: DapAggregator<S> {
                     &early_rejects,
                 )?;
 
+                // Check that helper state with task_id and agg_job_id does not exist.
+                if self
+                    .get_helper_state(&agg_init_req.task_id, &agg_init_req.agg_job_id)
+                    .await?
+                    .is_some()
+                {
+                    // TODO spec: Consider an explicit abort for this case.
+                    return Err(DapAbort::BadRequest(
+                        "unexpected message for aggregation job (already exists)".into(),
+                    ));
+                }
+
                 let agg_resp = match transition {
                     DapHelperTransition::Continue(state, agg_resp) => {
                         self.put_helper_state(
--- a/daphne/src/roles_test.rs
+++ b/daphne/src/roles_test.rs
@@ -4,95 +4,72 @@
 use crate::{
     auth::BearerToken,
     constants::{MEDIA_TYPE_AGG_INIT_REQ, MEDIA_TYPE_AGG_SHARE_REQ, MEDIA_TYPE_COLLECT_REQ},
-    hpke::{HpkeDecrypter, HpkeReceiverConfig},
+    hpke::HpkeReceiverConfig,
     messages::{
         AggregateInitializeReq, AggregateResp, AggregateShareReq, CollectReq, HpkeCiphertext, Id,
         Interval, Nonce, Report, ReportShare, TransitionFailure, TransitionVar,
     },
     roles::{DapAggregator, DapHelper, DapLeader},
-    testing::{MockAggregator, HPKE_RECEIVER_CONFIG_LIST, LEADER_BEARER_TOKEN},
-    DapAbort, DapError, DapMeasurement, DapRequest, Prio3Config, VdafConfig,
+    testing::{
+        BucketInfo, MockAggregator, ReportStore, HPKE_RECEIVER_CONFIG_LIST, LEADER_BEARER_TOKEN,
+    },
+    DapAbort, DapMeasurement, DapRequest, Prio3Config, VdafConfig,
 };
 use assert_matches::assert_matches;
 use prio::codec::{Decode, Encode};
-use rand::prelude::*;
-use std::vec;
+use rand::{thread_rng, Rng};
+use std::{ops::DerefMut, vec};
 
-#[tokio::test]
-async fn http_post_aggregate_unauthorized_request() {
+fn gen_test_agg_init_req(
+    agg: &MockAggregator,
+    report_shares: Vec<ReportShare>,
+) -> DapRequest<BearerToken> {
     let mut rng = thread_rng();
-    let helper = MockAggregator::new();
-    let task_id = helper.nominal_task_id();
-    let task_config = helper.get_task_config_for(task_id).unwrap();
+    let task_id = agg.nominal_task_id();
+    let task_config = agg.get_task_config_for(task_id).unwrap();
 
-    let mut req = DapRequest {
+    DapRequest {
         media_type: Some(MEDIA_TYPE_AGG_INIT_REQ),
         payload: AggregateInitializeReq {
             task_id: task_id.clone(),
             agg_job_id: Id(rng.gen()),
             agg_param: Vec::default(),
-            report_shares: Vec::default(),
+            report_shares,
         }
         .get_encoded(),
         url: task_config.helper_url.join("/aggregate").unwrap(),
         sender_auth: None,
-    };
-
-    // Expect failure due to missing bearer token.
-    assert_matches!(
-        helper.http_post_aggregate(&req).await,
-        Err(DapAbort::UnauthorizedRequest)
-    );
-
-    // Expect failure due to incorrect bearer token.
-    req.sender_auth = Some(BearerToken::from("incorrect auth token!".to_string()));
-    assert_matches!(
-        helper.http_post_aggregate(&req).await,
-        Err(DapAbort::UnauthorizedRequest)
-    );
+    }
 }
 
-#[tokio::test]
-async fn http_post_aggregate_share_unauthorized_request() {
-    let helper = MockAggregator::new();
-    let task_id = helper.nominal_task_id();
-    let task_config = helper.get_task_config_for(task_id).unwrap();
+fn gen_test_agg_share_req(
+    agg: &MockAggregator,
+    report_count: u64,
+    checksum: [u8; 32],
+) -> DapRequest<BearerToken> {
+    let task_id = agg.nominal_task_id();
+    let task_config = agg.get_task_config_for(task_id).unwrap();
 
-    let mut req = DapRequest {
+    DapRequest {
         media_type: Some(MEDIA_TYPE_AGG_SHARE_REQ),
         payload: AggregateShareReq {
             task_id: task_id.clone(),
             batch_interval: Interval::default(),
             agg_param: Vec::default(),
-            report_count: 0,
-            checksum: [0; 32],
+            report_count,
+            checksum,
         }
         .get_encoded(),
         url: task_config.helper_url.join("/aggregate_share").unwrap(),
         sender_auth: None,
-    };
-
-    // Expect failure due to missing bearer token.
-    assert_matches!(
-        helper.http_post_aggregate_share(&req).await,
-        Err(DapAbort::UnauthorizedRequest)
-    );
-
-    // Expect failure due to incorrect bearer token.
-    req.sender_auth = Some(BearerToken::from("incorrect auth token!".to_string()));
-    assert_matches!(
-        helper.http_post_aggregate_share(&req).await,
-        Err(DapAbort::UnauthorizedRequest)
-    );
+    }
 }
 
-#[tokio::test]
-async fn http_post_collect_unauthorized_request() {
-    let leader = MockAggregator::new();
-    let task_id = leader.nominal_task_id();
-    let task_config = leader.get_task_config_for(task_id).unwrap();
+fn gen_test_collect_req(agg: &MockAggregator) -> DapRequest<BearerToken> {
+    let task_id = agg.nominal_task_id();
+    let task_config = agg.get_task_config_for(task_id).unwrap();
 
-    let mut req = DapRequest {
+    DapRequest {
         media_type: Some(MEDIA_TYPE_COLLECT_REQ),
         payload: CollectReq {
             task_id: task_id.clone(),
@@ -102,71 +79,10 @@ async fn http_post_collect_unauthorized_request() {
         .get_encoded(),
         url: task_config.leader_url.join("/collect").unwrap(),
         sender_auth: None,
-    };
-
-    // Expect failure due to missing bearer token.
-    assert_matches!(
-        leader.http_post_collect(&req).await,
-        Err(DapAbort::UnauthorizedRequest)
-    );
-
-    // Expect failure due to incorrect bearer token.
-    req.sender_auth = Some(BearerToken::from("incorrect auth token!".to_string()));
-    assert_matches!(
-        leader.http_post_collect(&req).await,
-        Err(DapAbort::UnauthorizedRequest)
-    );
+    }
 }
 
-#[tokio::test]
-async fn http_post_aggregate_invalid_ciphertext() {
-    let helper = MockAggregator::new();
-    let task_id = helper.nominal_task_id();
-    let task_config = helper.get_task_config_for(task_id).unwrap();
-
-    let req = DapRequest {
-        media_type: Some(MEDIA_TYPE_AGG_INIT_REQ),
-        payload: AggregateInitializeReq {
-            task_id: task_id.clone(),
-            agg_job_id: Id([1; 32]),
-            agg_param: b"this is an aggregation parameter".to_vec(),
-            report_shares: vec![ReportShare {
-                nonce: Nonce {
-                    time: 1637361337,
-                    rand: [1; 16],
-                },
-                extensions: Vec::default(),
-                encrypted_input_share: HpkeCiphertext {
-                    config_id: 23,
-                    enc: b"invalid encapsulated key".to_vec(),
-                    payload: b"invalid ciphertext".to_vec(),
-                },
-            }],
-        }
-        .get_encoded(),
-        url: task_config.helper_url.join("/aggregate").unwrap(),
-        sender_auth: Some(BearerToken::from(LEADER_BEARER_TOKEN.to_string())),
-    };
-
-    // Get AggregateResp and then extract the transition data from inside.
-    let agg_resp =
-        AggregateResp::get_decoded(&helper.http_post_aggregate(&req).await.unwrap().payload)
-            .unwrap();
-    let transition = &agg_resp.transitions[0];
-
-    // Expect failure due to invalid ciphertext.
-    assert_matches!(
-        transition.var,
-        TransitionVar::Failed(TransitionFailure::HpkeDecryptError)
-    );
-}
-
-#[tokio::test]
-async fn http_post_aggregate_valid_ciphertext() {
-    let helper = MockAggregator::new();
-    let task_id = helper.nominal_task_id();
-    let task_config = helper.get_task_config_for(task_id).unwrap();
-
+fn gen_test_report(task_id: &Id) -> Report {
     // Construct HPKE receiver config List.
     let hpke_receiver_config_list: Vec<HpkeReceiverConfig> =
         serde_json::from_str(HPKE_RECEIVER_CONFIG_LIST)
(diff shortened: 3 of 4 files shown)


nakatsuka-y:
Currently only the basic functionality is implemented.
Test code will be added in the next couple of commits.

daphne/src/testing.rs
```diff
@@ -272,3 +303,150 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+impl HelperStateStore {
+    pub fn new() -> Self {
+        Self {
+            helper_state: Arc::new(Mutex::new(None)),
+        }
+    }
+
+    pub async fn put(&self, helper_state: DapHelperState) -> Result<(), DapError> {
+        if self.helper_state.lock().map_err(int_err)?.deref().is_some() {
+            return Err(int_err("tried to overwrite helper state"));
```
cjpatton, line 431: Can we make this more explicit?
```suggestion
            return Err(DapError::fatal("tried to overwrite helper state"));
```

daphne/src/testing.rs
```diff
@@ -272,3 +303,150 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+#[derive(Deserialize, Serialize)]
+pub(crate) struct HelperStateStoreGetPending {
+    pub(crate) reports_requested: u64,
+}
+
+// TODO(nakatsuka-y remove dead_code once all functions are used)
+#[allow(dead_code)]
+pub(crate) struct HelperStateStore {
+    helper_state: Arc<Mutex<Option<DapHelperState>>>,
+}
```
cjpatton, line 418: Two comments here:
* Only one helper state blob is stored at any given time. However, it's possible that the helper will be handling multiple aggregation flows simultaneously. The data type here should be something like a `HashMap<Id, DapHelperState>` that maps [aggregation job IDs](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.3.1) to state blobs. 
* The contents of this struct are fairly trivial. How about doing away with `HelperStateStore` and moving the `Arc<Mutex<_>>` into `MockAggregator`?

daphne/src/testing.rs
```diff
@@ -61,23 +67,31 @@ pub const COLLECTOR_BEARER_TOKEN: &str = "syfRfvcvNFF5MJk4Y-B7xjRIqD_iNzhaaEB9mY
...
 #[allow(dead_code)]
 impl MockAggregator {
     pub fn new() -> Self {
-        // Construct task list
+        // Construct task list.
         let tasks: HashMap<Id, DapTaskConfig> =
             serde_json::from_str(TASK_LIST).expect("failed to parse task list");
 
-        // Construct HPKE receiver config List
+        // Construct HPKE receiver config List.
```
cjpatton, line 81: nit
```suggestion
        // Construct HPKE receiver config list.
```

daphne/src/testing.rs
```diff
@@ -95,6 +109,11 @@ impl MockAggregator {
         // Just use the first key in the hash map.
         self.tasks.keys().next().as_ref().unwrap()
     }
+
+    #[cfg(test)]
+    pub fn get_report_store(&self) -> &ReportStore {
+        &self.report_store
```
cjpatton, line 115: you could avoid having to add this method by making `report_store` itself `pub(crate)`:

```rust

pub(crate) struct MockAggregator {
    tasks: HashMap<Id, DapTaskConfig>,
    hpke_receiver_config_list: Vec<HpkeReceiverConfig>,
    pub(crate) report_store: ReportStore,
    helper_state_store: HelperStateStore,
}
```

daphne/src/testing.rs
```diff
@@ -272,3 +303,150 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+        TransitionFailure::ReportDropped
+    }
+}
+
+// TODO(nakatsuka-y remove dead_code once all functions are used)
+#[allow(dead_code)]
+pub(crate) struct ReportStore {
+    pending: Arc<Mutex<Vec<Report>>>,
+    processed: Arc<Mutex<HashSet<Nonce>>>,
+    collected: Arc<Mutex<bool>>,
```
cjpatton, line 323: I'm not sure a mux on each individual field is what we want. For example, it's possible that the leader will add more reports to `pending` a task that has already been collected, which would be an error. How about putting a mux around the entire data structure?

daphne/src/testing.rs
```diff
@@ -61,23 +67,31 @@ pub const COLLECTOR_BEARER_TOKEN: &str = "syfRfvcvNFF5MJk4Y-B7xjRIqD_iNzhaaEB9mY
 pub(crate) struct MockAggregator {
     tasks: HashMap<Id, DapTaskConfig>,
     hpke_receiver_config_list: Vec<HpkeReceiverConfig>,
+    report_store: ReportStore,
```
cjpatton, line 70: Thinking ahead, how does the leader pick a set of reports to aggregate for a given task? It'll have scan through this monololthic list to find a set of reports with matching task IDs. This should be something like an `HashMap<Id, Vec<Report>>`

cjpatton (changes requested):
Good start. A couple high-level comments:

* The main thing I notice is that the new ReportStore and HelperStateStore structs are derived more or less from the DOs with the same names. However a key concept has been omitted: Data are stored in a monolothic data structure rather than being grouped. For example:
   * The helper state store structure only permits one aggregation flow at a time, since the helper has effectively one state blob at any given time. In fact, there may be many aggregation jobs running at the same time. (See inline comments.)
   * he report structure contains the same `pending`, `processed`, and `collected` fields of the DO. These fields apply to a given *task* and *batch window*. This allows us to group repors and report meta data in a way that lets us determine efficiently if a particular collect request applies to it.
   *

daphne/src/testing.rs
nakatsuka-y: Good idea, I've made this and another error explicit.

daphne/src/testing.rs
nakatsuka-y: Good point, I've moved `HelperStateStore` directly into `MockAggregator` and mapped aggregation job IDs to `DapHelperState`.

daphne/src/testing.rs
nakatsuka-y: Fixed.

daphne/src/testing.rs
nakatsuka-y: Thank you, I totally missed this.

daphne/src/testing.rs
nakatsuka-y: You are completely right.
I've moved the mutex to surround `ReportStore`.

daphne/src/testing.rs
nakatsuka-y: I have implemented a struct called `BatchInfo` that houses the three data used to store `ReportStore` in DO: `task_id`, `window`, and `bucket`.
I thought some compatibility with `DaphneWorker` would be good, but if you think this is overkill, I can remove `BatchInfo` and replace it with task ID.

nakatsuka-y: `BatchInfo` is now renamed into `BucketInfo` to better represent what it is.
The latest commit also removes `bucket`, which is only necessary for DOs.

daphne/src/testing.rs
```diff
@@ -195,32 +209,103 @@ impl DapAggregator<BearerToken> for MockAggregator {
...
+                    .get(task_id)
+                    .ok_or_else(|| DapError::fatal("no task found"))?,
+                task_id,
+                &report_share.nonce,
+            );
+
+            if !self
+                .report_store
+                .lock()
+                .map_err(int_err)?
```
cjpatton, line 229: Explicit is better:
```suggestion
                .map_err(|e| DapError::Fatal(e.to_string()))?
```

daphne/src/testing.rs
```diff
@@ -195,32 +209,103 @@ impl DapAggregator<BearerToken> for MockAggregator {
...
+
+        for report_share in report_shares.iter() {
+            let bucket_info = BucketInfo::new(
+                self.tasks
+                    .get(task_id)
+                    .ok_or_else(|| DapError::fatal("no task found"))?,
+                task_id,
+                &report_share.nonce,
+            );
+
```
cjpatton, line 225: It looks like you acquire the lock on report_store, then release it up to three times. It would be easier to read (and probably more efficient) to acquire once:
```rust
              let mut report_store = self.report_store.lock().expect("hella");
              if !report_store.contains_key(&bucket_info) {
                  report_store.insert(bucket_info.clone(), ReportStore::new());
              }

              match report_store
                  .get_mut(&bucket_info)
                  .ok_or_else(|| DapError::fatal("no report store found"))?
                  .process_put_processed(report_share.nonce.clone())
              {
                  Ok(()) => (),
                  Err(failure_reason) => {
                      early_fails.insert(report_share.nonce.clone(), failure_reason);
                  }
              }
```

daphne/src/testing.rs
```diff
@@ -272,3 +357,103 @@ impl DapLeader<BearerToken> for MockAggregator {
         unreachable!("not implemented");
     }
 }
+
+// Information associated to a certain report for a given task and nonce to decide which bucket it would be put into.
```
cjpatton, line 376: Docucomments begin with "///" rather than "//".

```suggestion
/// Information associated to a certain report for a given task and nonce to decide which bucket it would be put into.
```

daphne/src/testing.rs
```diff
@@ -272,3 +357,103 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+    pub(crate) reports_requested: u64,
+}
+
+impl From<DapError> for TransitionFailure {
+    fn from(_e: DapError) -> TransitionFailure {
+        TransitionFailure::ReportDropped
+    }
+}
+
+// TODO(nakatsuka-y remove dead_code once all functions are used)
```
cjpatton, line 405:
```suggestion
// TODO(nakatsuka-y) Remove dead_code once all functions are used.
```

daphne/src/testing.rs
```diff
@@ -61,23 +67,31 @@ pub const COLLECTOR_BEARER_TOKEN: &str = "syfRfvcvNFF5MJk4Y-B7xjRIqD_iNzhaaEB9mY
 pub(crate) struct MockAggregator {
     tasks: HashMap<Id, DapTaskConfig>,
     hpke_receiver_config_list: Vec<HpkeReceiverConfig>,
+    pub report_store: Arc<Mutex<HashMap<BucketInfo, ReportStore>>>,
```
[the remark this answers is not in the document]
nakatsuka-y: This is because we have a unit test that directly edits the `collected` field in the report_store.
We can remove this `pub` if we remove that unit test.
(See [here](https://github.com/nakatsuka-y/daphne/blob/11b9205ff349ba7729c58dc6ed797d0e3b530949/daphne/src/roles_test.rs#L248))

daphne/src/testing.rs
```diff
@@ -61,23 +67,31 @@ pub const COLLECTOR_BEARER_TOKEN: &str = "syfRfvcvNFF5MJk4Y-B7xjRIqD_iNzhaaEB9mY
...
         let tasks: HashMap<Id, DapTaskConfig> =
             serde_json::from_str(TASK_LIST).expect("failed to parse task list");
 
-        // Construct HPKE receiver config List
+        // Construct HPKE receiver config list.
         let hpke_receiver_config_list: Vec<HpkeReceiverConfig> =
             serde_json::from_str(HPKE_RECEIVER_CONFIG_LIST)
                 .expect("failed to parse hpke_receiver_config_list");
 
+        // Construct empty report_store and helper_state_store objects.
```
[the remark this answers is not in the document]
nakatsuka-y: Sounds good. I've remove this and several other comments that seemed unnecessary.

daphne/src/lib.rs
```diff
@@ -529,6 +529,11 @@ pub struct DapLeaderProcessTelemetry {
     pub reports_processed: u64,
 }
 
+pub(crate) fn int_err<S: ToString>(s: S) -> DapError {
```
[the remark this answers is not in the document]
nakatsuka-y: I've made these errors explicit and removed `int_err`, so that the errors carry a better message.

daphne/src/testing.rs
nakatsuka-y: Thank you for the suggestion, I completely forgot about this.
I think this makes the code look much cleaner.

daphne/src/testing.rs
nakatsuka-y: Fixed!

daphne/src/testing.rs
nakatsuka-y: Fixed this one as well :)

daphne/src/testing.rs
```diff
@@ -272,3 +357,103 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+            task_id: task_id.clone(),
+            window,
+        }
+    }
+}
+
+#[derive(Deserialize, Serialize)]
+pub(crate) struct ReportStoreGetPending {
+    pub(crate) reports_requested: u64,
+}
```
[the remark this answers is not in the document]
nakatsuka-y: Yes, you are completely right.
I forgot about the existence of this, since it doesn't get used until we call a function we don't use right now.

daphne/src/testing.rs
```diff
@@ -272,3 +365,110 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+
+impl From<DapError> for TransitionFailure {
+    fn from(_e: DapError) -> TransitionFailure {
+        TransitionFailure::ReportDropped
```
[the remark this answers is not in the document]
nakatsuka-y: We need this to propagate `?` in `testing.rs`.

daphne/src/roles_test.rs
```diff
@@ -286,3 +289,176 @@ fn hpke_decrypter() {
...
+
+    // Expect failure due to invalid ciphertext.
+    assert_matches!(
+        transition.var,
+        TransitionVar::Failed(TransitionFailure::BatchCollected)
+    );
+}
+
+#[tokio::test]
+async fn http_post_aggregate_invalid_helper_state_store() {
```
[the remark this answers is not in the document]
nakatsuka-y: Thank you for the suggestion, I was also thinking the same thing.
I've refactored `roles_test.rs` so that the request generation is shared between multiple tests.

daphne/src/roles_test.rs
```diff
@@ -286,3 +289,176 @@ fn hpke_decrypter() {
...
+    };
+
+    let bucket_info = BucketInfo::new(task_config, task_id, &report.nonce.clone());
+
+    // Create a new instance of report store associated to the bucket_info.
+    helper
+        .report_store
+        .lock()
+        .map_err(int_err)
+        .unwrap()
```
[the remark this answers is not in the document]
nakatsuka-y: This has been reflected while removing `int_err`.

daphne/src/testing.rs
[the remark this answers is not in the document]
cjpatton: Because the test is in the same crate as this struct, we can make this `pub(crate)` instead of `pub`. This is better because it ensures that, once we make `MockAggregator` `pub` instead of `pub(crate)`, this field doesn't get exposed in the public API.

daphne/src/testing.rs
```diff
@@ -272,3 +349,103 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+
+// TODO(nakatsuka-y) remove dead_code once all functions are used
+#[allow(dead_code)]
+pub(crate) struct ReportStore {
+    pending: Vec<Report>,
+    processed: HashSet<Nonce>,
+    collected: bool,
+}
+
+macro_rules! checked_process {
```
cjpatton, line 386: For readability, I would suggest replacing this macro with a method on `ReportStore`.

daphne/src/testing.rs
```diff
@@ -195,32 +205,99 @@ impl DapAggregator<BearerToken> for MockAggregator {
...
+            agg_job_id: agg_job_id.clone(),
+        };
+
+        let mut helper_state_store_mutex_guard = self
+            .helper_state_store
+            .lock()
+            .map_err(|e| DapError::Fatal(e.to_string()))?;
+
+        let helper_state_store = helper_state_store_mutex_guard.deref_mut();
+
```
cjpatton, line 265: This code is correct, however it won't be correct once we add a VDAF that has more than one round of preparation. (Prio3 has one round.) Can we leave a note here so that we don't forget?

```rust 
// NOTE: This code is only correct for VDAFs with exactly one round of preparation.
// For VDAFs with more rounds, the helper state blob will need to be updated here.
```

daphne/src/testing.rs
```diff
@@ -195,32 +205,99 @@ impl DapAggregator<BearerToken> for MockAggregator {
...
```
cjpatton, line 291: We should repeat the same `NOTE` here.

daphne/src/roles_test.rs
```diff
@@ -4,39 +4,117 @@
...
+    let task_id = agg.nominal_task_id();
+    let task_config = agg.get_task_config_for(task_id).unwrap();
 
-    let mut req = DapRequest {
+    DapRequest {
         media_type: Some(MEDIA_TYPE_AGG_INIT_REQ),
         payload: AggregateInitializeReq {
             task_id: task_id.clone(),
-            agg_job_id: Id(rng.gen()),
+            agg_job_id: Id([1; 32]),
```
cjpatton, line 34: This should be generated at random:
```suggestion
            agg_job_id: Id(rng.gen()),
```
You'll need `let mut rng = thread_rng()` above.

daphne/src/roles_test.rs
```diff
@@ -4,39 +4,117 @@
...
-#[tokio::test]
-async fn http_post_aggregate_unauthorized_request() {
-    let mut rng = thread_rng();
-    let helper = MockAggregator::new();
-    let task_id = helper.nominal_task_id();
-    let task_config = helper.get_task_config_for(task_id).unwrap();
+fn gen_agg_init_req(
+    agg: &MockAggregator,
+    report_shares: Vec<ReportShare>,
+    sender_auth: Option<BearerToken>,
```
cjpatton, line 25: I think it would be cleaner to drop this input parameter and force the caller to add sender auth as needed.

daphne/src/testing.rs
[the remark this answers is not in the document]
nakatsuka-y: Ah, I understood your comment now. I wasn't sure what the difference between `pub` and `pub(crate)` was, but I think I got the idea.

daphne/src/roles_test.rs
```diff
@@ -218,71 +225,77 @@ async fn http_post_aggregate_valid_ciphertext() {
...
-                &report.task_id,
-                info,
-                aad,
-                &report.encrypted_input_shares[0]
-            )
-            .unwrap(),
-        plaintext
-    );
+#[tokio::test]
+async fn http_post_aggregate_abort_helper_state_store() {
```
cjpatton, line 278:
```suggestion
async fn http_post_aggregate_abort_helper_state_overwritten() {
```

daphne/src/roles_test.rs
```diff
@@ -218,71 +225,77 @@ async fn http_post_aggregate_valid_ciphertext() {
...
+    let req = gen_agg_init_req(&helper, report_shares, sender_auth);
+
+    // Send aggregate request.
+    let _ = &helper.http_post_aggregate(&req).await;
+
+    // Send another aggregate request.
+    let res = &helper.http_post_aggregate(&req).await;
+
+    // Expect failure due to overwriting helper state.
+    assert_matches!(res, Err(DapAbort::Internal(_)));
```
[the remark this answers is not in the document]
nakatsuka-y: The exact error I am expecting is the following:
```rust
// Expect failure due to overwriting helper state.
assert_matches!(
    res,
    Err(DapError::fatal("tried to overwrite helper state").into())
);
```
However, I keep getting the following error:
```rust
expected tuple struct or tuple variant, found associated function DapError::fatal
```
I've been circling around different sources and code, but I can't seem to wrap my head around this issue.
Can you help me solve this?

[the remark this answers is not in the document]
cjpatton: `fatal()` is a method that constructs a `DapError`; `Fatal` is a variant of `DapError`.

daphne/src/testing.rs
```diff
@@ -223,25 +219,18 @@ impl DapHelper<BearerToken> for MockAggregator {
...
-                .contains_key(&bucket_info)
-            {
-                self.report_store
-                    .lock()
-                    .map_err(int_err)?
-                    .deref_mut()
-                    .insert(bucket_info.clone(), ReportStore::new());
+                .map_err(|e| DapError::Fatal(e.to_string()))?;
+
+            let report_store = report_store_mutex_guard.deref_mut();
```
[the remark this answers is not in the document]
nakatsuka-y: This makes total sense. I've moved the mutex lock out of the for loop.

daphne/src/testing.rs
nakatsuka-y: Agreed. Changed `checked_process` into struct method.

daphne/src/testing.rs
```diff
@@ -272,3 +349,103 @@ impl DapLeader<BearerToken> for MockAggregator {
         unreachable!("not implemented");
     }
 }
+
+/// Information associated to a certain helper state for a given task ID and aggregate job ID.
+#[derive(Clone, Eq, Hash, PartialEq, Deserialize, Serialize)]
+pub(crate) struct HelperStateId {
```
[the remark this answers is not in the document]
nakatsuka-y: Great point! I agree that "info" makes more sense here.

daphne/src/testing.rs
```diff
@@ -272,3 +349,103 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+
+/// Information associated to a certain report for a given task and nonce to decide which bucket it would be put into.
+#[derive(Clone, Eq, Hash, PartialEq, Deserialize, Serialize)]
+pub(crate) struct BucketInfo {
+    task_id: Id,
+    window: u64,
+}
+
+impl BucketInfo {
+    pub fn new(task_config: &DapTaskConfig, task_id: &Id, nonce: &Nonce) -> Self {
```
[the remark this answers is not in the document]
nakatsuka-y: Thank you for pointing this out. I've removed unnecessary `pub`s

daphne/src/testing.rs
```diff
@@ -272,3 +349,103 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+        let collected = false;
+
+        Self {
+            pending,
+            processed,
+            collected,
+        }
+    }
+
+    pub fn process_delete_all(&mut self) -> Result<(), TransitionFailure> {
```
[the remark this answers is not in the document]
nakatsuka-y: Yet another great point. This comes from the DO reference. Removed unnecessary errors.
We still need to return `TransitionFailure`, since the parent function that calls these functions is a trait, and it expects `TransitionFailure`, not `DapError`.

daphne/src/testing.rs
nakatsuka-y: I've added this note to here and another location below this.

daphne/src/testing.rs
nakatsuka-y: Ditto.

daphne/src/roles_test.rs
nakatsuka-y: You are right. Got carried away when writing the request generation functions.

daphne/src/roles_test.rs
```diff
@@ -4,39 +4,117 @@
...
-use std::vec;
+use std::{ops::DerefMut, vec};
 
-#[tokio::test]
-async fn http_post_aggregate_unauthorized_request() {
-    let mut rng = thread_rng();
-    let helper = MockAggregator::new();
-    let task_id = helper.nominal_task_id();
-    let task_config = helper.get_task_config_for(task_id).unwrap();
+fn gen_agg_init_req(
```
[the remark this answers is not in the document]
nakatsuka-y: I didn't know this, but makes total sense. I'll make sure to do this.

daphne/src/roles_test.rs
nakatsuka-y: Sounds good. Callers of these functions are now responsible for adding `sender_auth` now.

daphne/src/roles_test.rs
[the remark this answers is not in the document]
nakatsuka-y: I have introduced a new error in `DapError` called `BadRequest`, which is already reflected in `DapAbort`.
While `Internal` represents errors coming internally, `BadRequest` represents errors coming externally.
The new error checking code is as follows:
```rust
// Expect failure due to overwriting helper state.
assert_matches!(res.unwrap_err(), DapAbort::BadRequest(s) => {
    assert_eq!(s, "bad request error: tried to overwrite helper state")
});
```

daphne/src/roles_test.rs
nakatsuka-y: Fixed :)

daphne/src/lib.rs
```diff
@@ -48,6 +48,10 @@ pub enum DapError {
     #[error("fatal error: {0}")]
     Fatal(String),
 
+    /// Bad request error. If this triggers an abort, then treat this as an external error.
+    #[error("bad request error: {0}")]
+    BadRequest(String),
```
cjpatton, line 53: I think an explicit error would be best here. Eventually we'll need to address this case in the spec itself.

```suggestion
    /// Received an unexpected message.
    // 
    // TODO spec: Handle this edge case in the spec itself.
    #[error("bad request error: {0}")]
    UnexpectedMessage,
```

cjpatton (changes requested):
This is a reasonable way to solve this bug. However it does require making a corresponding change in Daphne-Worker.

One downside to this approach is that it pushes protocol logic into the Daphne backend (i.e., Daphne-Worker or `MockAggregator`). Another solution is to modify `http_post_aggregate()` in roles.rs so that the helper first checks if the state for the agg job already exists before overwriting it. (This would require a call to `get_helper_state()`.)

daphne/src/lib.rs
```diff
@@ -188,6 +197,7 @@ impl From<DapError> for DapAbort {
     fn from(e: DapError) -> Self {
         match e {
             e @ DapError::Fatal(..) => Self::Internal(Box::new(e)),
+            e @ DapError::BadRequest(..) => Self::BadRequest(e.to_string()),
```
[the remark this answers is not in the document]
nakatsuka-y: Outdated, since we have removed `DapError::BadRequest`.

daphne/src/testing.rs
```diff
@@ -263,11 +263,14 @@ impl DapHelper<BearerToken> for MockAggregator {
 
         let helper_state_store = helper_state_store_mutex_guard.deref_mut();
 
-        if helper_state_store.contains_key(&helper_state_id) {
-            return Err(DapError::fatal("tried to overwrite helper state"));
+        // NOTE: This code is only correct for VDAFs with exactly one round of preparation.
+        // For VDAFs with more rounds, the helper state blob will need to be updated here.
+
```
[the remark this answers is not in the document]
nakatsuka-y: Fixed!

daphne/src/vdaf/mod.rs
```diff
@@ -261,17 +263,18 @@ impl VdafConfig {
...
             processed.insert(report.nonce.clone());
 
             if &report.task_id != task_id || report.encrypted_input_shares.len() != 2 {
-                return Err(
-                    DapError::fatal("tried to process report with incorrect task ID").into(),
-                );
+                return Err(DapError::bad_request(
+                    "tried to process report with incorrect task ID",
+                )
+                .into());
```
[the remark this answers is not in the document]
nakatsuka-y: Ditto.

daphne/src/vdaf/mod.rs
```diff
@@ -722,7 +725,7 @@ impl VdafConfig {
         }
 
         if agg_shares.len() != encrypted_agg_shares.len() {
-            return Err(DapError::Fatal(
+            return Err(DapError::BadRequest(
```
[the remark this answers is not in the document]
nakatsuka-y: Updated!

daphne/src/vdaf/mod.rs
```diff
@@ -743,7 +746,7 @@ fn produce_encrypted_agg_share(
     let agg_share_data = agg_share
         .data
         .as_ref()
-        .ok_or_else(|| DapError::fatal("empty aggregate share"))?
+        .ok_or_else(|| DapError::bad_request("empty aggregate share"))?
```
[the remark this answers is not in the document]
nakatsuka-y: Updated.

daphne/src/lib.rs
nakatsuka-y: We have removed `BadRequest` entirely.

daphne/src/roles.rs
```diff
@@ -444,6 +444,15 @@ pub trait DapHelper<S>: DapAggregator<S> {
                     &early_rejects,
                 )?;
 
+                // Check that helper state with task_id and agg_job_id does not exist.
+                if self
+                    .get_helper_state(&agg_init_req.task_id, &agg_init_req.agg_job_id)
+                    .await?
+                    .is_some()
+                {
+                    return Err(DapAbort::ReplayedReport);
```
cjpatton, line 453:
```suggestion
                    // TODO spec: Consider an explicit abort for this case.
                    return Err(DapAbort::BadRequest("unexpected message for aggregation job".into()));
```

daphne/src/roles_test.rs
```diff
@@ -218,71 +224,76 @@ async fn http_post_aggregate_valid_ciphertext() {
...
+    req.sender_auth = sender_auth;
+
+    // Send aggregate request.
+    let _ = helper.http_post_aggregate(&req).await;
+
+    // Send another aggregate request.
+    let res = helper.http_post_aggregate(&req).await;
+
+    // Expect failure due to overwriting existing helper state.
+    assert_matches!(res.unwrap_err(), DapAbort::ReplayedReport);
```
cjpatton, line 298: After fixing `roles.rs`, the coresponding change here will look like
```rust
  assert_matches!(res.unwrap_err(), DapAbort::BadRequest(e) if e == "blah blah blah");
```

daphne/src/testing.rs
```diff
@@ -195,32 +205,98 @@ impl DapAggregator<BearerToken> for MockAggregator {
...
+        let mut helper_state_store_mutex_guard = self
+            .helper_state_store
+            .lock()
+            .map_err(|e| DapError::Fatal(e.to_string()))?;
+
+        let helper_state_store = helper_state_store_mutex_guard.deref_mut();
+        // NOTE: This code is only correct for VDAFs with exactly one round of preparation.
+        // For VDAFs with more rounds, the helper state blob will need to be updated here.
+
```
cjpatton, line 292: nit: Remove empty line between the comment and the code it applies to.

cjpatton (changes requested):
Great. Just a few minor comments to address, squash, and merge.

cjpatton (approved):

Issue #20, opened 2022-06-23
Daphne: Expose a "mock" backend in a public module


cjpatton:
`MockAggregator` in `daphne/src/roles_test.rs`, once finished, will be a useful tool for demonstrating the functionality the Daphne backend is supposed to provide. This should be put in a public module for others to copy-paste or refer to as needed.

While at, let's write tests that excercise the protocol logic with the mock backend.

cjpatton:
cc/ @nakatsuka-y

nakatsuka-y:
# Immediate tasks
- [ ] Implement HpkeDecrypter methods (Ref: #32)
- [ ] Implement DapAggregator methods
- [ ] Implement DapHelper methods
- [ ] Implement DapLeader methods

# Next steps
- [ ] Use MockAggregator to simulate Leader - Helper interactions
- [ ] Use MockAggregator to simulate Leader - Collector interactions
- [ ] Expose MockAggregator to public

nakatsuka-y:
We should have implemented all the necessary components.
I will be closing this issue, but feel free to re-open and ping me if there are any missing things related to this issue.

PR #49, opened 2022-07-19
Implement `put_report` and test `http_post_upload`

nakatsuka-y:
This implements the `put_report` method called within `http_post_upload`.
`http_post_upload` is used by the Leader to handle http post requests from Clients.

The changes include unit tests that check for malformed reports as well as a test that successfully receives a report.

Ref: #20.

### DIFF
--- a/daphne/src/roles_test.rs
+++ b/daphne/src/roles_test.rs
@@ -3,7 +3,10 @@
 
 use crate::{
     auth::BearerToken,
-    constants::{MEDIA_TYPE_AGG_INIT_REQ, MEDIA_TYPE_AGG_SHARE_REQ, MEDIA_TYPE_COLLECT_REQ},
+    constants::{
+        MEDIA_TYPE_AGG_INIT_REQ, MEDIA_TYPE_AGG_SHARE_REQ, MEDIA_TYPE_COLLECT_REQ,
+        MEDIA_TYPE_REPORT,
+    },
     hpke::HpkeReceiverConfig,
     messages::{
         AggregateInitializeReq, AggregateResp, AggregateShareReq, CollectReq, HpkeCiphertext, Id,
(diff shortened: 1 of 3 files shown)


daphne/src/roles_test.rs
```diff
@@ -367,9 +367,10 @@ async fn http_post_upload_fail_send_invalid_report() {
     );
 }
 
-// This tests a successful Client --> Leader functionality.
+// This currently tests successful Client --> Leader functionality.
```
cjpatton, line 370: By the way, it is still useful to have this test in addition to a full end-to-end test. It provides more feedback to the developer in case a change is causing one to fail but not the other.

```suggestion
// Test the upload sub-protocol.
```

daphne/src/roles_test.rs
```diff
@@ -20,6 +23,18 @@ use prio::codec::{Decode, Encode};
 use rand::{thread_rng, Rng};
 use std::{ops::DerefMut, vec};
 
+fn gen_test_upload_req(agg: &MockAggregator, report: Report) -> DapRequest<BearerToken> {
```
cjpatton, line 26: Suggestion (here and below): It would be a bit cleaner to define this function as part of MockAggregator's implementation:

```rust
impl MockAggregator {
    fn gen_test_upload(&self, report: Report) -> DapRequest<BearerToken> {
        ...
    }
}
```

daphne/src/roles_test.rs
```diff
@@ -299,3 +314,69 @@ async fn http_post_aggregate_abort_helper_state_overwritten() {
         assert_eq!(e, "unexpected message for aggregation job (already exists)")
     );
 }
+
+#[tokio::test]
+async fn http_post_upload_fail_send_invalid_report() {
+    let leader = MockAggregator::new();
+    let task_id = leader.nominal_task_id();
+
+    // Construct a report payload with a 0'd out task ID.
```
cjpatton, line 323: It doesn't matter that it's 0'd out; what is relevant is that it's not the right task ID.

daphne/src/lib.rs
```diff
@@ -95,6 +95,12 @@ impl From<::hpke::HpkeError> for DapError {
     }
 }
 
+impl From<TransitionFailure> for DapError {
```
cjpatton, line 98: I think the code would be clearer without this. Beware of overusing the `From<A> for B` -- in general it's best to handle things explicitly.

cjpatton (changes requested):
Nice, just a few comments. Also: In the future, before writing tests, check the rest of the crate to see if there is already test code that covers the change. Adding more tests than needed can't hurt, but it does add maintenance burden, so it's important to think carefully about what we need. In particular, `vdaf/mod_test.rs` already has tests that exercise some of the corner cases in the new tests.

That said, let's keep the new tests you added. Just something to keep in mind for the future :)

daphne/src/roles_test.rs
nakatsuka-y: I've added the sub-protocol test as well.

daphne/src/roles_test.rs
nakatsuka-y: Great suggestion, I think this would make the code a bit more readable as well.

daphne/src/roles_test.rs
nakatsuka-y: Right. Updated comment.

daphne/src/lib.rs
nakatsuka-y: I thought the code would look cleaner because we just have to use `?`, but I have removed this and reverted back to handling the error explicitly.

cjpatton (commented):
Changes look good, go ahead and squash. By the way the title of of this PR is "Implement and test http_post_upload", which is not entirely accurate. What is implemented is `put_report` for `MockAggregator`, which allows `http_post_upload` to be tested.

cjpatton (approved):

PR #50, opened 2022-07-20
Remove ignored profile stanza

divergentdave:
I see the following warning when building:

```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /tmp/dap_test/daphne_worker_test/Cargo.toml
workspace: /tmp/dap_test/Cargo.toml
```

This same setting is already present in the workspace root `Cargo.toml`, so we can just remove the stanza under `daphne_worker_test/`.


cjpatton (approved):
Thanks, David!

PR #51, opened 2022-07-20
Implement `get_reports`

nakatsuka-y:
This PR implements `get_reports`, which is used by the Leader to retrieve reports from `pending` report store when preparing aggregation initialization requests.

Several test cases that checks the functionality of this method are also added.

Partially addresses issue #20.

### DIFF
--- a/daphne/src/roles_test.rs
+++ b/daphne/src/roles_test.rs
@@ -14,9 +14,10 @@ use crate::{
     },
     roles::{DapAggregator, DapHelper, DapLeader},
     testing::{
-        BucketInfo, MockAggregator, ReportStore, HPKE_RECEIVER_CONFIG_LIST, LEADER_BEARER_TOKEN,
+        BucketInfo, MockAggregateInfo, MockAggregator, ReportStore, HPKE_RECEIVER_CONFIG_LIST,
+        LEADER_BEARER_TOKEN,
     },
-    DapAbort, DapMeasurement, DapRequest, Prio3Config, VdafConfig,
+    DapAbort, DapError, DapMeasurement, DapRequest, Prio3Config, VdafConfig,
 };
 use assert_matches::assert_matches;
 use prio::codec::{Decode, Encode};
@@ -306,10 +307,10 @@ async fn http_post_aggregate_abort_helper_state_overwritten() {
     let _ = helper.http_post_aggregate(&req).await;
 
     // Send another aggregate request.
-    let res = helper.http_post_aggregate(&req).await;
+    let err = helper.http_post_aggregate(&req).await.unwrap_err();
 
     // Expect failure due to overwriting existing helper state.
-    assert_matches!(res.unwrap_err(), DapAbort::BadRequest(e) =>
+    assert_matches!(err, DapAbort::BadRequest(e) =>
         assert_eq!(e, "unexpected message for aggregation job (already exists)")
     );
 }
@@ -366,6 +367,66 @@ async fn http_post_upload_fail_send_invalid_report() {
     );
 }
 
+#[tokio::test]
+async fn get_reports_fail_invalid_batch_interval() {
+    let leader = MockAggregator::new();
+    let task_id = leader.nominal_task_id();
+
+    // Attempt to get reports using an empty interval.
+    let empty_interval = Interval {
+        start: 0,
+        duration: 0,
+    };
+    let selector = &MockAggregateInfo {
+        batch_info: Some(empty_interval),
+        agg_rate: 1,
+    };
+    let err = leader.get_reports(task_id, selector).await.unwrap_err();
+
+    // Fails because a 0 second duration interval is not permitted.
+    assert_matches!(err, DapError::Fatal(e) =>
+        assert_eq!(e, "invalid batch interval")
+    );
+}
+
+#[tokio::test]
+async fn get_reports_empty_response() {
+    let leader = MockAggregator::new();
+    let task_id = leader.nominal_task_id();
+    let task_config = leader.get_task_config_for(task_id).unwrap();
+
+    let report = leader.gen_test_report(task_id);
+    let req = leader.gen_test_upload_req(report.clone());
+
+    // Upload report.
+    leader
+        .http_post_upload(&req)
+        .await
+        .expect("upload failed unexpectedly");
+
+    // Get report.
+    let now = report.nonce.time - 200000000;
+    let selector = &MockAggregateInfo {
+        batch_info: Some(task_config.current_batch_window(now)),
+        agg_rate: 1,
+    };
+    let reports = leader.get_reports(task_id, selector).await.unwrap();
+
+    // We get an empty response due to no reports existing within requested batch_window.
+    assert_eq!(reports.len(), 0);
+
+    // Attempt to get reports from the future.
+    let now = report.nonce.time + 200000000;
+    let selector = &MockAggregateInfo {
+        batch_info: Some(task_config.current_batch_window(now)),
+        agg_rate: 1,
+    };
+    let reports = leader.get_reports(task_id, selector).await.unwrap();
+
+    // We get an empty response due to no reports existing within requested batch_window.
+    assert_eq!(reports.len(), 0);
+}
+
 // Test the upload sub-protocol.
 #[tokio::test]
 async fn successful_http_post_upload() {
@@ -375,8 +436,10 @@ async fn successful_http_post_upload() {
     let report = leader.gen_test_report(task_id);
     let req = leader.gen_test_upload_req(report);
 
-    let res = &leader.http_post_upload(&req).await;
-    assert_matches!(res, Ok(()));
+    leader
+        .http_post_upload(&req)
+        .await
+        .expect("upload failed unexpectedly");
 }
 
 // Test the end-to-end protocol.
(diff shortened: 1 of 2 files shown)


daphne/src/roles_test.rs
```diff
@@ -385,10 +448,22 @@ async fn successful_http_post_upload() {
...
+    let task_config = leader.get_task_config_for(task_id).unwrap();
 
     let report = leader.gen_test_report(task_id);
-    let req = leader.gen_test_upload_req(report);
+    let req = leader.gen_test_upload_req(report.clone());
 
-    let res = &leader.http_post_upload(&req).await;
+    // Client --> Leader
+    let res = leader.http_post_upload(&req).await;
     assert_matches!(res, Ok(()));
```
cjpatton, line 458: Here again, I think we should unwrap the result.

daphne/src/testing.rs
```diff
@@ -341,10 +348,66 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+            if !interval.is_valid_for(task_config) {
+                return Err(DapError::fatal("invalid batch interval"));
+            }
+            interval.clone()
+        } else {
+            task_config.current_batch_window(Utc::now().timestamp() as u64)
+        };
+
+        // Do some checks on batch interval.
+        if batch_interval.start > Utc::now().timestamp() as u64 {
```
cjpatton, line 369: This check would fail if the start of the batch interval is in the future, but I'm not sure that's actually what we want.

Tip: Test code should not be time sensitive, unless time sensitivity is the thing being tested. Here I'm not sure it is.

daphne/src/testing.rs
```diff
@@ -341,10 +348,66 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+            .map_err(|e| DapError::Fatal(e.to_string()))?;
+        let report_store = report_store_mutex_guard.deref_mut();
+
+        // Fetch reports.
+        let mut window = batch_interval.start;
+        // TODO(nakatsuka-y) Should the batch window be inclusive?
+        while window < batch_interval.end() {
+            let num_reports_remaining = agg_rate - reports.len();
+            let bucket_info = BucketInfo {
+                task_id: task_id.clone(),
```
cjpatton, line 390: We can easily avoid this clone by moving `bucket_info` out of the loop and updating `bucket_info.window` as needed.

daphne/src/testing.rs
```diff
@@ -341,10 +348,66 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+            // If not, move on to the next bucket.
+            if let Some(ref mut store) = report_store.get_mut(&bucket_info) {
+                reports.append(&mut store.process_get_pending(num_reports_remaining as u64));
+                match reports.len().cmp(&agg_rate) {
+                    Ordering::Less => (),
+                    Ordering::Greater => return Err(DapError::fatal(
+                        "number of reports received from report_store exceeds the number requested",
+                    )),
+                    Ordering::Equal => break,
+                }
```
cjpatton, line 404: Hmm, this `match` statement strikes me as somewhat awkward. In particular, since `store` is in memory (and not implemented by some external service, as is the case for the DO code), why do we need the `store_proecess_getpending()` callback at all? Why not access `store`'s memory directly? This would let us simplify this code a great deal.

daphne/src/testing.rs
```diff
@@ -341,10 +348,66 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+            .ok_or_else(|| DapError::fatal("no task found"))?;
+
+        // Calculate batch interval.
+        let batch_interval = if let Some(ref interval) = selector.batch_info {
+            if !interval.is_valid_for(task_config) {
+                return Err(DapError::fatal("invalid batch interval"));
+            }
+            interval.clone()
+        } else {
+            task_config.current_batch_window(Utc::now().timestamp() as u64)
```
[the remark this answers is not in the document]
nakatsuka-y: I didn't know that such crate existed. Thanks for the pointer.

daphne/src/roles_test.rs
```diff
@@ -366,6 +367,68 @@ async fn http_post_upload_fail_send_invalid_report() {
...
+        duration: 0,
+    };
+    let selector = &MockAggregateInfo {
+        batch_info: Some(empty_interval),
+        agg_rate: 1,
+    };
+    let res = leader.get_reports(task_id, selector).await;
+
+    // Fails due to is_valid_for check.
+    assert_matches!(res.unwrap_err(), DapError::Fatal(e) =>
```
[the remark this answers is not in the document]
nakatsuka-y: I think this is a better representation of the variable.

daphne/src/roles_test.rs
```diff
@@ -366,6 +367,68 @@ async fn http_post_upload_fail_send_invalid_report() {
...
+    let leader = MockAggregator::new();
+    let task_id = leader.nominal_task_id();
+    let task_config = leader.get_task_config_for(task_id).unwrap();
+
+    let report = leader.gen_test_report(task_id);
+    let req = leader.gen_test_upload_req(report.clone());
+
+    // Upload report.
+    let res = leader.http_post_upload(&req).await;
+    assert_matches!(res, Ok(()));
```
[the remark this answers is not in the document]
nakatsuka-y: Good point. Removed and other places that do unnecessary asserts.

daphne/src/roles_test.rs
```diff
@@ -385,10 +448,22 @@ async fn successful_http_post_upload() {
...
     let leader = MockAggregator::new();
     let task_id = leader.nominal_task_id();
+    let task_config = leader.get_task_config_for(task_id).unwrap();
 
     let report = leader.gen_test_report(task_id);
-    let req = leader.gen_test_upload_req(report);
+    let req = leader.gen_test_upload_req(report.clone());
 
-    let res = &leader.http_post_upload(&req).await;
+    // Client --> Leader
```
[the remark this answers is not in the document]
nakatsuka-y: Added a more descriptive comment.

daphne/src/testing.rs
nakatsuka-y: Probably we should return an empty `Vec<Report>` even if the interval is to the future.
I'll work on this.

daphne/src/testing.rs
```diff
@@ -341,10 +348,66 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+
+        // Do some checks on batch interval.
+        if batch_interval.start > Utc::now().timestamp() as u64 {
+            return Err(DapError::fatal(
+                "invalid batch interval: start time cannot be in the future",
+            ));
+        }
+
+        // Lock report_store.
+        let agg_rate = selector.agg_rate as usize;
```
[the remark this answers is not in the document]
nakatsuka-y: Changed :)

daphne/src/testing.rs
nakatsuka-y: Great catch.

daphne/src/testing.rs
nakatsuka-y: I've replaced `store.process_get_pending` with directly accessing `pending`.
I think the code looks cleaner now.

daphne/src/testing.rs
nakatsuka-y: Simply removing this checks returns an empty vector, which I think is expected.

daphne/src/roles_test.rs
```diff
@@ -366,6 +367,68 @@ async fn http_post_upload_fail_send_invalid_report() {
     );
 }
 
+#[tokio::test]
+async fn get_reports_fail_invalid_batch_interval() {
+    let leader = MockAggregator::new();
+    let task_id = leader.nominal_task_id();
+    let task_config = leader.get_task_config_for(task_id).unwrap();
+
+    // Test an empty interval.
```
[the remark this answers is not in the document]
nakatsuka-y: Updated comments.

nakatsuka-y:
Since reports are stored in buckets of time windows (see `BucketInfo`), I think it is natural to specify a time interval when selecting such buckets.
If we want to remove `Interval`s, then we would also need to replace the bucket representation with a different data.
I also don't think it would be a good idea to deviate too much from the specs, as nonces and batch intervals seem to be a center part of managing reports sent from Clients.

cjpatton (commented):
Looks good, go ahead and squash and I'll approve. One quick comment for you to consider first:

> Since reports are stored in buckets of time windows (see `BucketInfo`), I think it is natural to specify a time interval when selecting such buckets. If we want to remove `Interval`s, then we would also need to replace the bucket representation with a different data. I also don't think it would be a good idea to deviate too much from the specs, as nonces and batch intervals seem to be a center part of managing reports sent from Clients.

I agree, we must not deviate from the spec. However I don't think the spec mandates any particular behavior as to how reports are selected for aggregation. In fact, I htink the only requirement is that they all pertain to the same task: See https://www.ietf.org/archive/id/draft-ietf-ppm-dap-01.html#name-aggregate-initialization.

I also don't think we would need to change the structure of the code. You simply iterate over the buckets the corresponding to the task ID:
```rust
          let mut reports = Vec::new();
          for bucket_info in report_store.keys() {
              if &bucket_info.task_id == task_id {
                  // drain report_store.get(bucket_info), break the `for` loop if `reports.len() ==
                  // agg_rate`.`
              }
          }
```
You might also consider doing it in sorted order by `bucket_info.batch_interval.start`. so that the oldest data is prioritized for aggregation.

In my opinion, this would make running tests with `MockAggregator` much easier, particularly because you don't have to muck around with the clock. (Again, test could should not use the clock if this can be avoided.)

cjpatton (changes requested):
The commit message needs more information:

> Implement get_reports

Imagine you're unfamiliar with Daphne and you're looking through history in 3 history. The message doesn't provide any information about what functionality this commit adds or changes.

Tip: It's helpful to prefix the message with the component that's changing.

> `MockAggregator`: Implement `get_reports`

Also, it's helpful to know that `get_reports` refers to a method on a trait:

> `MockAggregator`: Implement `DapLeader::get_reports()`

Finally, in the body of the commit, I think it would be useful to note that new tests are added that exercise more of the code.

cjpatton (approved):

PR #56, opened 2022-07-21
Implement `put_out_shares()`

nakatsuka-y:
This PR implements `put_out_shares()` which is used by Leader to merge processed share from itself and Helper after finishing the continuation phase.

A unit test that tests whether an initialization request has been sent before the continuation request has been sent.
`put_out_shares()` has also been integrated to the e2e test alongside the Leader sending the aggregation continuation request to Helper.

This partially implements #20 (continuation phase).

### DIFF
--- a/daphne/src/roles_test.rs
+++ b/daphne/src/roles_test.rs
@@ -4,20 +4,21 @@
 use crate::{
     auth::BearerToken,
     constants::{
-        MEDIA_TYPE_AGG_INIT_REQ, MEDIA_TYPE_AGG_SHARE_REQ, MEDIA_TYPE_COLLECT_REQ,
-        MEDIA_TYPE_REPORT,
+        MEDIA_TYPE_AGG_CONT_REQ, MEDIA_TYPE_AGG_INIT_REQ, MEDIA_TYPE_AGG_SHARE_REQ,
+        MEDIA_TYPE_COLLECT_REQ, MEDIA_TYPE_REPORT,
     },
     hpke::HpkeReceiverConfig,
     messages::{
-        AggregateInitializeReq, AggregateResp, AggregateShareReq, CollectReq, HpkeCiphertext, Id,
-        Interval, Nonce, Report, ReportShare, TransitionFailure, TransitionVar,
+        AggregateContinueReq, AggregateInitializeReq, AggregateResp, AggregateShareReq, CollectReq,
+        HpkeCiphertext, Id, Interval, Nonce, Report, ReportShare, Transition, TransitionFailure,
+        TransitionVar,
     },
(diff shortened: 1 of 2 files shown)


daphne/src/roles_test.rs
```diff
@@ -469,4 +506,55 @@ async fn e2e() {
...
+        .vdaf
+        .produce_agg_init_req(
+            &leader,
+            &task_config.vdaf_verify_key,
+            task_id,
+            &agg_job_id,
+            reports,
+        )
+        .unwrap();
+    assert_matches!(transition, DapLeaderTransition::Continue(_, _));
```
cjpatton, line 523: A bit nicer:
```suggestion
    assert_matches!(transition, DapLeaderTransition::Continue(..));
```

cjpatton (changes requested):
Tests:
* We should also add a test to make sure that sender auth is enforce for agg continue request. (Also for agg init request if missing.)
* For the e2e tests, it looks like you're manually constructiong the agg init and agg continue requests. Why not use `DapLeader::process()` to drive this instead?

daphne/src/roles_test.rs
nakatsuka-y: Fixed! Thanks for the suggestion.

nakatsuka-y:
Sender auth test for `AggregateContReq` was missing, so I've added it. This test for `AggregateInitReq` was already there.

`DapLeader::process()` has many function that we don't currently support, mainly the collection routines.
This is why I'm writing a separate e2e test.
At the end of the development cycle we should be able to run `process()`.

daphne/src/testing.rs
```diff
@@ -212,7 +212,7 @@ impl DapAggregator<BearerToken> for MockAggregator {
             if let Some(agg_share) = agg_store.get_mut(&bucket_info) {
                 agg_share.merge(agg_share_delta)?;
             } else {
-                agg_store.insert(bucket_info.clone(), DapAggregateShare::default());
+                agg_store.insert(bucket_info.clone(), agg_share_delta);
```
cjpatton, line 215: Good catch :)

cjpatton (commented):
> Sender auth test for `AggregateContReq` was missing, so I've added it. This test for `AggregateInitReq` was already there.
> 
> `DapLeader::process()` has many function that we don't currently support, mainly the collection routines. This is why I'm writing a separate e2e test. At the end of the development cycle we should be able to run `process()`.

ACK, sounds good! Go ahead and squash.

daphne/src/testing.rs
nakatsuka-y: 😄

cjpatton (approved):

PR #57, opened 2022-07-22
Implement `get_agg_share()` and test `http_post_aggregate_share()`

nakatsuka-y:
This PR implements necessary methods to test the aggregate share request/response procedure between leader/helper.

It also add a unit test that tests when a leader obtains an empty aggregate share.
The e2e unit test add the procedure where the Leader picks a collect job, produces its encrypted aggregate share, and lastly prepares an AggregateShareReq and sends it to the Helper to obtain the helper's encrypted aggregate share.

This partially implements #20, in particular the beginning of the Collection phase.

### DIFF
--- a/daphne/src/roles_test.rs
+++ b/daphne/src/roles_test.rs
@@ -9,9 +9,9 @@ use crate::{
     },
     hpke::HpkeReceiverConfig,
     messages::{
-        AggregateContinueReq, AggregateInitializeReq, AggregateResp, AggregateShareReq, CollectReq,
-        HpkeCiphertext, Id, Interval, Nonce, Report, ReportShare, Transition, TransitionFailure,
-        TransitionVar,
+        AggregateContinueReq, AggregateInitializeReq, AggregateResp, AggregateShareReq,
+        AggregateShareResp, CollectReq, HpkeCiphertext, Id, Interval, Nonce, Report, ReportShare,
+        Transition, TransitionFailure, TransitionVar,
     },
     roles::{DapAggregator, DapHelper, DapLeader},
     testing::{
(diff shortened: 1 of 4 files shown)


daphne/src/testing.rs
```diff
@@ -221,18 +221,67 @@ impl DapAggregator<BearerToken> for MockAggregator {
...
-        _batch_interval: &Interval,
+        task_id: &Id,
+        batch_interval: &Interval,
     ) -> Result<DapAggregateShare, DapError> {
-        unreachable!("not implemented");
+        let task_config = self
+            .get_task_config_for(task_id)
+            .ok_or_else(|| DapError::fatal("no task found"))?;
+
+        let mut agg_share = DapAggregateShare::default();
```
cjpatton, line 231: nit: For style I would move this declaration just before `agg_share` is first used. (This is the `for` loop below.)

daphne/src/testing.rs
```diff
@@ -221,18 +221,67 @@ impl DapAggregator<BearerToken> for MockAggregator {
...
+            window: batch_interval.start,
+        };
+        // TODO(nakatsuka-y) Should the batch window be inclusive?
+        while bucket_info.window < batch_interval.end() {
+            if let Some(delta) = agg_store.get_mut(&bucket_info) {
+                agg_share.merge(delta.clone())?;
+            }
+
+            bucket_info.window += task_config.min_batch_duration;
+        }
```
cjpatton, line 252: This is extremely inefficient if the keys for agg_store are sparse. What about iterate over the map? Something like
```rust
for (inner_bucket_Info, agg_share_deltaj) in agg_store.iter() {
      if bucket_info.task_id == inner_bucket_info.task_id &&
         bucket_Info.start <= inner_bucket_info.start &&
         bucket_info.start + bucket_Info.duration > inner_bucket_Info.start {
               // Add delta into `agg_share`
      }
}
```

daphne/src/testing.rs
```diff
@@ -221,18 +221,67 @@ impl DapAggregator<BearerToken> for MockAggregator {
...
+            .report_store
+            .lock()
+            .map_err(|e| DapError::Fatal(e.to_string()))?;
+        let report_store = report_store_mutex_guard.deref_mut();
+
+        let mut bucket_info = BucketInfo {
+            task_id: task_id.clone(),
+            window: batch_interval.start,
+        };
+        while bucket_info.window < batch_interval.end() {
```
cjpatton, line 276: Here again it's better to iterate over `report_store` than it is to iterate over batch windows.

cjpatton (changes requested):
The `while bucket_info.window < batch_interval.end() {..}` pattern is not very efficient if the keys in `agg_store` are sparse. (Similarly for `report_store`.) In general, the number of iterations in a loop should not depend on the input but the size of the data structure. 

I've suggested an alternative pattern in-line. Can you go through the rest of the module and fix accordingly?

daphne/src/roles_test.rs
```diff
@@ -486,6 +486,22 @@ async fn get_reports_empty_response() {
     assert_eq!(reports.len(), 0);
 }
 
+#[tokio::test]
```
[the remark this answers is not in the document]
nakatsuka-y: I wasn't paying too much attention when I was writing this test.
I've deleted it.

daphne/src/testing.rs
nakatsuka-y: Great catch! I've moved this just before the for loop.

daphne/src/testing.rs
nakatsuka-y: Agreed. I've changed the while loop to iterate over the records in the HashMap.

daphne/src/testing.rs
nakatsuka-y: See above.

cjpatton (commented):
LGTM. Go ahead and squash

cjpatton (approved):

Issue #25, opened 2022-06-24
Scaling up the Leader


cjpatton:
Aggregation and collect jobs are driven by a call to `DapLeader::process()`. This method is quite simple: It starts a single aggregation job, waits for it to finish, then completes any collect job waiting in the queue that is ready to complete. In general, it's not safe to run multiple instances of `process()` simultaneously for a given task. (It can be made safe by choosing the value of the `ReportSelector` carefully.)

The `process()` method is aimed primarily at driving tests. If used in a large deployment, it's likely to lead to a bottleneck. This can be avoided by running multiple aggregation jobs simultaneously. However, `DapLeader` will have to be refactored to support this.

PR #58, opened 2022-07-22
Move `task_id` out of `DapLeader::get_reports()`

cjpatton:
Partially addresses #25.

Moves `task_id` out of the function parameters. Existing implementors of
`DapLeader` can put this field intp the report selector. This allows for
deployments in which work is scheduled not per task but for the entire
deployment. In particular, this unblocks progress on scaling up
Daphne-Worker.

### DIFF
--- a/daphne/src/roles.rs
+++ b/daphne/src/roles.rs
@@ -147,13 +147,12 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
     /// Store a report for use later on.
     async fn put_report(&self, report: &Report) -> Result<(), DapError>;
 
-    /// Fetch a sequence of reports to aggregate. The reports returned are removed from persistent
-    /// storage.
+    /// Fetch a sequence of reports to aggregate, grouped by task ID. The reports returned are
+    /// removed from persistent storage.
     async fn get_reports(
         &self,
-        task_id: &Id,
         selector: &Self::ReportSelector,
-    ) -> Result<Vec<Report>, DapError>;
+    ) -> Result<HashMap<Id, Vec<Report>>, DapError>;
 
     /// Create a collect job.
     //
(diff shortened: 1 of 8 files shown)


daphne/src/roles.rs
```diff
@@ -239,148 +238,176 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
         Ok(self.init_collect_job(&collect_req).await?)
     }
```
cjpatton, line 240: Reviewer note: This is copy-pasted from `process()` below. (Similarly for `run_collect_job()` below.)

cjpatton:
Reviewer note: There should be no functional changes, *except* in Daphne-Worker:
```
/internal/process/task/:task_id
```
has been changed to
```
/internal/process
```
The task ID is now carried in the body of the request.

daphne/src/roles_test.rs
```diff
@@ -25,6 +25,16 @@ use prio::codec::{Decode, Encode};
 use rand::{thread_rng, Rng};
 use std::{ops::DerefMut, vec};
 
+// MockAggregator's implementation of DapLeader::get_report() always returns reports for a single
+// task. This macro is used to conveniently unwrap the task ID and reports for testing purposes.
+macro_rules! get_reports {
+    ($leader:expr, $selector:expr) => {{
+        let reports_per_task = $leader.get_reports($selector).await.unwrap();
+        assert_eq!(reports_per_task.len(), 1);
```
nakatsuka-y, line 33: I wonder whether we really need this assertion here.

nakatsuka-y (commented):
I think the changes look good, other than the one comment I put in `roles_test.rs`.

armfazh (approved):
this looks cleaner, so no need to represent taskIDs in a url-safe encoding. (maybe for DOs, but this is specific for workers)

daphne/src/roles_test.rs
cjpatton: As mentioned, this is here to validate an assumption that users of `get_reports!` make, which is that there is one and only one task.

nakatsuka-y (approved):

PR #59, opened 2022-07-23
daphne_worker: Process reports in order of arrival

cjpatton:
Based on #58 (merge that first).
Partially addresses #25.

Change report selector for `DaphneWorkerConfig` so that instead of
scheduling work per task and batch interval, reports are processed in
order of arrival.

A new DO is implemented, `LeaderAggregationJobQueue`, which maintaines
a queue of `ReportStore` instances. Once a report arrrives to an empty
`ReportStore` instance, an "aggregation job" is added to the queue; once
the instance becomes empty, the job is removed from the queue. Jobs are
ordered by the first report's time of arrival so that the oldest data is
processed first.

Other changes:

* daphne: Remove `task_id` from `DapLeader::get_pending_collect_jobs()`
  so that scheduling work on collect jobs doesn't have to depend on a
  particular task.

* Remove a TODO that was addressed by a previous change.

### DIFF
--- a/daphne/src/roles.rs
+++ b/daphne/src/roles.rs
@@ -166,11 +166,9 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
         collect_id: &Id,
     ) -> Result<DapCollectJob, DapError>;
 
-    /// Return the sequence of pending collect jobs for a given task, in order of priority.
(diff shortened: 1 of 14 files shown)


armfazh (approved):

cjpatton:
Rebased on top of @nakatsuka-y's last PR.

PR #60, opened 2022-07-23
Various operational improvements for Daphne-Worker

cjpatton:
Based on #59 (merge that first)

Various operational improvements in preparation for deployment:

*    daphne_worker: Move secrests out of environment variables

*    daphne_worker: Put test endpoints behind a flag

*    daphne_worker: Ensure non-zero minimum agg rate

    Adds a test that checks that each run of `DapLeader::process()`
    processes at least one report as long as there are reports in the job
    queue. Lo and behold, this test failed. Fix it.

*    daphne_worker: Align APIs for agg and collection job queues

    While at it, remove `daphen_worker/wrangler.toml`, which is no longer
    needed.


daphne/src/testing.rs
```diff
@@ -24,11 +24,19 @@ use std::{
...
 pub const TASK_LIST: &str = r#"{
     "f285be3caf948fcfc36b7d32181c14db95c55f04f55a2db2ee439c5879264e1f": {
         "leader_url": "https://leader.biz/leadver/v1/",
         "helper_url": "http://helper.com:8788",
-        "collector_hpke_config": "f40020000100010020a761d90c8c76d3d76349a3794a439a1572ab1fb8f13531d69744c92ea7757d7f",
+        "collector_hpke_config": {
+            "id": 23,
+            "kem_id": "X25519HkdfSha256",
+            "kdf_id": "HkdfSha256",
+            "aead_id": "Aes128Gcm",
```
cjpatton, line 37: @armfazh let me konw if you want to make this snake case in this PR.

armfazh (approved):
LGTM

daphne/src/testing.rs
armfazh: whatever JSON standard recommends

cjpatton: Folks seem split: https://stackoverflow.com/questions/5543490/json-naming-convention-snake-case-camelcase-or-pascalcase

I think the main thing is to be consistent. I'll raise an issue for this.

PR #61, opened 2022-07-25
Implement LeaderState and Collect job related methods

nakatsuka-y:
Implements final missing pieces of #20, in particular methods used by the Leader to handle collect jobs.

This PR implements LeaderState, which stores collect jobs it received from the Collector.
Collect job related methods, `init_collect_job`, `poll_collect_job`, `get_pending_collect_jobs`, and `finish_collect_job` are also implemented.

The full e2e test is also implemented, starting from the Client submitting a report, the Leader and Helper processing it, and finally the Leader processing the collect job received from the Collector.

### DIFF
--- a/daphne/Cargo.toml
+++ b/daphne/Cargo.toml
@@ -37,4 +37,5 @@ reqwest = { version = "0.11.11", features = ["blocking"] } # Required for daph C
 anyhow = "1.0.58" # Required for daph CLI
 
 [dev-dependencies]
+matchit = "0.6.0"
 tokio = { version = "^1.19", features = ["macros","rt"] }
--- a/daphne/src/roles_test.rs
+++ b/daphne/src/roles_test.rs
@@ -10,17 +10,19 @@ use crate::{
     hpke::HpkeReceiverConfig,
     messages::{
         AggregateContinueReq, AggregateInitializeReq, AggregateResp, AggregateShareReq,
-        AggregateShareResp, CollectReq, HpkeCiphertext, Id, Interval, Nonce, Report, ReportShare,
-        Transition, TransitionFailure, TransitionVar,
+        AggregateShareResp, CollectReq, CollectResp, HpkeCiphertext, Id, Interval, Nonce, Report,
+        ReportShare, Transition, TransitionFailure, TransitionVar,
     },
     roles::{DapAggregator, DapHelper, DapLeader},
     testing::{
-        BucketInfo, MockAggregateInfo, MockAggregator, ReportStore, HPKE_RECEIVER_CONFIG_LIST,
-        LEADER_BEARER_TOKEN,
+        BucketInfo, MockAggregateInfo, MockAggregator, ReportStore, COLLECTOR_BEARER_TOKEN,
+        HPKE_RECEIVER_CONFIG_LIST, LEADER_BEARER_TOKEN,
     },
-    DapAbort, DapError, DapLeaderTransition, DapMeasurement, DapRequest, Prio3Config, VdafConfig,
+    DapAbort, DapCollectJob, DapError, DapLeaderTransition, DapMeasurement, DapRequest,
+    Prio3Config, VdafConfig,
 };
 use assert_matches::assert_matches;
+use matchit::Router;
 use prio::codec::{Decode, Encode};
 use rand::{thread_rng, Rng};
 use std::{ops::DerefMut, vec};
@@ -499,6 +501,112 @@ async fn get_reports_empty_response() {
     assert_eq!(reports.len(), 0);
 }
 
+#[tokio::test]
+async fn poll_collect_job_test_results() {
+    let leader = MockAggregator::new();
+    let task_id = leader.nominal_task_id();
+    let task_config = leader.get_task_config_for(task_id).unwrap();
+    let now = 1637361337;
+
+    // Collector: Create a CollectReq.
+    let collector_collect_req = CollectReq {
+        task_id: task_id.clone(),
+        batch_interval: task_config.current_batch_window(now),
+        agg_param: Vec::default(),
+    };
+    let req = DapRequest {
+        media_type: Some(MEDIA_TYPE_COLLECT_REQ),
+        payload: collector_collect_req.get_encoded(),
+        url: task_config.helper_url.join("/collect").unwrap(),
+        sender_auth: Some(BearerToken::from(COLLECTOR_BEARER_TOKEN.to_string())),
+    };
+
+    // Leader: Handle the CollectReq received from Collector.
+    leader.http_post_collect(&req).await.unwrap();
+
+    // Expect DapCollectJob::Unknown due to invalid collect ID.
+    assert_eq!(
+        leader
+            .poll_collect_job(task_id, &Id::default())
+            .await
+            .unwrap(),
+        DapCollectJob::Unknown
+    );
+
+    // Leader: Get pending collect job to obtain collect_id
+    let resp = leader.get_pending_collect_jobs(&task_id).await.unwrap();
+    let collect_id = resp.keys().collect::<Vec<&Id>>()[0];
+    let collect_resp = CollectResp {
+        encrypted_agg_shares: Vec::default(),
+    };
+
+    // Expect DapCollectJob::Pending due to pending collect job.
+    assert_eq!(
+        leader.poll_collect_job(task_id, &collect_id).await.unwrap(),
+        DapCollectJob::Pending
+    );
+
+    // Leader: Complete the collect job by storing CollectResp in LeaderStore.processed.
+    leader
+        .finish_collect_job(&task_id, &collect_id, &collect_resp)
+        .await
+        .unwrap();
+
+    // Expect DapCollectJob::Done due to processed collect job.
+    assert_matches!(
+        leader.poll_collect_job(task_id, &collect_id).await.unwrap(),
+        DapCollectJob::Done(..)
+    );
+}
+
+// Test a successful collect request submission.
+// This checks that the Leader reponds with the collect ID with the ID associated to the request.
+//
+// TODO(nakatsuka-y) Implement a test case when a Leader rejects a collect request.
+#[tokio::test]
+async fn http_post_collect_success() {
+    let leader = MockAggregator::new();
+    let task_id = leader.nominal_task_id();
+    let task_config = leader.get_task_config_for(task_id).unwrap();
+    let now = 1637361337;
+
+    // Collector: Create a CollectReq.
+    let collector_collect_req = CollectReq {
+        task_id: task_id.clone(),
+        batch_interval: task_config.current_batch_window(now),
+        agg_param: Vec::default(),
+    };
+    let req = DapRequest {
+        media_type: Some(MEDIA_TYPE_COLLECT_REQ),
+        payload: collector_collect_req.get_encoded(),
+        url: task_config.helper_url.join("/collect").unwrap(),
+        sender_auth: Some(BearerToken::from(COLLECTOR_BEARER_TOKEN.to_string())),
+    };
+
+    // Leader: Handle the CollectReq received from Collector.
+    let url = leader.http_post_collect(&req).await.unwrap();
+    let mut resp = leader.get_pending_collect_jobs(&task_id).await.unwrap();
+    let leader_collect_id = resp.keys().collect::<Vec<&Id>>()[0].clone();
+    let leader_collect_req = resp.remove(&leader_collect_id).unwrap();
+
+    // Check that the CollectReq sent by Collector is the same that is received by Leader.
+    assert_eq!(collector_collect_req, leader_collect_req[0]);
+
+    // Check that the collect_id included in the URI is the same with the one received
+    // by Leader.
+    let path = url.path().to_string();
+    let mut router = Router::new();
+    router
+        .insert("/collect/task/:task_id/req/:collect_id", true)
+        .unwrap();
+    let url_match = router.at(&path).unwrap();
+    let collector_collect_id = url_match.params.get("collect_id").unwrap();
+    assert_eq!(
+        collector_collect_id.to_string(),
+        leader_collect_id.to_base64url()
+    );
+}
+
 // Test the upload sub-protocol.
 #[tokio::test]
 async fn successful_http_post_upload() {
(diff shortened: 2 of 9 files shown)


daphne/src/testing.rs
```diff
@@ -470,32 +474,134 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+            .leader_state_store
+            .lock()
+            .map_err(|e| DapError::Fatal(e.to_string()))?;
+        let leader_state_store = leader_state_store_mutex_guard.deref_mut();
+
+        // Check whether a collect job already exists for this task_id.
+        // If so, push CollectReq to LeaderState and retrieve the pre-computed collect URI.
+        // If not, create new LeaderState with new collect URI and CollectReq.
+        if leader_state_store.contains_key(&collect_req.task_id) {
+            let state = leader_state_store.get_mut(&collect_req.task_id).unwrap();
```
cjpatton, line 494: Alternatively:
```rust
if lest Some(state) = leader_state_store.contains_key(&collect_Req.task_id) {
    // use state
} else {
   // create state
}
```

daphne/src/roles_test.rs
```diff
@@ -578,21 +579,30 @@ async fn e2e() {
...
     };
+    let req = DapRequest {
+        media_type: Some(MEDIA_TYPE_COLLECT_REQ),
+        payload: collect_req.get_encoded(),
+        url: task_config.helper_url.join("/collect").unwrap(),
+        sender_auth: Some(BearerToken::from(COLLECTOR_BEARER_TOKEN.to_string())),
+    };
+
+    // Leader: Handle the CollectReq received from Collector.
+    leader.http_post_collect(&req).await.unwrap();
```
cjpatton, line 596: Test that the  collect URI encodes `collect_id`.

cjpatton (changes requested):

daphne/src/testing.rs
nakatsuka-y: Using this would've been nice, but there are restrictions in the code that would require me to write the same code twice, which I would like to avoid.

daphne/src/roles_test.rs
nakatsuka-y: I've implemented it in the new commits.

daphne/src/roles_test.rs
```diff
@@ -499,6 +501,49 @@ async fn get_reports_empty_response() {
     assert_eq!(reports.len(), 0);
 }
 
+// Test a successful collect request submission.
+// This checks that the Leader reponds with the correct collect ID.
+// TODO(nakatsuka-y) Implement a test case when a Leader rejects a collect request.
```
cjpatton, line 506: * There is no "correct ID", right? What we're checking here is that the collect URI uniquely identifies the collect request.
* nit: The first two lines should be in the same paragraph (i.e., remove newline between them).
* nit: I like to put a blank line between the comment and the TODO:

```rust
// A comment that describes the function.
//
// TODO(blah) Blah.
```

daphne/src/roles_test.rs
```diff
@@ -499,6 +501,49 @@ async fn get_reports_empty_response() {
...
+    let (leader_collect_id, leader_collect_req) =
+        leader.get_pending_collect_jobs(&task_id).await.unwrap()[0].clone();
+
+    // Check that the CollectReq sent by Collector is the same that is received by Leader.
+    assert_eq!(collector_collect_req, leader_collect_req);
+
+    // Check that the collect_id included in the URI is the same with the one received
+    // by Leader.
+    // TODO(nakatsuka-y) Use better way to parse the URI.
+    let parts = Url::parse(url.as_str()).expect("Unable to parse URL");
```
cjpatton, line 538: I think [matchit](https://docs.rs/matchit/latest/matchit/) would be a good choice for parsing this. See this example: https://docs.rs/matchit/latest/matchit/#named-parameters.

daphne/src/testing.rs
```diff
@@ -470,32 +474,156 @@ impl DapLeader<BearerToken> for MockAggregator {
...
     }
 
-    async fn init_collect_job(&self, _collect_req: &CollectReq) -> Result<Url, DapError> {
-        unreachable!("not implemented");
+    // Called after receiving a CollectReq from Collector.
+    async fn init_collect_job(&self, collect_req: &CollectReq) -> Result<Url, DapError> {
+        let mut rng = thread_rng();
+        let task_config = self
+            .get_task_config_for(&collect_req.task_id)
+            .ok_or_else(|| DapError::fatal("no task found"))?;
```
cjpatton, line 482: A bit clearer:
```suggestion
            .ok_or_else(|| DapError::fatal("task not found"))?;
```

daphne/src/testing.rs
```diff
@@ -470,32 +474,156 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+            .leader_state_store
+            .lock()
+            .map_err(|e| DapError::Fatal(e.to_string()))?;
+        let leader_state_store = leader_state_store_mutex_guard.deref_mut();
+
+        // Store Collect ID and CollectReq into LeaderState.
+        let collect_id = Id(rng.gen());
+        if !leader_state_store.contains_key(&collect_req.task_id) {
+            leader_state_store.insert(collect_req.task_id.clone(), LeaderState::new());
+        }
```
cjpatton, line 494: It should be possible to do this more efficiently with https://doc.rust-lang.org/stable/std/collections/hash_map/enum.Entry.html#method.or_insert_with.

daphne/src/testing.rs
```diff
@@ -470,32 +474,156 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+
+        let leader_state = leader_state_store.get_mut(&collect_req.task_id).unwrap();
+        leader_state.collect_ids.push_back(collect_id.clone());
+        let collect_job_state = CollectJobState {
+            collect_req: Some(collect_req.clone()),
```
cjpatton, line 499: It should be possible to avoid this clone. Hint: You need to borrow from `collect_req` below, but you can do that computation before getting here.

daphne/src/testing.rs
```diff
@@ -470,32 +474,156 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+            collect_resp: None,
+        };
+        leader_state
+            .collect_jobs
+            .insert(collect_id.clone(), collect_job_state);
```
cjpatton, line 504: This clone is also avoidable.

daphne/src/testing.rs
```diff
@@ -470,32 +474,156 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+
+        // Construct a new Collect URI for this CollectReq.
+        let collect_uri = task_config
```
cjpatton, line 507: This is not a valid URI: https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.2

daphne/src/testing.rs
```diff
@@ -470,32 +474,156 @@ impl DapLeader<BearerToken> for MockAggregator {
...
-        unreachable!("not implemented");
+        let mut leader_state_store_mutex_guard = self
+            .leader_state_store
+            .lock()
+            .map_err(|e| DapError::Fatal(e.to_string()))?;
+        let leader_state_store = leader_state_store_mutex_guard.deref_mut();
+
+        let leader_state = leader_state_store
+            .get(task_id)
+            .ok_or_else(|| DapError::fatal("cannot get collect job for task_id"))?;
```
cjpatton, line 533: A bit more clear
```suggestion
            .ok_or_else(|| DapError::fatal("collect job not found"))?;
```

daphne/src/testing.rs
```diff
@@ -470,32 +474,156 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+                .collect_resp
+                .is_none()
+            {
+                Ok(DapCollectJob::Pending)
+            } else {
+                Ok(DapCollectJob::Unknown)
+            }
+        } else {
+            Err(DapError::fatal("cannot get collect job for collect_id"))
+        }
```
cjpatton, line 555: Hmm, this doesn't look quite right. Is the outer `else` statement reachable? What about the inner `else`?

Taking a step back: Your goal here is to check if collect request is "finished", "pending", or "unknown". It might be helpful to try to first describe this algorithm to yourself in natural language, then try to transcribe that algorithm into Rust.

daphne/src/testing.rs
```diff
@@ -470,32 +474,156 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+            res.push((
+                collect_id.clone(),
+                leader_state
+                    .collect_jobs
+                    .get(&collect_id)
+                    .unwrap()
+                    .collect_req
+                    .as_ref()
+                    .unwrap()
+                    .clone(),
```
cjpatton, line 584: This long list of calls is what we call "code smell": It's highly likely that this code can be simplified in a way that makes it more readable.

cjpatton (changes requested):
Great start! Some comments in-line. Also, let's also add a test to ensure that http_post_collect() properly authorizes requests.

daphne/src/roles_test.rs
nakatsuka-y: You are correct. I've updated this comment and others based on your feedback.

daphne/src/roles_test.rs
nakatsuka-y: This works great! Thanks for suggesting.

daphne/src/testing.rs
nakatsuka-y: Sounds good. I've updated the error messages.

daphne/src/testing.rs
nakatsuka-y: Great suggestion, I've replaced the original code with `or_insert_with`.

daphne/src/testing.rs
nakatsuka-y: I've eliminated as much `clone()`s as possible.

daphne/src/testing.rs
nakatsuka-y: Ditto.

daphne/src/testing.rs
nakatsuka-y: I'm a bit confused with this comment. Could you elaborate further?
Should I rename URI to a different term?

daphne/src/testing.rs
nakatsuka-y: Ditto :)

daphne/src/testing.rs
nakatsuka-y: Thanks for the suggestion and based on your feedback during our call, I've fixed this.

I've also added a test case that tests all these return types.

daphne/src/testing.rs
nakatsuka-y: Agreed. The code looks much cleaner now.

daphne/Cargo.toml
```diff
@@ -32,6 +32,7 @@ serde = { version = "^1.0", features = ["derive"] }
 rand = "0.8.5"
 thiserror = "1.0"
 url = { version = "2.2.2", features = ["serde"] }
+matchit = "0.6.0"
```
cjpatton, line 35: "matchit" is only needed for tests, so it should go in "dev-dependencies" below. Otherwise it'll get built even if not needed.

daphne/src/testing.rs
cjpatton: This clone can be avoided by moving `collect_id.to_base64url()`, above this point and moving `collect_id` into `.insert()` here.

daphne/src/testing.rs
cjpatton: A URI is just a URL, it needs to be prefixed by a hostname. For example "http://example.com/thing" is a valid URI but "/thing" is not.

daphne/src/testing.rs
```diff
@@ -470,32 +474,126 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+            Ok(DapCollectJob::Unknown)
+        }
     }
 
+    // Called to retrieve pending CollectReq.
     async fn get_pending_collect_jobs(
         &self,
-        _task_id: &Id,
+        task_id: &Id,
     ) -> Result<Vec<(Id, CollectReq)>, DapError> {
```
cjpatton, line 543: Somewhat orthogonal to this PR: Do you think the return type should be `HashMap<Id, CollectReq>` instead of `Vec<(Id, CollectReq)>`, in order to match `get_reports()`?

Feel free to make this change in this PR if you concur. Otherwise, I will consider doing it later.

daphne/src/testing.rs
```diff
@@ -470,32 +474,126 @@ impl DapLeader<BearerToken> for MockAggregator {
...
+            .lock()
+            .map_err(|e| DapError::Fatal(e.to_string()))?;
+        let leader_state_store = leader_state_store_mutex_guard.deref_mut();
+
+        let leader_state = leader_state_store
+            .get_mut(task_id)
+            .ok_or_else(|| DapError::fatal("collect job not found for task_id"))?;
+        let mut res: Vec<(Id, CollectReq)> = Vec::default();
+
+        // Pop collect IDs from the queue and insert them and the associated requests to the response.
```
cjpatton, line 555: We don't want to remove the collect jobs since we don't necessarily consume them. (Take a look at `DapLeader::process()` to see why.) Instead we just want to copy them.

cjpatton (changes requested):

daphne/Cargo.toml
nakatsuka-y: Thanks for pointing this out.

daphne/src/testing.rs
nakatsuka-y: I've replaced `Vec<(Id, CollectReq)>` with `HashMap<Id, CollectReq>`.
Let me know what you think!

daphne/src/testing.rs
nakatsuka-y: I've moved the deletion process in `finish_collect_job` method.

daphne/src/testing.rs
cjpatton: Wait, did I miss that you're `join()`ing `leader_url` with the path? My mistake!

cjpatton (commented):
LGTM. Please squash and I'll approve.

cjpatton (approved):

Issue #45, opened 2022-07-13
Handle collect requests properly


cjpatton:
Daphne is currently missing https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/277, which is needed for privacy. However the exact requirements for collect requests (and constraints on them) are a bit of a moving target. (See https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/195.)

cjpatton:
(This isn't really a bug per se, it's more of a feature parity issue.)

cjpatton:
One thought I had here while reviewing #63: The Leader should reject collect requests for which:
* The batch interval is *huge*, in the sense that it spans more than, say, 100 time windows. Otherwise updating ReportStore buckets after a batch is aggregated will be expensive.
* The start of the batch interval is too far in the past, say more than a few days. Otherwise we might end up holding onto a request for which we're not likely to get any data.
* The end of the batch interval is too far in the future, say more than a few hours. Similar reasoning as above.

cjpatton:
Suggestion for how to proceed:
* Mark each AggreagteStore DO instance as collected or not collected.
* When a collect request comes in, for each batch window, check the AggreagteStore to see if it has been collected. If so, reject. Otherwise, if none have been collected, accept.

nakatsuka-y:
Closing as #63 has been merged to main branch.

cjpatton:
Ropened due to bug spotted after merging the main PR. This condition is wrong:
https://github.com/cloudflare/daphne/blob/main/daphne_worker/src/dap.rs#L358-L359

There is a similar issue on the Helper side.

Issue #54, opened 2022-07-21
Clean up: Remove timestamps from unit tests


nakatsuka-y:
Currently, the `get_reports` unit tests implemented in `roles_test.rs` are required to specify a timestamp  to get reports that fall within that time interval.

We should be able to remove this by implementing `get_reports` in a way that the function sweeps the `ReportStore` and automatically fetches the requested number of reports. See below.

> You simply iterate over the buckets the  corresponding to the task ID:
> ```rust
>          let mut reports = Vec::new();
>          for bucket_info in report_store.keys() {
>              if &bucket_info.task_id == task_id {
>                  // drain report_store.get(bucket_info),  break the `for` loop if `reports.len() ==
>                  // agg_rate`.`
>              }
>          }
> ```
> You might also consider doing it in sorted order by `bucket_info.batch_interval.start`. so that the oldest data is prioritized for aggregation.

_Originally posted by @cjpatton in https://github.com/cloudflare/daphne/pull/51#pullrequestreview-1046802846_

PR #63, opened 2022-07-27
Implement `CollectReq` boundary check

nakatsuka-y:
This PR implements a mechanism to handle `CollectReq` properly (#45).

A malicious Collector may request aggregation results over overlapping intervals.
This will  allow the malicious Collector to obtain aggregation results that are in batches smaller than `min_batch_interval`.

The Leader holds a binary tree for each Collector (distinguished via bearer token) per task.
The nodes in the binary tree hold the Interval requested by the Collector for this task in the past.
When the Leader receives a CollectReq from a Collector, it searches the binary tree for an overlapping node and if so, will reject the request.
If the new CollectReq does not overlap with the past intervals, then the Leader stores the new Interval into the tree and proceed to store the collect Id and request.

Unit test cases for the binary tree is written.
Another test case where the Collector tries to send a collect request with an overlapping interval is also implemented.

### DIFF
--- a/daphne/src/lib.rs
+++ b/daphne/src/lib.rs
@@ -48,6 +48,10 @@ pub enum DapError {
     #[error("fatal error: {0}")]
     Fatal(String),
 
+    /// Error triggered by peer, resulting in an abort.
+    #[error("abort: {0}")]
+    Abort(DapAbort),
+
     /// Transition failure. This error blocks processing of a paritcular report and may, under
     /// certain conditions, trigger an abort.
     #[error("transition error: {0}")]
@@ -108,6 +112,11 @@ pub enum DapAbort {
     #[error("batchMismatch")]
     BatchMismatch,
 
+    /// Batch overlap. Sent in response to an CollectReq for which the Leader detects the same
+    /// Collector requesting an aggregate share which it has collected in the past.
+    #[error("batchOverlap")]
+    BatchOverlap,
+
     /// Internal error.
     #[error("internalError: {0}")]
     Internal(#[source] Box<dyn std::error::Error + 'static + Send + Sync>),
@@ -162,6 +171,7 @@ impl DapAbort {
     pub fn to_problem_details(&self) -> ProblemDetails {
         let (typ, detail) = match self {
             Self::BatchMismatch
+            | Self::BatchOverlap
             | Self::InvalidBatchInterval
             | Self::InsufficientBatchSize
             | Self::ReplayedReport
@@ -188,6 +198,7 @@ impl From<DapError> for DapAbort {
     fn from(e: DapError) -> Self {
         match e {
             e @ DapError::Fatal(..) => Self::Internal(Box::new(e)),
+            DapError::Abort(e) => e,
             DapError::Transition(t) => Self::from(t),
         }
     }
@@ -223,6 +234,27 @@ pub struct ProblemDetails {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub detail: Option<String>,
+}
+
+/// Global DAP parameters common across tasks.
+//
+// NOTE: Consider whether the following parameters should be included
+// in the spec.
+#[derive(Deserialize, Serialize)]
+pub struct DapGlobalConfig {
+    /// Maximum window size permitted in CollectReq.
+    /// Prevents Collectors from requesting wide range or reports.
+    pub max_batch_duration: u64,
+
+    /// Lower bound of an acceptable batch interval for collect requests.
+    /// Batch intervals cannot start more than `min_batch_interval_start`
+    /// apart from the current batch interval.
+    pub min_batch_interval_start: u64,
+
+    /// Higher bound of an acceptable batch interval for collect requests.
+    /// Batch intervals cannot end more than `max_batch_interval_end`
+    /// apart from the current batch interval.
+    pub max_batch_interval_end: u64,
 }
 
 /// Per-task DAP parameters.
--- a/daphne/src/roles_test.rs
+++ b/daphne/src/roles_test.rs
@@ -19,13 +19,13 @@ use crate::{
         HPKE_RECEIVER_CONFIG_LIST, LEADER_BEARER_TOKEN,
     },
     DapAbort, DapCollectJob, DapError, DapLeaderTransition, DapMeasurement, DapRequest,
-    Prio3Config, VdafConfig,
+    DapTaskConfig, Prio3Config, VdafConfig,
 };
 use assert_matches::assert_matches;
 use matchit::Router;
 use prio::codec::{Decode, Encode};
 use rand::{thread_rng, Rng};
-use std::{ops::DerefMut, vec};
+use std::{ops::DerefMut, time::SystemTime, vec};
 
 // MockAggregator's implementation of DapLeader::get_report() always returns reports for a single
 // task. This macro is used to conveniently unwrap the task ID and reports for testing purposes.
@@ -144,16 +144,140 @@ impl MockAggregator {
 
         // Construct report.
         let vdaf_config: &VdafConfig = &VdafConfig::Prio3(Prio3Config::Count);
+        let now = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap()
+            .as_secs();
         let report = vdaf_config
-            .produce_report(
-                &hpke_config_list,
-                1637361337,
-                task_id,
-                DapMeasurement::U64(1),
-            )
+            .produce_report(&hpke_config_list, now, task_id, DapMeasurement::U64(1))
             .unwrap();
 
         report
+    }
+
+    async fn run_test_agg_job(
+        &self,
+        helper: &MockAggregator,
+        now: u64,
+        task_id: &Id,
+        task_config: &DapTaskConfig,
+    ) {
+        // Leader: Store received report to ReportStore.
+        let selector = &MockAggregateInfo {
+            task_id: task_id.clone(),
+            batch_info: Some(task_config.current_batch_window(now)),
+            agg_rate: 1,
+        };
+        let (task_id, reports) = get_reports!(self, selector);
+
+        // Leader: Consume report share.
+        let mut rng = thread_rng();
+        let agg_job_id = Id(rng.gen());
+        let transition = task_config
+            .vdaf
+            .produce_agg_init_req(
+                self,
+                &task_config.vdaf_verify_key,
+                &task_id,
+                &agg_job_id,
+                reports,
+            )
+            .unwrap();
+        assert_matches!(transition, DapLeaderTransition::Continue(..));
+        let (leader_state, agg_init_req) = transition.unwrap_continue();
+
+        // Leader: Send aggregate initialization request to Helper and receive response.
+        let req = DapRequest {
+            media_type: Some(MEDIA_TYPE_AGG_INIT_REQ),
+            payload: agg_init_req.get_encoded(),
+            url: task_config.helper_url.join("/aggregate").unwrap(),
+            sender_auth: Some(BearerToken::from(LEADER_BEARER_TOKEN.to_string())),
+        };
+        let res = helper.http_post_aggregate(&req).await.unwrap();
+        let agg_resp = AggregateResp::get_decoded(&res.payload).unwrap();
+
+        // Leader: Produce Leader output share and prepare aggregate continue request for Helper.
+        let transition = task_config
+            .vdaf
+            .handle_agg_resp(&task_id, &agg_job_id, leader_state, agg_resp)
+            .unwrap();
+        assert_matches!(transition, DapLeaderTransition::Uncommitted(..));
+        let (leader_uncommitted, agg_cont_req) = transition.unwrap_uncommitted();
+
+        // Leader: Send aggregate continue request to Helper and receive response.
+        let req = DapRequest {
+            media_type: Some(MEDIA_TYPE_AGG_CONT_REQ),
+            payload: agg_cont_req.get_encoded(),
+            url: task_config.helper_url.join("/aggregate").unwrap(),
+            sender_auth: Some(BearerToken::from(LEADER_BEARER_TOKEN.to_string())),
+        };
+        let res = helper.http_post_aggregate(&req).await.unwrap();
+        let agg_resp = AggregateResp::get_decoded(&res.payload).unwrap();
+
+        // Leader: Commit output shares of Leader and Helper.
+        let out_shares = task_config
+            .vdaf
+            .handle_final_agg_resp(leader_uncommitted, agg_resp)
+            .unwrap();
+        self.put_out_shares(&task_id, out_shares).await.unwrap();
+    }
+
+    async fn run_test_col_job(
+        &self,
+        task_id: &Id,
+        collect_id: &Id,
+        collect_req: &CollectReq,
+        task_config: &DapTaskConfig,
+    ) {
+        // Leader: Get Leader's encrypted aggregate share.
+        let leader_agg_share = self
+            .get_agg_share(&collect_req.task_id, &collect_req.batch_interval)
+            .await
+            .unwrap();
+
+        let leader_enc_agg_share = task_config
+            .vdaf
+            .produce_leader_encrypted_agg_share(
+                &task_config.collector_hpke_config,
+                &collect_req.task_id,
+                &collect_req.batch_interval,
+                &leader_agg_share,
+            )
+            .unwrap();
+
+        // Leader: Prepare AggregateShareReq.
+        let agg_share_req = AggregateShareReq {
+            task_id: collect_req.task_id.clone(),
+            batch_interval: collect_req.batch_interval.clone(),
+            agg_param: collect_req.agg_param.clone(),
+            report_count: leader_agg_share.report_count,
+            checksum: leader_agg_share.checksum,
+        };
+
+        // Leader: Send AggregateShareReq to Helper and receive AggregateShareResp.
+        let req = DapRequest {
+            media_type: Some(MEDIA_TYPE_AGG_SHARE_REQ),
+            payload: agg_share_req.get_encoded(),
+            url: task_config.helper_url.join("/aggregate_share").unwrap(),
+            sender_auth: Some(BearerToken::from(LEADER_BEARER_TOKEN.to_string())),
+        };
+        let res = self.http_post_aggregate_share(&req).await.unwrap();
+        let agg_share_resp = AggregateShareResp::get_decoded(&res.payload).unwrap();
+        let helper_enc_agg_share = agg_share_resp.encrypted_agg_share;
+
+        // Leader: Complete the collect job by storing CollectResp in LeaderStore.processed.
+        let collect_resp = CollectResp {
+            encrypted_agg_shares: vec![leader_enc_agg_share, helper_enc_agg_share],
+        };
+
+        self.finish_collect_job(task_id, collect_id, &collect_resp)
+            .await
+            .unwrap();
+
+        // Leader: Mark the reports as collected.
+        self.mark_collected(task_id, &agg_share_req.batch_interval)
+            .await
+            .unwrap();
     }
 }
 
@@ -237,9 +361,13 @@ async fn http_post_collect_unauthorized_request() {
 #[tokio::test]
 async fn http_post_aggregate_failure_hpke_decrypt_error() {
     let helper = MockAggregator::new();
+    let now = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .unwrap()
+        .as_secs();
     let report_shares = vec![ReportShare {
         nonce: Nonce {
-            time: 1637361337,
+            time: now,
             rand: [1; 16],
         },
         extensions: Vec::default(),
@@ -477,7 +605,7 @@ async fn get_reports_empty_response() {
         .expect("upload failed unexpectedly");
 
     // Get report.
-    let now = report.nonce.time - 200000000;
+    let now = report.nonce.time - task_config.min_batch_duration - 1;
     let selector = &MockAggregateInfo {
         task_id: task_id.clone(),
         batch_info: Some(task_config.current_batch_window(now)),
@@ -489,7 +617,7 @@ async fn get_reports_empty_response() {
     assert_eq!(reports.len(), 0);
 
     // Attempt to get reports from the future.
-    let now = report.nonce.time + 200000000;
+    let now = report.nonce.time + task_config.min_batch_duration + 1;
     let selector = &MockAggregateInfo {
         task_id: task_id.clone(),
         batch_info: Some(task_config.current_batch_window(now)),
@@ -506,7 +634,10 @@ async fn poll_collect_job_test_results() {
     let leader = MockAggregator::new();
     let task_id = leader.nominal_task_id();
     let task_config = leader.get_task_config_for(task_id).unwrap();
-    let now = 1637361337;
+    let now = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .unwrap()
+        .as_secs();
 
     // Collector: Create a CollectReq.
     let collector_collect_req = CollectReq {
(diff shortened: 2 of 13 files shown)


cjpatton (commented):
Based on a quick look and reading the top comment this seems like a great approach! I suspect that O(log n) is pretty close to optimal runtime for this check. A few comments before diving in:
1. Enforcement per collector (i.e., bearer token) is not the right level of abstraction. The collector authentication mechanism is one piece of the protocol that is likely to change, so we'll need to decouple collect request enforcement from bearer tokens. I think enforcement per task would be better.
2. This mechanism is going to have to be implemented by both MockAggregator and DaphneWorkerConfig. Because we have to implement it twice and get good test coverage for both: (a) we need to maximize the amount of code shared by both; and (b) the pieces that must be implemented separately by both should be as simple as possible.
3. Bare in mind that it's more important that DaphneWorkerConfig is efficient than MockAggreegator.

daphne_worker/src/durable/aggregate_store.rs
```diff
@@ -52,9 +61,21 @@ impl DurableObject for AggregateStore {
             }
 
             (DURABLE_AGGREGATE_STORE_GET, Method::Post) => {
-                let agg_share: DapAggregateShare =
-                    state_get_or_default(&self.state, "agg_share").await?;
-                Response::from_json(&agg_share)
+                let collected: bool = state_get_or_default(&self.state, "collected").await?;
+                if !collected {
+                    let agg_share: DapAggregateShare =
+                        state_get_or_default(&self.state, "agg_share").await?;
```
cjpatton, line 67: It's posible to get `collected` and `agg_share` concurrently, which would save on latency:
```rust
let agg_share_future = state_get_or_default(&self.state, "agg_share");
let collected: bool = state_get_or_default(&self.state, "collected").await?;
if !collected {
    let agg_sahre: DapAggregateShare = agg_share_future.await?;
}
```

cjpatton (changes requested):
Looking good! I've only looked at Daphne-Worker so far.

daphne/src/auth.rs
```diff
@@ -17,7 +17,7 @@ use serde::{Deserialize, Serialize};
 use std::io::Cursor;
 
 /// A bearer token used for authorizing DAP requests as specified in draft-ietf-ppm-dap-01.
-#[derive(Clone, Debug, Deserialize, Serialize)]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Hash, Serialize)]
```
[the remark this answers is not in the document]
nakatsuka-y: This was from a past change. Thanks for pointing it out, I've removed it.

daphne/src/messages.rs
```diff
@@ -288,6 +288,7 @@ pub enum TransitionFailure {
     HpkeUnknownConfigId = 3,
     HpkeDecryptError = 4,
     VdafPrepError = 5,
+    InvalidBatchInterval = 6,
```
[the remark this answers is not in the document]
nakatsuka-y: I've removed this and replaced with `DapError::PeerError`.

daphne_worker/src/dap.rs
```diff
@@ -192,6 +193,7 @@ impl<D> DapAggregator<BearerToken> for DaphneWorkerConfig<D> {
         let namespace = self.durable_object("DAP_AGGREGATE_STORE")?;
         let task_id_base64url = task_id.to_base64url();
         let mut agg_share = DapAggregateShare::default();
+        // TODO(nakatsuka-y) Restrict the batch_interval or the following iteration will be expensive.
```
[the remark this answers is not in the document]
nakatsuka-y: I've added changes to incorporate this check.

daphne_worker/src/dap.rs
```diff
@@ -200,13 +202,13 @@ impl<D> DapAggregator<BearerToken> for DaphneWorkerConfig<D> {
...
-                    .await?
-                    .json()
-                    .await?;
+            let mut resp = durable_post!(stub, DURABLE_AGGREGATE_STORE_GET, &()).await?;
 
-            agg_share.merge(agg_share_delta)?;
+            // If this agg share has been collected before, return BatchCollected error.
+            match resp.json().await? {
+                AggregateStoreResult::Ok(agg_share_delta) => agg_share.merge(agg_share_delta)?,
+                AggregateStoreResult::Err(t) => return Err(DapError::Transition(t)),
```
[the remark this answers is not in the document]
nakatsuka-y: Changed this to `DapError::PeerError`

daphne_worker/src/durable/aggregate_store.rs
```diff
@@ -14,6 +15,14 @@ pub(crate) fn durable_agg_store_name(task_id_base64url: &str, window: u64) -> St
 
 pub(crate) const DURABLE_AGGREGATE_STORE_GET: &str = "/internal/do/aggregate_store/get";
 pub(crate) const DURABLE_AGGREGATE_STORE_MERGE: &str = "/internal/do/aggregate_store/merge";
+pub(crate) const DURABLE_AGGREGATE_STORE_MARK_COLLECTED: &str =
+    "/internal/do/aggregate_store/mark_collected";
+
+#[derive(Deserialize, Serialize)]
+pub(crate) enum AggregateStoreResult {
+    Ok(DapAggregateShare),
+    Err(TransitionFailure),
```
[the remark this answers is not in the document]
nakatsuka-y: Ditto :)

daphne_worker/src/durable/aggregate_store.rs
```diff
@@ -52,9 +61,21 @@ impl DurableObject for AggregateStore {
             }
 
             (DURABLE_AGGREGATE_STORE_GET, Method::Post) => {
-                let agg_share: DapAggregateShare =
-                    state_get_or_default(&self.state, "agg_share").await?;
-                Response::from_json(&agg_share)
+                let collected: bool = state_get_or_default(&self.state, "collected").await?;
```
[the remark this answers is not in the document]
nakatsuka-y: Added note :)

daphne_worker/src/durable/aggregate_store.rs
nakatsuka-y: Thanks for the suggestion!

daphne/src/lib.rs
```diff
@@ -48,6 +48,10 @@ pub enum DapError {
     #[error("fatal error: {0}")]
     Fatal(String),
 
+    /// Error caused due to messages from a peer.
+    #[error("peer error: {0}")]
+    PeerError(String),
```
cjpatton, line 53: If you do this instead, you can get rid of the hard-coded stuff. By the way, you're going to need to add a new abort type: See the changes in pr #297 on the DAP spec.
```suggestion
    PeerError(DapAbort),
```

daphne/src/lib.rs
```diff
@@ -227,13 +238,19 @@ pub struct ProblemDetails {
...
+// TODO spec: `max_window_size`, `min_duration_start`, and `max_duration_end`
+//            should be reflected to the DAP spec.
 #[derive(Deserialize, Serialize)]
 #[serde(try_from = "ShadowDapTaskConfig")]
 pub struct DapTaskConfig {
     pub leader_url: Url,
     pub helper_url: Url,
     pub min_batch_duration: u64, // seconds
     pub min_batch_size: u64,     // number of reports
+    pub max_window_size: u64,    // number of acceptable batches of reports
```
cjpatton, line 251: Let's do this instead
```suggestion
    /// Number of acceptable batches of reports.
    pub max_window_size: u64,
```

Also, this comment is not very descriptive. It's especially important to be clear here since this isn't in the spec.

daphne/src/roles_test.rs
```diff
@@ -144,13 +144,12 @@ impl MockAggregator {
 
         // Construct report.
         let vdaf_config: &VdafConfig = &VdafConfig::Prio3(Prio3Config::Count);
+        let now = SystemTime::now()
```
cjpatton, line 147: Can we avoid using the system time for this test case? See #54.

daphne_worker/src/durable/aggregate_store.rs
```diff
@@ -61,14 +62,16 @@ impl DurableObject for AggregateStore {
...
+                    state_get_or_default(&self.state, "agg_share").await?;
                 let collected: bool = state_get_or_default(&self.state, "collected").await?;
                 if !collected {
-                    let agg_share: DapAggregateShare =
-                        state_get_or_default(&self.state, "agg_share").await?;
                     Response::from_json(&AggregateStoreResult::Ok(agg_share))
                 } else {
                     Response::from_json(&AggregateStoreResult::Err(
-                        TransitionFailure::InvalidBatchInterval,
+                        PEER_ERR_BATCH_OVERLAP.to_string(),
```
cjpatton, line 74: This should be a `DapAbort` variant rather than a string.

cjpatton (changes requested):

cjpatton:
High level comment: I think we should move the new max_window_size, ... parameters out of VdafConfig. They should be picked globally per deployment, not per task.

daphne/src/lib.rs
nakatsuka-y: We are now using DapAbort instead of String.

daphne/src/lib.rs
nakatsuka-y: Added a more descriptive comment and moved the three variables out of task_config and into their own config struct.

daphne_worker/src/durable/aggregate_store.rs
nakatsuka-y: Since we cannot use DapAbort (we cannot attach `Serialize`, `Deserialize` which is necessary to use it in `from_json`), we defined a new enum `ErrBatchOverlap`.

daphne/src/lib.rs
```diff
@@ -225,6 +236,45 @@ pub struct ProblemDetails {
     pub detail: Option<String>,
 }
 
+/// Global DAP parameters common across tasks.
+//
+// TODO spec: `max_window_size`, `min_duration_start`, and `max_duration_end`
+//            should be reflected to the DAP spec.
+#[derive(Deserialize, Serialize)]
+#[serde(try_from = "ShadowDapGlobConfig")]
```
cjpatton, line 244: This shadow struct isn't needed because we don't have any fields that require custom serialization.
```suggestion
```

daphne/src/lib.rs
```diff
@@ -225,6 +236,45 @@ pub struct ProblemDetails {
...
+pub struct DapGlobConfig {
+    /// Maximum window size permitted in CollectReq.
+    /// Prevents Collectors from requesting wide range or reports.
+    pub max_window_size: u64,
```
cjpatton, line 248: I think a better name would be "max_batch_duration", as it aligns better with "min_batch_duration".

daphne/src/lib.rs
```diff
@@ -225,6 +236,45 @@ pub struct ProblemDetails {
...
+
+    /// Lower bound of an acceptable batch window in CollectReq.
+    /// Batch intervals must start after `min_duration_start`.
+    pub min_duration_start: u64,
```
cjpatton, line 252:
```suggestion
    pub min_batch_interval_start: u64,
```

daphne/src/lib.rs
```diff
@@ -225,6 +236,45 @@ pub struct ProblemDetails {
...
+
+    /// Higher bound of an acceptable batch window in CollectReq.
+    /// Batch intervals must end before `max_duration_end`.
+    pub max_duration_end: u64,
```
cjpatton, line 256: Is this a duration (i.e., a number of seconds) or a timestamp (i.e., number of seconds after UNIX epoch)?

daphne/src/lib.rs
```diff
@@ -225,6 +236,45 @@ pub struct ProblemDetails {
...
+}
+
+#[derive(Deserialize, Serialize)]
+struct ShadowDapGlobConfig {
```
cjpatton, line 260: "Glob" is a word in English that means "a lump of a semiliquid substance" :)
```suggestion
struct ShadowDapGlobalConfig {
```

daphne_worker/src/config.rs
```diff
@@ -60,6 +61,10 @@ pub(crate) struct DaphneWorkerConfig<D> {
 impl<D> DaphneWorkerConfig<D> {
     /// Fetch DAP parameters from environment variables.
     pub(crate) fn from_worker_context(ctx: RouteContext<D>) -> Result<Self> {
+        let glob_config: DapGlobConfig =
```
cjpatton, line 64:
```suggestion
        let global_config: DapGlobalConfig =
```

daphne/src/lib.rs
```diff
@@ -48,6 +48,10 @@ pub enum DapError {
     #[error("fatal error: {0}")]
     Fatal(String),
 
+    /// Error caused due to messages from a peer.
+    #[error("peer error: {0}")]
+    PeerError(DapAbort),
```
cjpatton, line 53:
```suggestion
    /// Error triggered by peer, resulting in an abort.
    #[error("abort: {0}")]
    Abort(DapAbort),
```

daphne_worker/src/dap.rs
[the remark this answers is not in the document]
cjpatton: Great, in that case, pleases remove the comment.

daphne_worker/src/lib.rs
```diff
@@ -14,6 +14,12 @@ use prio::codec::{Decode, Encode};
 use serde::{Deserialize, Serialize};
 use worker::*;
 
+/// Error Parameters used by the Leader to report errors caused by peers.
+#[derive(Deserialize, Serialize)]
+pub(crate) enum PeerErr {
+    ErrBatchOverlap,
+}
```
cjpatton, line 21: I think I see what you're getting at here: There may be other reasons for a DO to signal an abort, and this would be the enum that defines them. However I think trying to be general here is cumbersome and makes internal APIs a little awkward. My preference would be to do away with this enum here and modify the enum in aggregate_store.rs as follows:

```rust
#[derive(Deserialize, Serialize)]
pub(crate) enum AggregateStoreResult {
    Ok(DapAggregateShare),
    ErrBatchOverlap,
}
```

This saves the caller of the AggregateStore DO from having to handle undefined behavior (e.g., suppose AggregateStore returned an error variant other than ErrBatchOverlap).

daphne_worker_test/helper.env
```diff
@@ -16,6 +16,9 @@ DAP_BUCKET_KEY = 'f79c352056982bae1737e34bdac24d63'
 # Number of buckets.
 DAP_BUCKET_COUNT = 2
 
+# Global configurations.
+GLOB_CONFIG = '{"max_window_size": 100,"min_duration_start": 432000,"max_duration_end": 18000}'
```
cjpatton, line 20: Here and below
```suggestion
GLOBAL_CONFIG = '{"max_window_size": 100,"min_duration_start": 432000,"max_duration_end": 18000}'
```

daphne_worker_test/tests/e2e.rs
```diff
@@ -537,3 +540,94 @@ async fn e2e_leader_collect_abort_invalid_batch_interval() {
     )
     .await;
 }
+
+#[tokio::test]
+#[cfg_attr(not(feature = "test_e2e"), ignore)]
+async fn e2e_leader_collect_abort_overlapping_batch_interval() {
+    let t = TestRunner::default().await;
+    let batch_interval = t.batch_interval();
+
```
cjpatton, line 549: nit: Remove this newline

daphne_worker_test/tests/e2e.rs
```diff
@@ -537,3 +540,94 @@ async fn e2e_leader_collect_abort_invalid_batch_interval() {
...
+    assert_eq!(
+        agg_telem.reports_aggregated, t.min_batch_size,
+        "reports aggregated"
+    );
+    assert_eq!(
+        agg_telem.reports_collected, t.min_batch_size,
+        "reports collected"
+    );
+
+    // Send a collect request that requests results from batches that overlaps with
```
cjpatton, line 603:
```suggestion
    // Send a collect request that overlaps with
```

daphne_worker_test/tests/e2e.rs
```diff
@@ -537,3 +540,94 @@ async fn e2e_leader_collect_abort_invalid_batch_interval() {
...
+        .internal_process(
+            &client,
+            &InternalAggregateInfo {
+                max_buckets: 100, // Needs to be sufficiently large to touch each bucket.
+                max_reports: 100,
+            },
+        )
+        .await
+        .unwrap_err();
+    assert_matches!(err, DapAbort::BatchOverlap);
```
cjpatton, line 632: Hmm, this is odd. I would expect `POST /internal/process` to succeed (although no reports would be aggregated). Conversely, I would expect the second `POST /collect`above to fail. In fact, this is what the spec mandates.

daphne_worker_test/tests/e2e.rs
```diff
@@ -537,3 +540,94 @@ async fn e2e_leader_collect_abort_invalid_batch_interval() {
...
+        task_id: t.task_id.clone(),
+        batch_interval: Interval {
+            start: batch_interval.start,
+            duration: batch_interval.duration * 2,
+        },
+        agg_param: Vec::new(),
+    };
+    let _collect_uri = t
+        .leader_post_collect(&client, collect_req.get_encoded())
+        .await;
```
cjpatton, line 619: Why doesn't the test fail here?

cjpatton (changes requested):

daphne/src/lib.rs
nakatsuka-y: Completely forgot about this! Thanks for pointing it out.

daphne/src/lib.rs
nakatsuka-y: I've changed the naming of this and the other variables in this struct.

daphne/src/lib.rs
nakatsuka-y: See above.

daphne/src/lib.rs
nakatsuka-y: This is a number of seconds.

daphne/src/lib.rs
nakatsuka-y: Didn't know this :)
Updated Glob to Global

daphne_worker/src/config.rs
nakatsuka-y: Ditto!

daphne/src/lib.rs
nakatsuka-y: Changed `PeerError` to `Abort`.

daphne_worker/src/lib.rs
nakatsuka-y: I thought it would be nice to have a separate enum, but your point makes sense.
I've updated the enum.

daphne_worker_test/helper.env
nakatsuka-y: Thanks :)

daphne_worker_test/tests/e2e.rs
nakatsuka-y: Sounds good.

daphne_worker_test/tests/e2e.rs
nakatsuka-y: Grammar issues :(

daphne_worker_test/tests/e2e.rs
nakatsuka-y: I've updated the code so that the collect req gets checked when it gets posted.

daphne_worker_test/tests/e2e.rs
nakatsuka-y: It should now.

daphne/src/lib.rs
```diff
@@ -225,6 +236,25 @@ pub struct ProblemDetails {
     pub detail: Option<String>,
 }
 
+/// Global DAP parameters common across tasks.
+//
+// TODO spec: `max_window_size`, `min_duration_start`, and `max_duration_end`
```
cjpatton, line 241: These need to be renamed to match struct.

daphne/src/lib.rs
```diff
@@ -225,6 +236,25 @@ pub struct ProblemDetails {
...
+//            should be reflected to the DAP spec.
+#[derive(Deserialize, Serialize)]
+pub struct DapGlobConfig {
+    /// Maximum window size permitted in CollectReq.
+    /// Prevents Collectors from requesting wide range or reports.
+    pub max_batch_duration: u64,
+
+    /// Lower bound of an acceptable batch window in CollectReq.
+    /// Batch intervals must start after `min_batch_interval_start`.
+    pub min_batch_interval_start: u64,
```
cjpatton, line 251: Is this a fixed point in time, or a duration between the current time and the min batch interval start? I think it should be the latter, otherwise we'd have to update this periodically.

daphne_worker_test/tests/test_runner.rs
```diff
@@ -354,21 +388,25 @@ impl TestRunner {
         &self,
         client: &reqwest::Client,
         agg_info: &InternalAggregateInfo,
-    ) -> DapLeaderProcessTelemetry {
+    ) -> Result<DapLeaderProcessTelemetry, DapAbort> {
```
cjpatton, line 391: Let's revert this change, since we no longer need and it's causing lots of clutter in this PR.

cjpatton (commented):
Alright this is heading in a good direction 👍 

Let's see if we can refactor so that check for overlapping batches can be moved into the `DapLeader` and `DapAggregator`. Here's what I suggest we do:

* Add a method to the `DapAggregator` trait for checking whether a given batch interval would overlap with a previous request.
* Modify `DapLeader::http_post_collect()` to call the new method
* Modify `DapHelper::http_post_aggregate_share()` to call the new method.

daphne/src/lib.rs
nakatsuka-y: Thanks for pointing this out, it seemed that there were other locations that used the legacy variable name, so I've updated those in addition to this.

daphne/src/lib.rs
nakatsuka-y: `min_batch_interval_start` and `max_batch_interval_end` are duration between the current time and the min/max batch interval start/end.
Do you have a better name for this variable?

daphne_worker_test/tests/test_runner.rs
nakatsuka-y: Sounds good, I've done this.

daphne/src/lib.rs
```diff
@@ -225,6 +236,25 @@ pub struct ProblemDetails {
     pub detail: Option<String>,
 }
 
+/// Global DAP parameters common across tasks.
+//
+// TODO spec: `max_batch_duration`, `min_batch_interval_start`,
+// and `max_batch_interval_end` should be reflected to the DAP spec.
```
cjpatton, line 242: Hmm, arguably this is not needed. Would you mind bringing this up in an issue? https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/new I'm wondering if anyone has thought about this yet.

daphne/src/lib.rs
```diff
@@ -225,6 +236,25 @@ pub struct ProblemDetails {
...
+#[derive(Deserialize, Serialize)]
+pub struct DapGlobConfig {
+    /// Maximum window size permitted in CollectReq.
+    /// Prevents Collectors from requesting wide range or reports.
+    pub max_batch_duration: u64,
+
+    /// Lower bound of an acceptable batch window in CollectReq.
```
cjpatton, line 249: Let's avoid overloading the term "batch window", which has a different definition in the spec.

```suggestion
    /// Lower bound of an acceptable batch interval for collect requests.
```

daphne/src/lib.rs
cjpatton: These names are fine.

daphne/src/roles_test.rs
```diff
@@ -559,16 +690,157 @@ async fn poll_collect_job_test_results() {
...
+        payload: collector_collect_req.get_encoded(),
+        url: task_config.helper_url.join("/collect").unwrap(),
+        sender_auth: Some(BearerToken::from(COLLECTOR_BEARER_TOKEN.to_string())),
+    };
+
+    // Leader: Handle the CollectReq received from Collector.
+    let err = leader.http_post_collect(&req).await.unwrap_err();
+
+    // Fails because the requested batch interval is too large.
+    assert_matches!(err, DapAbort::InvalidBatchInterval);
```
cjpatton, line 724: The "invalidBatchInterval" has a very specific semantics in DAP-01, this is technically non-compliant. We should consider putting this in the spec, but in the mean time I think we need to handle this explicitly. Can we return `DapAbort::BadRequest` with a string that signifies why we're rejecting the report?

daphne_worker/src/dap.rs
```diff
@@ -297,12 +356,27 @@ impl<D> DapLeader<BearerToken> for DaphneWorkerConfig<D> {
             .get_task_config_for(&collect_req.task_id)
             .ok_or_else(|| DapError::fatal(INT_ERR_UNRECOGNIZED_TASK))?;
```
cjpatton, line 358: Sampling the clock twice can lead to strange errors. Sample it once, then use it twice
```rust
let now = now();
if some_condition_that_depends_on_the_current_time(now) {
 ....
}
```

daphne_worker/src/dap.rs
```diff
@@ -297,12 +356,27 @@ impl<D> DapLeader<BearerToken> for DaphneWorkerConfig<D> {
...
 
+        if collect_req.batch_interval.duration / task_config.min_batch_duration
+            > self.global_config.max_batch_duration
+            || now().abs_diff(collect_req.batch_interval.start)
+                > self.global_config.min_batch_interval_start
+            || now().abs_diff(collect_req.batch_interval.end())
+                > self.global_config.max_batch_interval_end
+        {
+            return Err(DapError::Abort(DapAbort::InvalidBatchInterval));
```
cjpatton, line 366: Just as for `MockAggregator`, this should return something like `DapAbort::BadRequest("batch interval ")`. You might also consider leaving a note here that we might consider addressing in the spec.

daphne_worker/src/durable/leader_col_job_queue.rs
```diff
@@ -24,10 +24,10 @@ pub(crate) const DURABLE_LEADER_COL_JOB_QUEUE_FINISH: &str =
...
     "/internal/do/leader_col_job_queue/get_result";
 
-#[derive(Debug, Deserialize, Serialize)]
-pub(crate) struct CollectionJobResult {
-    pub(crate) collect_id: Id,
-    pub(crate) collect_resp: CollectResp,
+#[derive(Deserialize, Serialize)]
+pub(crate) enum CollectionJobQueueResult {
+    Ok(Id),
+    ErrBatchOverlap,
```
cjpatton, line 30: This is no longer returned. Do we still need CollectionJobQueueResult? It seem slike this can be revereted

cjpatton (commented):
Nice! Just a bit of clean up.

daphne/src/lib.rs
nakatsuka-y: Since this is a Daphne-specific implementation detail, I've swapped the TODO with a note regarding whether to consider putting this into the spec or not.

daphne/src/lib.rs
nakatsuka-y: Good point. I've also updated the comment so that it describes what these variables are in more detail.

daphne/src/roles_test.rs
nakatsuka-y: I've used BadRequest and used the string to explicitly report about the reason.

daphne_worker/src/dap.rs
nakatsuka-y: I totally missed this. Thanks for pointing this out.

daphne_worker/src/dap.rs
nakatsuka-y: Changed!

daphne_worker/src/durable/leader_col_job_queue.rs
nakatsuka-y: Great point, I've removed this.

cjpatton (commented):
LGTM, go ahead and squash

cjpatton (approved):
🦆

PR #66, opened 2022-08-02
daphne_worker: Dequeue pending collect jobs properly

cjpatton:
After a collect job is completed, we fail to properly remove the collect
request from the job queue. This causes the /internal/process endpoint
to encounter an internal error.

POST /internal/process 500 Internal Server Error (146.19ms)
 ... [log line redacted] ...
DAP_ENV: Hostname override applied
internal error: LeaderCollectionJobQueue: tried to overwrite collect response
internal error: internalError: fatal error: worker: internalError
internalError

This change modifies the LeaderCollectJobQueue DO so that it removes the
job from the queue.

Note: This behavior was triggered by an interop test with Janus' Helper.

### DIFF
--- a/daphne_worker/src/durable/leader_col_job_queue.rs
+++ b/daphne_worker/src/durable/leader_col_job_queue.rs
@@ -157,17 +157,14 @@ impl DurableObject for LeaderCollectionJobQueue {
                     ));
                 }
 
+                let mut storage = self.state.storage();
                 let pending_key = format!("pending/{}", collect_id_hex);
-                let pending: Option<OrderedCollectReq> =
-                    state_get(&self.state, &pending_key).await?;
-                if pending.is_none() {
-                    return Err(int_err("LeaderCollectionJobQueue: missing collect request"));
-                }
-
+                let delete_pending_future = storage.delete(&pending_key);
                 self.state
                     .storage()
                     .put(&processed_key, res.collect_resp)
                     .await?;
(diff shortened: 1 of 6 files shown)


daphne_worker/src/durable/leader_col_job_queue.rs
```diff
@@ -157,17 +157,14 @@ impl DurableObject for LeaderCollectionJobQueue {
...
+                let mut storage = self.state.storage();
                 let pending_key = format!("pending/{}", collect_id_hex);
-                let pending: Option<OrderedCollectReq> =
-                    state_get(&self.state, &pending_key).await?;
-                if pending.is_none() {
-                    return Err(int_err("LeaderCollectionJobQueue: missing collect request"));
-                }
-
+                let delete_pending_future = storage.delete(&pending_key);
                 self.state
```
armfazh, line 163: are you able to reuse the storage variable for the put operation?
also 
`let delete_pending_future = storage.delete(&pending_key).await?;`

armfazh (approved):

daphne_worker/src/durable/leader_col_job_queue.rs
cjpatton: * Unfortunately no, rustc complains.
* I don't want to await here because I want to be able to run the request below in parallel. That's why the await comes later.

PR #70, opened 2022-08-04
daphne_worker: Improve garbage collection

cjpatton:
Adds a new durable object, `GarbageCollector`, that keeps track of every
DO instance that is created. When called, it invokes .delete_all() on
each DO instance scheduled for deletion. A DO instance schedules itself
for deletion by issuing a request to the garbage collector. This occurs
only once when the instance is first touched.

Replaces the `/internal/test/reset` endpoint with a new endpoint,
`/internal/delete_all`, that invokes the garbage collector. This allows
us to perform this task without needing to specify a range of buckets.
In particular, a batch interval and task ID are no longer needed.

This endpoint is only suitable for non-production deployments.
Eventually, we will schedule a cron job that periodically prunes the
oldest DO instances. In anticipation of this need, DO instances deleted
in the order in which they were touched.

### DIFF
--- a/daphne_worker/src/durable/garbage_collector.rs
+++ b/daphne_worker/src/durable/garbage_collector.rs
@@ -1,0 +1,107 @@
+// Copyright (c) 2022 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+use crate::{durable, int_err, now};
+use rand::prelude::*;
+use worker::*;
+
+pub(crate) const DURABLE_GARBAGE_COLLECTOR_PUT: &str = "/internal/do/garbage_collector/put";
+
+/// Durable Object (DO) for keeping track of all persistent DO storage.
+#[durable_object]
+pub struct GarbageCollector {
+    #[allow(dead_code)]
+    state: State,
+    env: Env,
+}
+
+#[durable_object]
+impl DurableObject for GarbageCollector {
+    fn new(state: State, env: Env) -> Self {
+        Self { state, env }
+    }
+
+    async fn fetch(&mut self, mut req: Request) -> Result<Response> {
+        let mut rng = thread_rng();
+        match (req.path().as_ref(), req.method()) {
+            // Schedule a durable object (DO) instance for deletion.
+            (DURABLE_GARBAGE_COLLECTOR_PUT, Method::Post) => {
+                let durable_ref: durable::DurableReference = req.json().await?;
+                match durable_ref.binding.as_ref() {
+                    durable::BINDING_DAP_REPORT_STORE
+                    | durable::BINDING_DAP_AGGREGATE_STORE
+                    | durable::BINDING_DAP_LEADER_AGG_JOB_QUEUE
+                    | durable::BINDING_DAP_LEADER_COL_JOB_QUEUE
+                    | durable::BINDING_DAP_HELPER_STATE_STORE => (),
+                    s => {
+                        return Err(int_err(format!(
+                            "GarbageCollector: unrecognized binding: {}",
+                            s
+                        )))
+                    }
+                };
+
+                // Reference handle encodes the current time and a random nonce in order to deal
+                // with collisions. This time goes first in order to ensure we can delete the
+                // oldest objects first.
+                let mut rand = [0; 16];
+                rng.fill(&mut rand);
+                let durable_handle = format!(
+                    "object/{}/time/{:020}/rand/{}",
+                    durable_ref.binding,
+                    now(),
+                    hex::encode(rand)
+                );
+                self.state
+                    .storage()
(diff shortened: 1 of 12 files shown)


daphne_worker/src/durable/garbage_collector.rs
```diff
@@ -0,0 +1,107 @@
...
+                        return Err(int_err(format!(
+                            "GarbageCollector: unrecognized binding: {}",
+                            s
+                        )))
+                    }
+                };
+
+                // Reference handle encodes the current time and a random nonce in order to deal
+                // with collisions. This time goes first in order to ensure we can delete the
+                // oldest objects first.
```
cjpatton, line 46: Reviewer: Please sanity check that this encoding actually results in things getting handled in order by (binding, timestamp).

daphne_worker/src/durable/leader_col_job_queue.rs
```diff
@@ -47,21 +47,24 @@ pub struct LeaderCollectionJobQueue {
...
     }
 
     async fn fetch(&mut self, mut req: Request) -> Result<Response> {
-        match (req.path().as_ref(), req.method()) {
-            (DURABLE_DELETE_ALL, Method::Post) => {
-                self.state.storage().delete_all().await?;
-                Response::empty()
-            }
+        let id_hex = self.state.id().to_string();
+        ensure_garbage_collected!(req, self, id_hex, BINDING_DAP_LEADER_COL_JOB_QUEUE);
```
armfazh, line 65: this macro expands two things, the end point for delete_all, and the logic for first touch this DO.
For the second task, isn't possible to register the durable object at creation time, rather than at the first touch. Imagine a Setup step, when everything is created and registered.

daphne_worker/src/durable/garbage_collector.rs
```diff
@@ -0,0 +1,107 @@
...
+            //
+            //   TODO Add a method for deleting all instances scheduled before a given time. This
+            //   will allow us to prune storage we're not likely to need anymore, e.g., for
+            //   replay protection. However, for replay protection in particular, it'll be
+            //   important to make sure the Leader rejects reports with old timestamps.
+            (durable::DURABLE_DELETE_ALL, Method::Post) => {
+                let opt = ListOptions::new().prefix("object/");
+                let iter = self.state.storage().list_with_options(opt).await?.entries();
+                let mut item = iter.next()?;
+                while !item.done() {
```
armfazh, line 82: when retrieving the entries, are they sorted?

cjpatton: Yes, according to Workers docs (search for "list"): https://developers.cloudflare.com/workers/runtime-apis/durable-objects/

daphne_worker/src/durable/leader_col_job_queue.rs
cjpatton: Yeah, it might be possible to stick this in the constructor `new()`, though I'm not sure if this would make a difference performance-wise. FWIW, I'm following a common pattern: https://github.com/cloudflare/workers-rs/blob/main/worker-sandbox/src/counter.rs#L23

daphne_worker/src/durable/garbage_collector.rs
```diff
@@ -0,0 +1,107 @@
...
+                    }
+                };
+
+                // Reference handle encodes the current time and a random nonce in order to deal
+                // with collisions. This time goes first in order to ensure we can delete the
+                // oldest objects first.
+                let mut rand = [0; 16];
+                rng.fill(&mut rand);
+                let durable_handle = format!(
+                    "object/{}/time/{:020}/rand/{}",
```
cjpatton, line 50: Hmm, maybe the binding should not be in front of the timestamp?

cjpatton: thinking about it more, I don't think there's a need.

daphne_worker/src/durable/leader_col_job_queue.rs
armfazh: it is not only related to performance, but to improve the clarity of code.

daphne_worker/src/durable/garbage_collector.rs
armfazh: this depends on whether the function returns elements sorted, and whether the map returned is sorted too. An extra countermeasure is to sort-again or check if it's sorted, but it may be too much overhead.

armfazh (approved):

daphne_worker/src/durable/leader_col_job_queue.rs
cjpatton: TBH I'm not sure when the constructor is called. Ultimately it's up to the Workers runtime. I think following the example would be a good idea.

Plus, the constructor doesn't return an error, which we would need it to do.

PR #76, opened 2022-08-07
Allow maximum batch duration to be accepted by Leader

nakatsuka-y:
This PR implements a test case where the maximum batch duration set by the global config is accepted by the Leader.

The global config numbers have been set to the following:
* max_batch_duration: 360,000 sec (100 batches of 3,600 sec)
* min_batch_interval_start: 279,200 sec (72 hours)
* max_batch_interval_end: 279,200 sec (72 hours)

### DIFF
--- a/daphne/src/roles.rs
+++ b/daphne/src/roles.rs
@@ -15,9 +15,9 @@ use crate::{
         AggregateShareReq, AggregateShareResp, CollectReq, CollectResp, Id, Interval, Nonce,
         Report, ReportShare, TransitionFailure,
     },
-    DapAbort, DapAggregateShare, DapCollectJob, DapError, DapHelperState, DapHelperTransition,
-    DapLeaderProcessTelemetry, DapLeaderTransition, DapOutputShare, DapRequest, DapResponse,
-    DapTaskConfig,
+    DapAbort, DapAggregateShare, DapCollectJob, DapError, DapGlobalConfig, DapHelperState,
+    DapHelperTransition, DapLeaderProcessTelemetry, DapLeaderTransition, DapOutputShare,
+    DapRequest, DapResponse, DapTaskConfig,
 };
 use async_trait::async_trait;
 use prio::codec::{Decode, Encode};
@@ -43,6 +43,28 @@ macro_rules! check_batch_param {
     }};
 }
 
+macro_rules! check_batch_interval_boundary {
+    (
+        $global_config:expr,
+        $now:expr,
+        $batch_interval:expr
+    ) => {{
+        if $batch_interval.duration > $global_config.max_batch_duration {
+            return Err(DapAbort::BadRequest("batch interval too large".to_string()));
+        }
+        if $now.abs_diff($batch_interval.start) > $global_config.min_batch_interval_start {
+            return Err(DapAbort::BadRequest(
+                "batch interval too far into past".to_string(),
+            ));
+        }
+        if $now.abs_diff($batch_interval.end()) > $global_config.max_batch_interval_end {
+            return Err(DapAbort::BadRequest(
+                "batch interval too far into future".to_string(),
+            ));
+        }
+    }};
+}
+
 /// A party in the DAP protocol who is authorized to send requests to another party.
 #[async_trait(?Send)]
 pub trait DapAuthorizedSender<S> {
@@ -61,8 +83,14 @@ pub trait DapAggregator<S>: HpkeDecrypter + Sized {
     /// Decide whether the given DAP request is authorized.
     async fn authorized(&self, req: &DapRequest<S>) -> Result<bool, DapError>;
 
+    /// Look up the DAP global configuration.
+    fn get_global_config(&self) -> &DapGlobalConfig;
+
     /// Look up the DAP task configuration for the given task ID.
     fn get_task_config_for(&self, task_id: &Id) -> Option<&DapTaskConfig>;
+
+    /// Get the current time (number of seconds since the beginning of UNIX time).
+    fn get_current_time(&self) -> u64;
 
     /// Check whether the given collect request interval does not overlap
     /// with requests received in the past.
@@ -230,6 +258,11 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
             collect_req.batch_interval,
             collect_req.agg_param
         );
+
+        // Check that the batch interval is not too wide or too far into the past/future.
+        let global_config = self.get_global_config();
+        let now = self.get_current_time();
+        check_batch_interval_boundary!(global_config, now, collect_req.batch_interval);
 
         // Check that the batch interval does not overlap with past requests.
         if self
@@ -580,6 +613,11 @@ pub trait DapHelper<S>: DapAggregator<S> {
             agg_share_req.batch_interval,
             agg_share_req.agg_param
         );
+
+        // Check that the batch interval is not too wide or too far into the past/future.
+        let global_config = self.get_global_config();
+        let now = self.get_current_time();
+        check_batch_interval_boundary!(global_config, now, agg_share_req.batch_interval);
 
         // Check that the batch interval does not overlap with past requests.
         if self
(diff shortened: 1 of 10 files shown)


daphne/src/testing.rs
```diff
@@ -554,9 +554,7 @@ impl DapLeader<BearerToken> for MockAggregator {
             .unwrap()
             .as_secs();
 
-        if collect_req.batch_interval.duration / task_config.min_batch_duration
-            > self.global_config.max_batch_duration
-        {
+        if collect_req.batch_interval.duration > self.global_config.max_batch_duration {
             return Err(DapError::Abort(DapAbort::BadRequest(
                 "batch interval too large".to_string(),
             )));
```
cjpatton, line 560: Why do we need these checks in MockAggregator, now that they're implemented in the generic DAP logic?

daphne_worker_test/tests/e2e.rs
```diff
@@ -490,6 +490,27 @@ async fn e2e_leader_collect_abort_unknown_request() {
     assert_eq!(resp.status(), 400);
 }
 
+#[tokio::test]
+#[cfg_attr(not(feature = "test_e2e"), ignore)]
+async fn e2e_leader_collect_accept_max_batch_interval() {
```
cjpatton, line 495:
```suggestion
async fn e2e_leader_collect_accept_max_batch_duration() {
```

daphne_worker_test/tests/e2e.rs
```diff
@@ -490,6 +490,27 @@ async fn e2e_leader_collect_abort_unknown_request() {
...
+    let t = TestRunner::default().await;
+    let client = t.http_client();
+    let batch_interval = t.batch_interval();
+
+    // Maximum allowed batch duration.
+    let collect_req = CollectReq {
+        task_id: t.task_id.clone(),
+        batch_interval: Interval {
+            start: batch_interval.start - t.max_batch_duration / 2,
```
cjpatton, line 504: Why subtract `t.max_batch_duration / 2`?

daphne/src/testing.rs
nakatsuka-y: The interval check has to be implemented in both MockAggregator and Daphne Workers, as it is called during within the `init_collect_job` function.

daphne_worker_test/tests/e2e.rs
nakatsuka-y: Thanks for catching this!

daphne_worker_test/tests/e2e.rs
nakatsuka-y: This is so that `batch_interval.start` and `batch_interval.end()` both fit into the `min_batch_interval_start` and `max_batch_interval_end` limits.

daphne/src/testing.rs
cjpatton: My point is: Why isn't this logic in `DapLeader::http_post_collect()` Instead?

daphne_worker_test/tests/e2e.rs
cjpatton: I see. It would be cleaner to construct `batch_interval` yourself, rather than use `t.batch_inteval()`, which is used as the default.

daphne/src/testing.rs
nakatsuka-y: I've moved this to http_post_collect.

daphne_worker_test/tests/e2e.rs
nakatsuka-y: Great idea, I've implemented this.

daphne/src/roles.rs
```diff
@@ -61,9 +61,15 @@ pub trait DapAggregator<S>: HpkeDecrypter + Sized {
     /// Decide whether the given DAP request is authorized.
     async fn authorized(&self, req: &DapRequest<S>) -> Result<bool, DapError>;
 
+    /// Look up the DAP global configuration.
+    fn get_global_config(&self) -> &DapGlobalConfig;
+
     /// Look up the DAP task configuration for the given task ID.
     fn get_task_config_for(&self, task_id: &Id) -> Option<&DapTaskConfig>;
 
+    /// Get the current time.
```
cjpatton, line 70: We should document the format of the time (number of seconds since UNIX epoch).

daphne/src/roles.rs
```diff
@@ -231,6 +237,24 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
             collect_req.agg_param
         );
```
cjpatton, line 239: `http_post_aggregate_share` should implement this check as well, right?

daphne/src/roles.rs
nakatsuka-y: Sounds good. I've updated the comment.

daphne/src/roles.rs
nakatsuka-y: Oops, you're right.

daphne/src/roles.rs
```diff
@@ -231,6 +237,24 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
...
 
+        // Check that the batch interval is not too wide or too far into the past/future.
```
cjpatton, line 240: Looks good. One more thing: Since the check is identical in the Leader and Helper, we should try to deduplicate it. There are two ways to do this:
* Create a macro and call it from both methods
* Add a derived method to `DapAggregator` that performs this check and call it from both methods.

daphne/src/roles_test.rs
```diff
@@ -764,6 +772,38 @@ async fn http_post_collect_fail_invalid_batch_interval() {
     assert_matches!(err, DapAbort::BadRequest(s) => assert_eq!(s, "batch interval too far into future".to_string()));
 }
 
+#[tokio::test]
+async fn http_post_collect_succeed_max_batch_interval() {
+    let leader = MockAggregator::new();
+    let task_id = leader.nominal_task_id();
+    let task_config = leader.get_task_config_for(task_id).unwrap();
+    let now = SystemTime::now()
```
cjpatton, line 780: How about using `leader.get_current_time()`?

daphne_worker_test/tests/test_runner.rs
```diff
@@ -264,6 +264,14 @@ impl TestRunner {
         }
     }
 
+    pub fn max_batch_interval(&self) -> Interval {
```
cjpatton, line 267: It's not clear what this method is supposed to do based on its signature alone.

* All methods in public APIs should have documentation. This is true no matter what language or codebase.
* This particular method is only used once and is not likely needed elsewhere. I think it's better to just construct the batch interval you need, where you need it.

nakatsuka-y: Maybe I'm not understanding it correctly, but this was to resolve this comment (https://github.com/cloudflare/daphne/pull/76#discussion_r940433535).
What do you mean when you say "don't use `t.batch_interval()`" and "construct batch_interval() yourself"?
We need `t.batch_interval()` in any case...

cjpatton: I mean construct the batch interval explicitly:
```rust
       batch_interval: Interval {
            start: t.now - t.max_batch_duration / 2,
            duration: t.max_batch_duration,
        },
```
This is much more readable.

daphne/src/roles.rs
nakatsuka-y: Good point. I've created a macro, since I didn't want to put too much methods in DapAggregator.

daphne/src/roles_test.rs
nakatsuka-y: Great idea!! I've replaced all SystemTime::now() with get_current_time.

daphne_worker_test/tests/test_runner.rs
nakatsuka-y: I've removed the function and directly create batch_interval.

cjpatton (commented):
LGTM. Go ahead and squash.

cjpatton (approved):

PR #77, opened 2022-08-08
Supporting different versions of DAP specs in Daphne

### LINKED ISSUE #71: Versioning the endpoints
We're anticipating the need to support multiple DAP versions simultaneously. In particular, https://github.com/cloudflare/daphne/issues/65 will introduce changes that are incompatible with DAP-01. Thus, we'll need to gate these changes for DAP-02.

As discussed on the mailing list [1], we could version our leader and helper URL endpoints so that the client (or collector) opts-in to a particular version by picking a particular endpoint. For example, to upload a report for DAP-02, the client POST to `/02/upload` instead of `/upload`.

Some additional requirements:
* We would need to plumb the draft version through `DapRequest`. We could add a field whose value is
```rust
enum DapVersion {
    Draft01,
    Draft02,
}
```
* We probably should lock each task to a particular version. (Not sure if this requires a spec change.) To do so we could add the `DapVersion` to the `DapTaskConfig` structure.

@nakatsuka-y what do you think of this as a first step towards solving #65?
  

[1] https://mailarchive.ietf.org/arch/msg/ppm/svv73HxLvTMdOqk2fLyuWmZRU6Q/

nakatsuka-y:
This closes #71.

As discussed in the issue, there would be cases where Daphne has multiple versions of DAP running at the same time.
To be able to distinguish between multiple versions, this PR allows clients to specify which version they are using through an URL endpoint (e.g. `/v01/collect`).

This functionality has been implemented in both MockAggregator and Daphne Workers and unit tests that aborts with `DapAbort::InvalidDapVersion` when clients uses the wrong DAP version is also added.


daphne/src/lib.rs
```diff
@@ -125,6 +125,11 @@ pub enum DapAbort {
     #[error("invalidBatchInterval")]
     InvalidBatchInterval,
 
+    /// Invalid DAP version. Sent in response to requests that use a DAP version URL endpoint
+    /// that does not match to the version set in TaskConfig or simply does not exist.
+    #[error("invalidDapVersion")]
+    InvalidDapVersion,
```
cjpatton, line 131:
```suggestion
    InvalidProtocolVersion,
```

daphne/src/lib.rs
```diff
@@ -236,6 +242,24 @@ pub struct ProblemDetails {
     pub detail: Option<String>,
 }
 
+/// DAP version used for a task.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub enum DapVersion {
+    Unknown,
+    Draft01,
+    Draft02,
+}
```
cjpatton, line 251: * I think the JSON serialization of this enum should match the endpoints (https://serde.rs/variant-attrs.html#rename)
* No need to add variant for  02 since we don't support this version yet.
* Serde has a feature that allows one of the variants to be the "default" if none of the other tags are matched.
* Prefer snake case to camel case.
```suggestion
pub enum DapVersion {
    #[serde(rename = "v01")]
    Draft01,
    #[serde(other)]
    Unknown,
}
```

daphne/src/lib.rs
```diff
@@ -236,6 +242,24 @@ pub struct ProblemDetails {
...
+
+impl DapVersion {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            DapVersion::Unknown => "unknown",
+            DapVersion::Draft01 => "v01",
+            DapVersion::Draft02 => "v02",
+        }
+    }
```
cjpatton, line 260: This method is used to convert a string into a `DapVersion`. The proper way to do this in Rust is to implement `From<&str> for DapVersion`.

daphne/src/roles.rs
```diff
@@ -239,6 +245,11 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
             return Err(DapAbort::BatchOverlap);
         }
 
+        // Check whether the DAP version in the request matches the one in task config.
```
cjpatton, line 248: What happens here if the `CollectReq` has a different wire format between two versions?

cjpatton (changes requested):

daphne_worker/src/lib.rs
```diff
@@ -102,9 +102,16 @@ impl DaphneWorkerRouter {
...
+        let router = Router::new().get_async("/:version/hpke_config", |req, ctx| async move {
+            let version = ctx.param("version").unwrap().to_string();
             let config = DaphneWorkerConfig::from_worker_context(ctx)?;
-            let req = config.worker_request_to_dap(req).await?;
+            let req = config.worker_request_to_dap(version, req).await?;
+
+            // Check that the DapVersion is valid.
+            if req.version == DapVersion::Unknown {
+                return abort(DapAbort::InvalidDapVersion);
+            }
```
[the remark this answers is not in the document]
nakatsuka-y: You are right. Moved this to request handler methods.

daphne_worker/src/config.rs
```diff
@@ -292,6 +292,7 @@ impl<D> DaphneWorkerConfig<D> {
 
     pub(crate) async fn worker_request_to_dap(
         &self,
+        version: String,
```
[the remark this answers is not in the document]
nakatsuka-y: I had to play around with `matchit::Router`, but I've managed to do this. Please take a look.

daphne/src/lib.rs
nakatsuka-y: Fixed!

daphne/src/lib.rs
nakatsuka-y: I didn't know about this feature in serde. This makes the code cleaner.

daphne/src/lib.rs
nakatsuka-y: Thanks for the suggestion. I've implemented this.

daphne/src/roles.rs
nakatsuka-y: Should we leave a note that says: "Decode `CollectReq` depending on the protocol version"?

daphne/src/lib.rs
```diff
@@ -236,6 +242,35 @@ pub struct ProblemDetails {
...
+        match self {
+            DapVersion::Draft01 => "v01",
+            DapVersion::Unknown => "unknown",
+        }
+    }
+}
+
+impl From<&str> for DapVersion {
+    fn from(version: &str) -> Self {
+        if version == DapVersion::Draft01.as_str() {
```
cjpatton, line 266: A `match` statement would be cleaner here.
```rust
match version {
   "v01" => DapVersion::Draft01,
    _ => DapVersion::Unknown,
}
```
That way you can do away with `as_str()` above altogether.

daphne/src/lib.rs
```diff
@@ -236,6 +242,35 @@ pub struct ProblemDetails {
...
+
+    #[serde(other)]
+    Unknown,
+}
+
+impl DapVersion {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            DapVersion::Draft01 => "v01",
+            DapVersion::Unknown => "unknown",
```
cjpatton, line 259: I don't think this variant needs to be representable as a string. (See comment below.)

daphne/src/roles.rs
```diff
@@ -263,6 +269,11 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
...
+        if task_config.version != req.version || req.version == DapVersion::Unknown {
+            return Err(DapAbort::InvalidProtocolVersion);
+        }
```
cjpatton, line 275: In all of these methods, the version check needs to be the very first thing that happens. Why?

cjpatton (changes requested):
Coming along! One high level comment: The collect URI should also contain the version prefix.

daphne/src/lib.rs
nakatsuka-y: Agreed, this is cleaner.

daphne/src/lib.rs
nakatsuka-y: I agree, I've removed the as_str() method.

daphne/src/roles.rs
nakatsuka-y: This is because (1) the version must not be "Unknown" to be able to decode the data correctly and (2) depending on the version, the struct the data is decoding to may be different, right?

cjpatton: Yes.

daphne/src/roles.rs
```diff
@@ -198,9 +213,19 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
...
-        if self.get_task_config_for(&report.task_id).is_none() {
-            return Err(DapAbort::UnrecognizedTask);
+        let task_config = self
+            .get_task_config_for(&report.task_id)
+            .ok_or(DapAbort::UnrecognizedTask)?;
+
+        // Check whether the DAP version in the request matches the one in task config.
+        if task_config.version != req.version {
+            return Err(DapAbort::InvalidProtocolVersion);
         }
```
cjpatton, line 229: It looks like you forgot to remove this code?

daphne/src/roles.rs
```diff
@@ -227,6 +252,11 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
             return Err(DapAbort::UnauthorizedRequest);
         }
 
+        // Check whether the DAP version in the request is not Unknown.
```
cjpatton, line 255: This check should be the very first thing (even before authorization).

daphne/src/roles.rs
```diff
@@ -198,9 +213,19 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
     /// Handle HTTP POST to `/upload`. The input is the encoded report sent in the body of the HTTP
     /// request.
     async fn http_post_upload(&self, req: &DapRequest<S>) -> Result<(), DapAbort> {
+        // Check whether the DAP version in the request is not Unknown.
```
cjpatton, line 216: Here and below.
```suggestion
        // Check whether the DAP version indicated by the sender is supported.
```

daphne/src/roles.rs
```diff
@@ -237,6 +267,11 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
             collect_req.agg_param
         );
 
+        // Check whether the DAP version in the request matches the one in task config.
```
cjpatton, line 270:
```suggestion
        // Check whether the DAP version in the request matches the task config.
```

daphne/src/roles.rs
```diff
@@ -313,7 +348,7 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
             self,
             task_id,
             task_config,
-            "/aggregate",
+            format!("/{}/{}", task_config.version, "aggregate").as_str(),
```
cjpatton, line 351: This breaks interop, since not every helper is going to implement versioning the way we do. (Here and below.) Instead of changing how the endpoints are constructed here, the helper URL in the task configuration will need to be changed.

daphne/src/roles.rs
```diff
@@ -595,6 +645,11 @@ pub trait DapHelper<S>: DapAggregator<S> {
             return Err(DapAbort::UnauthorizedRequest);
         }
 
+        // Check whether the DAP version in the request is not Unknown.
```
cjpatton, line 648: This should happen first (before authorization).

cjpatton (changes requested):

daphne/src/roles.rs
nakatsuka-y: Do we want to remove this check? I think it is important that the version in the request and the one that is agreed in the task config are the same.

cjpatton: Oops, sorry I forgot to remove this. I realized after making this comment that you're checking first whether the version is known and then whether the version is supported, which is exactly right.

Now, it would be fair to delegate the latter to the function that needs it. For example, if decoding needs the version number, then we could just pass the version number there. But I think what you have is fine too.

daphne/src/roles.rs
nakatsuka-y: Moved this before authorization.

daphne/src/roles.rs
nakatsuka-y: Thanks for pointing out the grammar issue :)

daphne/src/roles.rs
nakatsuka-y: Fixed!

daphne/src/roles.rs
nakatsuka-y: I've removed all the endpoint construction using versions.

daphne/src/roles.rs
nakatsuka-y: Ditto :)

daphne/src/roles.rs
```diff
@@ -322,7 +357,7 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
             self,
             task_id,
             task_config,
-            "/aggregate",
+            "aggregate",
```
cjpatton, line 360: This appears to be an accidental regression (here and below)
```suggestion
            "/aggregate",
```

daphne/src/roles.rs
```diff
@@ -402,7 +437,7 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
             self,
             &collect_req.task_id,
             task_config,
-            "/aggregate_share",
+            "aggregate_share",
```
cjpatton, line 440:
```suggestion
            "/aggregate_share",
```

daphne_worker/src/config.rs
```diff
@@ -300,8 +301,16 @@ impl<D> DaphneWorkerConfig<D> {
             None => None,
         };
 
+        let url = req.url()?;
+        let path = url.path();
+        let mut router: Router<bool> = Router::new();
+        router.insert("/:version/*remaining", true).unwrap();
+        let url_match = router.at(path).unwrap();
+        let version = url_match.params.get("version").unwrap();
```
cjpatton, line 309: Which of these unwrap()'s do you believe is safe? If the version prefix is missing or malformed, we'll want to handle this gracefully.

cjpatton (changes requested):
It looks like you've changed how paths are computed in several places, e.g.: https://github.com/cloudflare/daphne/pull/77/files#diff-6edcf4a9c3adc1ac1e8061cd876a931a9dedf381bdf3ce49d70b34b95a697205L325

What's the reasoning for this? Is this needed for the current PR? If not, then it's always best to keep out unecessary changes in order to reduce reviewer effort.

daphne_worker/src/lib.rs
```diff
@@ -198,7 +203,7 @@ impl DaphneWorkerRouter {
         };
 
         let router = if self.enable_internal_test {
-            router.post_async("/internal/delete_all", |_req, ctx| async move {
+            router.post_async("/:version/internal/delete_all", |_req, ctx| async move {
```
cjpatton, line 206: Internal APIs don't need to be versioned (and they shouldn't be).

daphne/src/lib.rs
```diff
@@ -126,6 +126,11 @@ pub enum DapAbort {
     #[error("invalidBatchInterval")]
     InvalidBatchInterval,
 
+    /// Invalid DAP version. Sent in response to requests that use a DAP version URL endpoint
+    /// that does not match to the version set in TaskConfig or simply does not exist.
```
cjpatton, line 130: This mechanism isn't specific to versioned URLs:
```suggestion
    /// Invalid DAP version. Sent in response to requests for an unsupported (or unknown) DAP version.
```

daphne/src/roles.rs
nakatsuka-y: This is a side-effect of using `url::Url` `join()`.

daphne_worker/src/config.rs
nakatsuka-y: The third one should be handled so that if there is no version, then we return an error.

daphne_worker/src/lib.rs
nakatsuka-y: I've used a sort of a "hack-y" way of removing "version" from the URL.
I've added a note just in case someone is curious.

daphne_worker_test/tests/test_runner.rs
```diff
@@ -382,6 +394,9 @@ impl TestRunner {
         client: &reqwest::Client,
         agg_info: &InternalAggregateInfo,
     ) -> DapLeaderProcessTelemetry {
+        // NOTE(nakatsuka-y) The URL below is made sort of a hack-y way where we
+        // use url::Url deleting the path from the base URL when calling "join()"
+        // to our advantage to prevent us from having "version" in this URL.
```
cjpatton, line 399: This is fine, but let's raise an issue to fix this later. Please convert this NOTE to a TODO and add a reference to the issue. (Here and below.)

daphne_worker_test/tests/test_runner.rs
```diff
@@ -59,8 +63,9 @@ const LEADER_TASK_LIST: &str = r#"{
         "vdaf_verify_key": "1fd8d30dc0e0b7ac81f0050fcab0782d"
     },
     "410d5e0abd94a88b8435a192cc458cc1667da2989827584cbf8a591626d5a61f": {
-        "leader_url": "http://leader:8787",
-        "helper_url": "http://127.0.0.1:9788",
+        "version": "v01",
```
cjpatton, line 66: Given that we require the version in the URL, it's a bit awkward to have a separate version as well. Let's remove this field from the TaskConfig and the corresponding version checks. Right now they're fairly cumbersome and don't add a lot of value.

cjpatton (changes requested):
Just two more minor comments.

daphne/src/lib.rs
```diff
@@ -126,6 +126,11 @@ pub enum DapAbort {
...
+    #[error("InvalidProtocolVersion")]
```
cjpatton, line 131:
```suggestion
    #[error("invalidProtocolVersion")]
```

daphne_worker_test/tests/test_runner.rs
nakatsuka-y: Done. See #80

daphne/src/lib.rs
nakatsuka-y: Thanks for pointing this out! Completely missed this

cjpatton (approved):

PR #78, opened 2022-08-08
Add support for HPKE configurations based on P-256

### LINKED ISSUE #39: Add support for HPKE P-256
We currently support X25519. P-256 is likely going to be needed by many users.

chris-wood:
Closes #39

This is a temporary solution. In the future, when we have a more ergonomic HPKE library that's friendly to agility, we can generalize `HpkeConfig` and its users over the HPKE algorithms.

### DIFF
--- a/daphne/src/hpke.rs
+++ b/daphne/src/hpke.rs
@@ -10,7 +10,7 @@ use crate::{
     },
     DapError,
 };
-use hpke::{aead, kdf, kem, Deserializable, HpkeError, Kem, OpModeR};
+use hpke::{aead, kdf, kem, Deserializable, Kem, OpModeR};
 use prio::codec::{CodecError, Decode, Encode};
 use serde::{Deserialize, Serialize};
 use std::io::Cursor;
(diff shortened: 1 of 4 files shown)


daphne/src/hpke.rs
```diff
@@ -39,15 +40,41 @@ impl HpkeConfig {
...
+                    _,
+                >(
+                    &OpModeS::Base, &pk, info, &mut rand::thread_rng()
+                )?;
+
+                Ok((enc.to_bytes().to_vec(), sender.seal(plaintext, aad)?))
+            }
+            _ => {
+                // We should never get here.
+                panic!("Unsupported KEM ({:?})", self.kem_id);
```
cjpatton, line 75: Here and below
```suggestion
                // We should never get here.
                unreachable!("Unsupported KEM ({:?})", self.kem_id);
```

daphne/src/hpke.rs
```diff
@@ -120,30 +176,74 @@ impl HpkeReceiverConfig {
...
     ) -> Result<Vec<u8>, HpkeError> {
-        let sk = <kem::X25519HkdfSha256 as Kem>::PrivateKey::from_bytes(&self.secret_key)?;
-        let enc = <kem::X25519HkdfSha256 as Kem>::EncappedKey::from_bytes(enc)?;
-        let mut receiver = hpke::setup_receiver::<
-            aead::AesGcm128,
-            kdf::HkdfSha256,
-            kem::X25519HkdfSha256,
-        >(&OpModeR::Base, &sk, &enc, info)?;
-        receiver.open(ciphertext, aad)
+        self.check_suite().map_err(|_| HpkeError::OpenError)?; // Note(caw): any error will do here
```
cjpatton, line 179: We could change the error type from `HpkeError` to `DapError`, then the `?` should work without mapping the error first.

cjpatton (changes requested):

daphne/src/hpke.rs
chris-wood: Done.

daphne/src/hpke.rs
chris-wood: Done.

cjpatton (approved):

PR #84, opened 2022-08-09
dapf: Construct API endpoints correctly

cjpatton:
Based on #79 (merge that first).

While at it, freeze prio at version 0.8.2.

### DIFF
--- a/daphne/src/bin/dapf.rs
+++ b/daphne/src/bin/dapf.rs
@@ -125,7 +125,7 @@ fn main() -> Result<()> {
                 reqwest::header::HeaderValue::from_static(constants::MEDIA_TYPE_REPORT),
             );
             let resp = http_client
-                .post(Url::parse(leader_url)?.join("/upload")?)
+                .post(Url::parse(leader_url)?.join("upload")?)
                 .body(report.get_encoded())
                 .headers(headers)
                 .send()?;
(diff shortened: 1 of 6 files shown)


daphne/src/bin/dapf.rs
```diff
@@ -125,7 +125,7 @@ fn main() -> Result<()> {
                 reqwest::header::HeaderValue::from_static(constants::MEDIA_TYPE_REPORT),
             );
             let resp = http_client
-                .post(Url::parse(leader_url)?.join("/upload")?)
+                .post(Url::parse(leader_url)?.join("upload")?)
```
chris-wood, line 128: Does this insert the path separator `/`?

chris-wood (approved):
Approved pending one question.

daphne/src/bin/dapf.rs
cjpatton: @nakatsuka-y observed an interesting and surprising behavior of the `url` crate. If `leader_url` ends ends with `/thing/`, then joining with `/upload` will overwrite `/thing`. For example if the the url is `example.com/thing/`, then the new url would be `example.com/upload`. Removing the `/` prefix ensures we get what we want, which is `example.com/thing/upload`. Very strange.

PR #86, opened 2022-08-10
daphne_worker: Helper: Coalesce requests to ReportStore

cjpatton:
Based on #85 (merge that first).

The Helper makes requests to ReportStore when deciding to reject
reports early. This change coalesces requests to each ReportStore
instance, thereby removing a performance bottleneck.

While at it, this change also adds timing information to logs:
* How long each request from Leader to Helper takes (round trip)
* How long it takes an Aggregator to answer a request

### DIFF
--- a/daphne_worker/src/dap.rs
+++ b/daphne_worker/src/dap.rs
@@ -490,10 +490,13 @@ impl<D> DapLeader<BearerToken> for DaphneWorkerConfig<D> {
             .body(payload)
             .headers(headers);
 
+        let start = Date::now().as_millis();
         let reqwest_resp = reqwest_req
             .send()
             .await
             .map_err(|e| DapError::Fatal(e.to_string()))?;
+        let end = Date::now().as_millis();
+        console_log!("request to {} completed in {}ms", url, end - start);
         let status = reqwest_resp.status();
         if status == 200 {
             // Translate the reqwest response into a Worker response.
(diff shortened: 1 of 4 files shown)


daphne_worker/src/dap.rs
```diff
@@ -531,28 +534,31 @@ impl<D> DapHelper<BearerToken> for DaphneWorkerConfig<D> {
...
+            let nonce_hex = hex::encode(report_share.nonce.get_encoded());
+            if let Some(nonce_hex_set) = durable_requests.get_mut(&durable_name) {
+                nonce_hex_set.push(nonce_hex);
+            } else {
+                durable_requests.insert(durable_name, vec![nonce_hex]);
+            }
+        }
 
-            // TODO Don't block on DO requests (issue multiple requests simultaneously).
+        // TODO Run requests in parallel.
```
chris-wood, line 550: Does this mean run the POSTs to each `durable_name` in parallel? (Just want to clarify)

daphne_worker/src/durable/report_store.rs
```diff
@@ -215,8 +215,18 @@ impl DurableObject for ReportStore {
...
             (DURABLE_REPORT_STORE_PUT_PROCESSED, Method::Post) => {
-                let nonce_hex: String = req.json().await?;
-                Response::from_json(&self.checked_process(&nonce_hex).await?)
+                let nonce_hex_set: Vec<String> = req.json().await?;
+                let mut futures = Vec::new();
+                for nonce_hex in nonce_hex_set.iter() {
+                    futures.push((nonce_hex.clone(), self.checked_process(nonce_hex)));
+                }
+                let mut res = Vec::with_capacity(futures.len());
+                for (nonce_hex, future) in futures.into_iter() {
```
chris-wood, line 224: Is it possible to benchmark this loop? It's not clear to me if it's the communication cost to the DO that dominates, or the cost for processing each nonce that dominates.

chris-wood (approved):

chris-wood:
LGTM with one suggestion for additional benchmarking.

daphne_worker/src/dap.rs
cjpatton: No, this change just coalesces requests to each distinct DO isntance. The requests themselves are still serial. Note that this gets fixed higher in the stack.

daphne_worker/src/durable/report_store.rs
cjpatton: This PR adds the duration of each request to logs. I can confirm on a live deployment that coalescing significantly improves runtime.

chris-wood: Sorry, I'm asking about benchmarking what's happening within each request. With this change, each request processes a set of nonces. How long does processing each nonce take?

cjpatton: OHhh got it. Yeah we don't have that yet.

PR #87, opened 2022-08-10
daphne_worker: Helper: Parallelize requests to ReportStore

cjpatton:
Based on #86 (merge that first).

Modifies `DapHelper::mark_aggregated()` to issue requests to each
ReportStore instance in parallel.

### DIFF
--- a/daphne_worker/src/dap.rs
+++ b/daphne_worker/src/dap.rs
@@ -534,7 +534,6 @@ impl<D> DapHelper<BearerToken> for DaphneWorkerConfig<D> {
     ) -> std::result::Result<HashMap<Nonce, TransitionFailure>, DapError> {
         let task_config = self.try_get_task_config_for(task_id)?;
 
-        let namespace = self.durable_object(BINDING_DAP_REPORT_STORE)?;
         let mut durable_requests: HashMap<String, Vec<String>> = HashMap::new();
         for report_share in report_shares.iter() {
             let durable_name =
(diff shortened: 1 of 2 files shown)


daphne_worker/src/dap.rs
```diff
@@ -547,14 +546,19 @@ impl<D> DapHelper<BearerToken> for DaphneWorkerConfig<D> {
             }
         }
 
-        // TODO Run requests in parallel.
```
chris-wood, line 550: 👍

daphne_worker/src/dap.rs
```diff
@@ -547,14 +546,19 @@ impl<D> DapHelper<BearerToken> for DaphneWorkerConfig<D> {
...
+                BINDING_DAP_REPORT_STORE,
+                DURABLE_REPORT_STORE_PUT_PROCESSED,
+                durable_name,
+                nonce_hex_set,
+            ));
+        }
 
-            let res: Vec<(String, TransitionFailure)> = resp.json().await?;
+        let mut early_fails = HashMap::new();
+        for future in futures.into_iter() {
```
chris-wood, line 560: Does this serialize the order in which we process the futures? Say, for example, the first DO message takes forever to process, but the latter ones complete quickly. Will we block here waiting on a response from the first future? Is it possible instead to run this functionality in a callback that is invoked when the DO future returns an answer?

chris-wood (approved):
This is a good improvement, but can we do more?

daphne_worker/src/dap.rs
cjpatton: > Does this serialize the order in which we process the futures? Say, for example, the first DO message takes forever to process, but the latter ones complete quickly. Will we block here waiting on a response from the first future?

I don't think any request is blocked on any other request. We `await` each request in the order in which they were issued, but waiting for one doesn't block handing of the other.

This is just my intuition, however. There may be hidden performance bottlenecks I don't see. 

> Is it possible instead to run this functionality in a callback that is invoked when the DO future returns an answer?
This can be done with `try_join_all()` from the `futures` crate. This is implemented in the next PR.

PR #88, opened 2022-08-10
daphne: Improve DapHelper::http_post_aggregate() parallelization

cjpatton:
Based on #87 (merge that first).

Check if the state already exists in parallel.

### DIFF
--- a/daphne_worker/src/dap.rs
+++ b/daphne_worker/src/dap.rs
@@ -46,6 +46,7 @@ use daphne::{
     DapAggregateShare, DapCollectJob, DapError, DapGlobalConfig, DapHelperState, DapOutputShare,
     DapRequest, DapResponse, DapTaskConfig,
 };
+use futures::future::try_join_all;
 use prio::codec::{Decode, Encode};
 use std::collections::HashMap;
 use worker::*;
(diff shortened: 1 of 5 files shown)


daphne_worker/src/dap.rs
```diff
@@ -546,20 +547,20 @@ impl<D> DapHelper<BearerToken> for DaphneWorkerConfig<D> {
...
-            futures.push(self.durable_post(
+            requests.push(self.durable_post(
                 BINDING_DAP_REPORT_STORE,
                 DURABLE_REPORT_STORE_PUT_PROCESSED,
                 durable_name,
                 nonce_hex_set,
             ));
         }
 
+        let responses: Vec<Vec<(String, TransitionFailure)>> = try_join_all(requests).await?;
```
chris-wood, line 560: Is this what we want? If any future fails then all futures are cancelled.

cjpatton: Great question. Any failure here would be a failure of the Workers runtime itself and would be handled by DaphneWorker as a 500 internal error.

It might be possible to recover by, say, "unmarking" any reports as aggregated. But this would require another call to DO instances, which is failing under this scenario :(

Thus I think hard fail is appropriate here.

chris-wood (approved):

PR #89, opened 2022-08-10
daphne_worker: Helper Don't garbage collect HelperStateStore

cjpatton:
Removes garbage collection of instances of the HelperStateStore. If the
Leader abandons an aggregation job before the last request, this will
result in an instance of this DO being left in storage indefinitely.

### DIFF
--- a/daphne_worker/src/durable/aggregate_store.rs
+++ b/daphne_worker/src/durable/aggregate_store.rs
@@ -44,7 +44,6 @@ pub(crate) enum AggregateStoreResult {
 /// duration.
 #[durable_object]
 pub struct AggregateStore {
-    #[allow(dead_code)]
     state: State,
     env: Env,
     touched: bool,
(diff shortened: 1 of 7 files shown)


chris-wood:
> This improves performance significantly, but there is a risk of the Workers runtime evicting the object memory between aggregate requests. Presumably this is unlikely.

Why do we think this would be unlikely? I would expect the runtime to reap memory after every single independent request to keep resource usage down.

cjpatton:
> Why do we think this would be unlikely? I would expect the runtime to reap memory after every single independent request to keep resource usage down.

Anecdotally, an instance of a DO often remains in storage hours after last use, but there's certainly no guarantee of this. In fact, we could trigger this bug by re-deploying the Worker in the midst of an aggregation job.

I've update the PR to persist the memory. I left garbage collection out, since the extra DO->DO communication is quite expensive.

cjpatton:
We're better off with a solution to #94.

PR #90, opened 2022-08-15
Remove `SystemTime::now` from test

### LINKED ISSUE #54: Clean up: Remove timestamps from unit tests

nakatsuka-y:
Closes #54.

`roles_test.rs` required tests to get a timestamp when issuing requests, which could cause tests to be flaky.

`ReportStore` is changed to drain reports from `pending` in FIFO.
Reports may not have a strict temporal order, but should be fine as long as each report has a chance to be processed.

`MockAggregator` includes `now` (similar to `TestRunner` in Daphne Workers) and tests use this instead of calling `MockAggregator::get_current_time`, which should only be called when checking the interval boundary.

### DIFF
--- a/daphne/src/roles_test.rs
+++ b/daphne/src/roles_test.rs
@@ -15,10 +15,10 @@ use crate::{
     },
     roles::{DapAggregator, DapHelper, DapLeader},
     testing::{
-        BucketInfo, MockAggregateInfo, MockAggregator, ReportStore, COLLECTOR_BEARER_TOKEN,
+        AggStoreState, BucketInfo, MockAggregateInfo, MockAggregator, COLLECTOR_BEARER_TOKEN,
         HPKE_RECEIVER_CONFIG_LIST, LEADER_BEARER_TOKEN,
     },
-    DapAbort, DapCollectJob, DapError, DapLeaderTransition, DapMeasurement, DapRequest,
+    DapAbort, DapAggregateShare, DapCollectJob, DapLeaderTransition, DapMeasurement, DapRequest,
     DapTaskConfig, DapVersion, Prio3Config, VdafConfig,
 };
 use assert_matches::assert_matches;
@@ -154,9 +154,8 @@ impl MockAggregator {
 
         // Construct report.
         let vdaf_config: &VdafConfig = &VdafConfig::Prio3(Prio3Config::Count);
-        let now = self.get_current_time();
         let report = vdaf_config
-            .produce_report(&hpke_config_list, now, task_id, DapMeasurement::U64(1))
+            .produce_report(&hpke_config_list, self.now, task_id, DapMeasurement::U64(1))
             .unwrap();
 
         report
@@ -165,14 +164,12 @@ impl MockAggregator {
     async fn run_test_agg_job(
         &self,
         helper: &MockAggregator,
-        now: u64,
         task_id: &Id,
         task_config: &DapTaskConfig,
     ) {
         // Leader: Store received report to ReportStore.
         let selector = &MockAggregateInfo {
             task_id: task_id.clone(),
-            batch_info: Some(task_config.current_batch_window(now)),
             agg_rate: 1,
         };
         let (task_id, reports) = get_reports!(self, selector);
@@ -374,10 +371,9 @@ async fn http_post_collect_unauthorized_request() {
 #[tokio::test]
 async fn http_post_aggregate_failure_hpke_decrypt_error() {
     let helper = MockAggregator::new();
-    let now = helper.get_current_time();
     let report_shares = vec![ReportShare {
         nonce: Nonce {
-            time: now,
+            time: helper.now,
             rand: [1; 16],
         },
         extensions: Vec::default(),
@@ -451,18 +447,14 @@ async fn http_post_aggregate_failure_batch_collected() {
 
     // This is to ensure that the lock is released before we call http_post_aggregate.
     {
-        let mut report_store_mutex_guard = helper.report_store.lock().expect("lock() failed");
-        let report_store = report_store_mutex_guard.deref_mut();
-
-        // Create a new instance of report store associated to the bucket_info.
+        let mut agg_store_mutex_guard = helper.agg_store.lock().expect("lock() failed");
+        let agg_store = agg_store_mutex_guard.deref_mut();
         let bucket_info = BucketInfo::new(task_config, task_id, &report.nonce);
-        report_store.insert(bucket_info.clone(), ReportStore::new());
-
-        // Intentionally mark report store as `collected`.
-        report_store
-            .get_mut(&bucket_info)
-            .expect("report_store not found")
-            .process_mark_collected();
+        let agg_store_state = AggStoreState {
+            agg_share: DapAggregateShare::default(),
+            collected: true,
+        };
+        agg_store.insert(bucket_info, agg_store_state);
     }
 
     // Get AggregateResp and then extract the transition data from inside.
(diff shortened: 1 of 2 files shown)


daphne/src/roles_test.rs
```diff
@@ -454,15 +449,11 @@ async fn http_post_aggregate_failure_batch_collected() {
...
-        report_store
-            .get_mut(&bucket_info)
-            .expect("report_store not found")
-            .process_mark_collected();
+        // Create a new instance of report store associated to the task which is intentionally
+        // marked as `collected`.
+        let mut store = ReportStore::new();
+        store.collected = true;
+        report_store.insert(task_id.clone(), store);
     }
```
cjpatton, line 457: Why is this code in its own block? Could we get rid of this?
```suggestion
```

daphne/src/roles_test.rs
```diff
@@ -614,42 +581,24 @@ async fn get_reports_empty_response() {
         .await
         .expect("upload failed unexpectedly");
 
-    // Get report.
-    let now = report.nonce.time - task_config.min_batch_duration - 1;
+    // We should get one report.
```
cjpatton, line 584: Why? The The name of the test is `get_reports_empty_response()`. Also, it looks like the output is not examiined. Let's take a step back and ask ourselves what this test case is for.

daphne/src/roles_test.rs
```diff
@@ -704,7 +653,7 @@ async fn http_post_collect_fail_invalid_batch_interval() {
     let leader = MockAggregator::new();
     let task_id = leader.nominal_task_id();
     let task_config = leader.get_task_config_for(task_id).unwrap();
-    let now = leader.get_current_time();
```
cjpatton, line 707: You could continue to use `get_current_time()`, but just have it always return the same value.

daphne/src/testing.rs
```diff
@@ -329,11 +336,8 @@ impl DapAggregator<BearerToken> for MockAggregator {
             .map_err(|e| DapError::Fatal(e.to_string()))?;
         let report_store = report_store_mutex_guard.deref_mut();
 
-        for (inner_bucket_info, store) in report_store.iter_mut() {
-            if task_id == &inner_bucket_info.task_id
-                && batch_interval.start <= inner_bucket_info.window
-                && batch_interval.end() > inner_bucket_info.window
-            {
+        for (id, store) in report_store.iter_mut() {
```
cjpatton, line 339: This doesn't look right. Why would we mark every report set for the task as collected?

cjpatton (changes requested):

daphne/src/roles_test.rs
nakatsuka-y: This is necessary because of there is no way to release mutex locks in Rust.

daphne/src/roles_test.rs
nakatsuka-y: Updated the test case.

daphne/src/roles_test.rs
nakatsuka-y: We are using the `now` that is initialized when creating a new `MockAggregator` via `MockAggregator.now`

daphne/src/testing.rs
nakatsuka-y: The new change fixes this. We are now checking the AggStore.

daphne/src/testing.rs
```diff
@@ -374,28 +374,32 @@ impl DapHelper<BearerToken> for MockAggregator {
...
-
-            if !report_store.contains_key(&bucket_info) {
-                report_store.insert(bucket_info.clone(), ReportStore::new());
+            // Check whether the report is collected via the AggStore.
+            let bucket_info = BucketInfo::new(task_config, task_id, &report_share.nonce);
+            let collected = match agg_store.get(&bucket_info) {
+                Some(agg_store_state) => agg_store_state.collected,
+                None => false,
+            };
+            if collected {
```
cjpatton, line 383: More idiomatic
```suggestion
            if matches!(agg_store.get(&bucket_info), Some(agg_store_state) if agg_store_state.collected) {
```

daphne/src/testing.rs
```diff
@@ -374,28 +374,32 @@ impl DapHelper<BearerToken> for MockAggregator {
...
-            {
-                Ok(()) => (),
-                Err(failure_reason) => {
-                    early_fails.insert(report_share.nonce.clone(), failure_reason);
-                }
+            // Check whether the same report has been submitted in the past.
+            let report_store = report_store
+                .entry(task_id.clone())
+                .or_insert_with(ReportStore::new);
+
```
cjpatton, line 394: nit: Spurious newline

daphne/src/testing.rs
```diff
@@ -467,54 +471,48 @@ impl DapLeader<BearerToken> for MockAggregator {
...
-                .get(task_id)
-                .ok_or_else(|| DapError::fatal("task not found"))?,
-            task_id,
-            &report.nonce,
-        );
+        let task_config = self
+            .get_task_config_for(task_id)
+            .ok_or_else(|| DapError::fatal("task not found"))?;
+
+        // Check whether the report is collected via the AggStore.
```
cjpatton, line 478: The code for batch-collected and report-replayed checks for Leader and Helper look pretty similar. Is there any way to refactor so that they share code?

cjpatton (commented):
Looks to be on the right track. Just a few clean up things.

daphne/src/roles_test.rs
```diff
@@ -576,34 +568,10 @@ async fn http_post_upload_fail_send_invalid_report() {
     );
 }
 
-#[tokio::test]
```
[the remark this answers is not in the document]
nakatsuka-y: This is removed because we removed the notion of `batch_interval` from the `selector` and the `get_reports` logic.

daphne/src/testing.rs
nakatsuka-y: Thanks, I didn't know about this way of writing.

daphne/src/testing.rs
nakatsuka-y: Removed :)

daphne/src/testing.rs
nakatsuka-y: The newest commit has implemented a function so that they share code.

daphne/src/testing.rs
```diff
@@ -136,6 +136,27 @@ impl MockAggregator {
         }
     }
```
cjpatton, line 138: `check_report()` is not a very descriptive name. Would you mind adding a comment describing what this function does?

daphne/src/testing.rs
```diff
@@ -360,45 +381,35 @@ impl DapHelper<BearerToken> for MockAggregator {
...
-            if report_store.processed.contains(&report_share.nonce) {
-                early_fails.insert(
-                    report_share.nonce.clone(),
-                    TransitionFailure::ReportReplayed,
-                );
-            }
+            // Check whether Report has been collected or replayed.
+            match self.check_report(&bucket_info, &report_share.nonce, report_store, agg_store) {
+                Ok(()) => (),
+                Err(transition_failure) => {
```
cjpatton, line 407:
```suggestion
            if let Err(transition_failure) = self.check_report(&bucket_info, &report_share.nonce, report_store, agg_store) {
```

daphne/src/testing.rs
```diff
@@ -474,37 +485,34 @@ impl DapLeader<BearerToken> for MockAggregator {
...
-
         let report_store = report_store
             .entry(task_id.clone())
             .or_insert_with(ReportStore::new);
 
-        if report_store.processed.contains(&report.nonce) {
-            return Err(DapError::Transition(TransitionFailure::ReportReplayed));
-        }
+        // Check whether Report has been collected or replayed.
+        match self.check_report(&bucket_info, &report.nonce, report_store, agg_store) {
```
cjpatton, line 508: Same comment here

daphne/src/roles_test.rs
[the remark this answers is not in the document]
cjpatton: Ohhh duh, got it.

daphne/src/testing.rs
nakatsuka-y: I've added the comment.

daphne/src/testing.rs
nakatsuka-y: I agree this is much better.

cjpatton (commented):
Looks good, go ahead and squash.

cjpatton (approved):

PR #92, opened 2022-08-15
Fix internal API URL formatting

### LINKED ISSUE #80: Clean up: Proper formatting for internal APIs in e2e tests
We are currently using a hack-y way to remove the version number from the URL.

This should be handled explicitly and clearly.

_Originally posted by @cjpatton in https://github.com/cloudflare/daphne/pull/77#discussion_r940783353_

nakatsuka-y:
Closes #80.

This change fixes the construction of internal API URLs.
Instead of using some hack-y way, we construct the URL so that it is clear.

### DIFF
--- a/daphne_worker_test/tests/test_runner.rs
+++ b/daphne_worker_test/tests/test_runner.rs
@@ -394,12 +394,10 @@ impl TestRunner {
         client: &reqwest::Client,
         agg_info: &InternalAggregateInfo,
     ) -> DapLeaderProcessTelemetry {
-        // TODO(nakatsuka-y) Make the following explicit and clear.
-        // The URL below is made sort of a hack-y way where we use url::Url
-        // deleting the path from the base URL when calling "join()" to our
-        // advantage to prevent us from having "version" in this URL.
-        // Ref: https://github.com/cloudflare/daphne/issues/80
-        let url = self.leader_url.join("/internal/process").unwrap();
+        // Replace path "/v01" with "/internal/process".
+        let mut url = self.leader_url.clone();
+        url.set_path("internal/process");
+
         let resp = client
             .post(url.as_str())
             .body(serde_json::to_string(&agg_info).unwrap())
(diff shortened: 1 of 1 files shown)


daphne_worker_test/tests/test_runner.rs
```diff
@@ -394,12 +394,13 @@ impl TestRunner {
...
         agg_info: &InternalAggregateInfo,
     ) -> DapLeaderProcessTelemetry {
-        // TODO(nakatsuka-y) Make the following explicit and clear.
-        // The URL below is made sort of a hack-y way where we use url::Url
-        // deleting the path from the base URL when calling "join()" to our
-        // advantage to prevent us from having "version" in this URL.
-        // Ref: https://github.com/cloudflare/daphne/issues/80
-        let url = self.leader_url.join("/internal/process").unwrap();
+        // Obtain base URL.
+        let mut url = self.leader_url.join("/").unwrap();
```
cjpatton, line 398: This isn't much clearer, in my opinion. We're still exploiting the same behavior, which, by the way, is documented [here](https://docs.rs/url/latest/url/struct.Url.html#method.join):

> Note: a trailing slash is significant. Without it, the last path component is considered to be a “file” name to be removed to get at the “directory” that is used as the base.

For Daphne-Worrker, the internal endpoints are always going to be at the base of the tree. A better approach might be to construct the URL from the protocol (e.g, "http" or "https"), hostname, and port: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=176456b8dea0f6242b9680c4b66869c2

cjpatton (changes requested):

daphne_worker_test/tests/test_runner.rs
nakatsuka-y: I didn't know that this method existed, thanks for pointing it out!

cjpatton (commented):
Looks good. I'll wait until the base PR is merged before apprvoing. (You will need to rebase.)

cjpatton (approved):

PR #95, opened 2022-08-23
Unblock DO requests in `roles.rs`

### LINKED ISSUE #93: Performance: Concurrency improvements for Helper
1. When handling AggregateInitializeReq in `http_post_aggregate()`, the helper calls `mark_aggregated()`, awaits the result, then processes the reports via `handle_agg_init_req()`. The first call is used to filter out any reports from the request that need to be rejected, e.g., for anti-replay protection. This sequencing is fairly wasteful, since `mark_aggregated()` needs to wait for a response from backend storage. It would be better to run `mark_aggregated()` in parallel with `handle_agg_init_req()`, which constitutes the bulk of the heavy VDAF/HPKE operations. Once we get the result from `mark_aggregated()`, we would remove the rejected reports from the AggregateResp before sending to the Leader.
2. The current implementation of `DapAggregator::put_out_shares()` for `DaphneWorker` serializes requests to different instances of the AggregateStore DO. These could be issued in parallel instead. Similarly for `get_agg_share()` and `mark_collected()`.

nakatsuka-y:
Closes #93.

This PR allows the Helper to submit DO requests while doing other things concurrently.

`http_post_aggregate` in `roles.rs` has been changed so that `mark_aggregated` is processed while `get_helper_state` is called.

### DIFF
--- a/daphne/src/roles.rs
+++ b/daphne/src/roles.rs
@@ -13,7 +13,7 @@ use crate::{
     messages::{
         constant_time_eq, AggregateContinueReq, AggregateInitializeReq, AggregateResp,
         AggregateShareReq, AggregateShareResp, CollectReq, CollectResp, Id, Interval, Nonce,
-        Report, ReportShare, TransitionFailure,
+        Report, ReportShare, TransitionFailure, TransitionVar,
     },
     DapAbort, DapAggregateShare, DapCollectJob, DapError, DapGlobalConfig, DapHelperState,
     DapHelperTransition, DapLeaderProcessTelemetry, DapLeaderTransition, DapOutputShare,
@@ -562,16 +562,14 @@ pub trait DapHelper<S>: DapAggregator<S> {
                     return Err(DapAbort::InvalidProtocolVersion);
                 }
 
-                let early_rejects = self
-                    .mark_aggregated(&agg_init_req.task_id, &agg_init_req.report_shares)
-                    .await?;
+                // Run mark_aggregated and handle_agg_init_req concurrently.
+                let early_rejects_future =
+                    self.mark_aggregated(&agg_init_req.task_id, &agg_init_req.report_shares);
 
-                // This is where the encrypted report share is decrypted.
                 let transition = task_config.vdaf.handle_agg_init_req(
                     self,
                     &task_config.vdaf_verify_key,
                     &agg_init_req,
-                    &early_rejects,
                 )?;
 
                 // Check that helper state with task_id and agg_job_id does not exist.
@@ -582,8 +580,30 @@ pub trait DapHelper<S>: DapAggregator<S> {
                     ));
                 }
 
+                // Remove reports that are rejected early.
+                let early_rejects = early_rejects_future.await?;
                 let agg_resp = match transition {
-                    DapHelperTransition::Continue(state, agg_resp) => {
+                    DapHelperTransition::Continue(mut state, mut agg_resp) => {
+                        let mut i = 0;
+                        while i < state.seq.len() {
+                            if let Some(failure) = early_rejects.get(&agg_resp.transitions[i].nonce)
+                            {
+                                // Mark reports that were rejected early as TransitionVar::Failed.
+                                agg_resp.transitions[i].var = TransitionVar::Failed(*failure);
+
+                                // Remove VDAF preparation state of reports that were rejected early.
+                                if state.seq[i].1 == agg_resp.transitions[i].nonce {
+                                    let _val = state.seq.remove(i);
+                                } else {
+                                    // The nonce in the Helper state and Aggregate response must be aligned.
+                                    // Abort with an internal error if this is not the case.
+                                    return Err(DapError::fatal("nonces not aligned").into());
+                                }
+                            } else {
+                                i += 1;
+                            }
+                        }
+
                         self.put_helper_state(
                             &agg_init_req.task_id,
                             &agg_init_req.agg_job_id,
--- a/daphne/src/roles_test.rs
+++ b/daphne/src/roles_test.rs
@@ -15,8 +15,8 @@ use crate::{
     },
     roles::{DapAggregator, DapHelper, DapLeader},
     testing::{
-        AggStoreState, BucketInfo, MockAggregateInfo, MockAggregator, COLLECTOR_BEARER_TOKEN,
-        HPKE_RECEIVER_CONFIG_LIST, LEADER_BEARER_TOKEN,
+        AggStoreState, BucketInfo, MockAggregateInfo, MockAggregator, ReportStore,
+        COLLECTOR_BEARER_TOKEN, HPKE_RECEIVER_CONFIG_LIST, LEADER_BEARER_TOKEN,
     },
     DapAbort, DapAggregateShare, DapCollectJob, DapLeaderTransition, DapMeasurement, DapRequest,
     DapTaskConfig, DapVersion, Prio3Config, VdafConfig,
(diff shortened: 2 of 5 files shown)


nakatsuka-y:
To add a bit of context on the `for` loop changes:
Previously we were defining and computing the `stub` within the `for` loop that iterates over the necessary variables.
`durable_post` is called within this loop while taking in the `stub` as an argument.
Therefore, if we try to remove `await` from `durable_post`, we cannot compile to code because the defined `stub` goes out of scope even though `durable_post` is still using it.

The new changes precompute the `stub` and put it into an array where we can reference it when calling `durable_post`.

cjpatton (changes requested):

daphne/src/roles.rs
```diff
@@ -582,15 +580,35 @@ pub trait DapHelper<S>: DapAggregator<S> {
                     ));
                 }
 
+                // Using early_rejects and transition, replace any element in transition
+                // that has the same nonce in early_rejects.
```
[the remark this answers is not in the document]
nakatsuka-y: Updated comment.

daphne/src/roles.rs
```diff
@@ -582,15 +580,35 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+                let early_rejects = early_rejects_future.await?;
                 let agg_resp = match transition {
                     DapHelperTransition::Continue(state, agg_resp) => {
-                        self.put_helper_state(
+                        // Run put_helper_state while running the Transition check.
```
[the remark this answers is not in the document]
nakatsuka-y: Although `DapHelperState.seq` is a vector of `VdafState` and `Nonce`, retrieving an element based on `transition.nonce` is much more efficient if we use a hashmap.
Changing the `DapHelperState.seq` into a hashmap will clutter this PR, so I have left a TODO note and convert the vector into a hashmap here.
Let me know your thoughts on this.

daphne/src/roles.rs
```diff
@@ -582,15 +580,35 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+                        for transition in agg_resp.transitions {
+                            if let Some(failure) = early_rejects.get(&transition.nonce) {
+                                transitions.push(Transition {
+                                    nonce: transition.nonce,
+                                    var: TransitionVar::Failed(*failure),
+                                });
+                            } else {
+                                transitions.push(transition);
+                            }
+                        }
```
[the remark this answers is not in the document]
nakatsuka-y: I totally forgot that you can reference a value like this. Thanks for pointing this out!

daphne/src/roles_test.rs
```diff
@@ -427,6 +431,51 @@ async fn http_post_aggregate_transition_continue() {
...
+        extensions: report.extensions,
+        // 1st share is for Leader and the rest is for Helpers (note that there is only 1 helper).
+        encrypted_input_share: report.encrypted_input_shares[1].clone(),
+    }];
+    let sender_auth = Some(BearerToken::from(LEADER_BEARER_TOKEN.to_string()));
+
+    let mut req = helper.gen_test_agg_init_req(report_shares);
+    req.sender_auth = sender_auth;
+
+    // This is to ensure that the lock is released before we call http_post_aggregate.
```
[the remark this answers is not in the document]
nakatsuka-y: Made comment more explicit.

daphne/src/roles.rs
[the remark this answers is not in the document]
nakatsuka-y: I just realized that removing `VdafState` from `DapHelperState.seq` is causing some Daphne Worker unit tests to fail.
Investigating them right now.

daphne/src/roles.rs
```diff
@@ -582,14 +581,44 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
 
+                // Remove reports that are rejected early.
+                let early_rejects = early_rejects_future.await?;
                 let agg_resp = match transition {
-                    DapHelperTransition::Continue(state, agg_resp) => {
+                    DapHelperTransition::Continue(state, mut agg_resp) => {
+                        // state.seq is a vector of VdafState and Nonce. To efficiently access and remove
+                        // unwanted VdafState, we convert the vector into a HashMap.
+                        //
+                        // TODO(nakatsuka-y) Change DapHelperState.seq from Vector to HashMap.
```
cjpatton, line 591: Building this hash map is going to be expensive if the number of reports is large, say >100. We should be able to avoid this.

Consider the following algorithm for deleting `state`s (from rust docs):
```rust
let mut i = 0;
while i < vec.len() {
    if some_predicate(&mut vec[i]) {
        let val = vec.remove(i);
        // your code here
    } else {
        i += 1;
    }
}
```

Since `states.seq` and `agg_resp.transitions` are in 1:1 correspondence, it should be possible to extend this code to iterate over `agg_resp.transitions`.

cjpatton (changes requested):

daphne/src/roles.rs
nakatsuka-y: The updated code should reflect this.

daphne/src/roles.rs
```diff
@@ -582,14 +580,33 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
-                    DapHelperTransition::Continue(state, agg_resp) => {
+                    DapHelperTransition::Continue(mut state, mut agg_resp) => {
+                        let mut i = 0;
+                        while i < state.seq.len() {
+                            if let Some(failure) = early_rejects.get(&agg_resp.transitions[i].nonce)
+                            {
+                                // Mark reports that were rejected early as TransitionVar::Failed.
+                                agg_resp.transitions[i].var = TransitionVar::Failed(*failure);
+
+                                // Remove VdafState of reports that were rejected early.
```
cjpatton, line 594:
```suggestion
                                // Remove VDAF preparation state of reports that were rejected early.
```

daphne/src/roles.rs
```diff
@@ -582,14 +580,33 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+                                if state.seq[i].1 == agg_resp.transitions[i].nonce {
```
cjpatton, line 595: If the nonces *don't* match here, then this is a pretty serious bug. (These nonce sequences are supposed to be the same, as enforced by tests in `vdaf/mod_test.rs`.) If this condition is false, then we should abort with an internal error.

daphne/src/roles.rs
```diff
@@ -582,14 +580,33 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+                            }
+                        }
+
                         self.put_helper_state(
                             &agg_init_req.task_id,
                             &agg_init_req.agg_job_id,
                             &state,
                         )
                         .await?;
+
```
cjpatton, line 609: nit: Remove spurious newline. (It's good practice to keep changes out of PRs that are not necessary.)

cjpatton (commented):
Looks good! A couple more comments to address, then squash and make sure the commit message is up to date.

daphne/src/roles.rs
nakatsuka-y: Fixed :)

daphne/src/roles.rs
nakatsuka-y: Sounds good, I've added an else statement that responds with and error if the nonces don't match.

daphne/src/roles.rs
nakatsuka-y: Sorry :(

cjpatton (changes requested):
nit: The commit message isn't completely accurate:
```
This change prevents functions that send requests to DO in roles.rs from blocking
other functions. Some functions that rely on the output of the DO request are
changed accordingly.
```

In fact, only one function is changed, `http_post_aggregate()`.

nakatsuka-y:
You are right. I've changed the commit message.

cjpatton (approved):

PR #96, opened 2022-08-23
Fix & shrink Miniflare-based test container.

branlwyd:
This container had stopped building successfully (I think the Daphne
codebase started using a Rust feature that wasn't stabilized until after
1.61; bumping to 1.63 was all that was needed to get a successful
build.)

While I was at it, I also split the container build into a "build"
portion and an "output" portion & switched to alpine-based base images
-- this shrunk the output image from 3.32GB to 390MB.

Also go back to building Daphne using the mainline version of workers-rs,
as the referenced issue has been fixed.

### DIFF
--- a/daphne_worker_test/docker/miniflare.Dockerfile
+++ b/daphne_worker_test/docker/miniflare.Dockerfile
@@ -19,6 +19,8 @@ RUN apk add --update npm bash
 RUN npm install -g n && \
     n 18.4.0
 RUN npm install -g miniflare@2.5.1
+COPY daphne_worker_test/leader.env /leader.env
+COPY daphne_worker_test/helper.env /helper.env
 COPY --from=builder /tmp/dap_test/daphne_worker_test/wrangler.toml /wrangler.toml
 COPY --from=builder /tmp/dap_test/daphne_worker_test/build/worker/* /build/worker/
 # `-B ""` to skip build command.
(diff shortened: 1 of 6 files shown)


branlwyd:
(Context: Janus currently tests w/ a precompiled version of Daphne, which among other issues will quickly skew badly out of date. Unless/until a public container is published, I'm hoping to dynamically build a container image for testing. I was able to feed the container built by this dockerfile into our test framework & it passed, so I'm moderately confident that this container works.)

cjpatton (changes requested):
Changes to the `miniflare` container look good! To get the end-to-end tests running via docker-compose, we also need to change `runtests` to use rust 1.63. Would you mind making this change here?

branlwyd:
OK, done -- the `docker-compose` invocation you mention now succeeds. I had to make a few other minor changes to get things running, mentioned in the commit description.

Other than what's mentioned in the commit description, I also:
* Only run the `daphne_worker_test` tests, to approximate only running the E2E tests. I suppose it's not so useful to run unit-tests that would be run by a normal `cargo test` in this case.
* Update the Dockerfile's `cargo` invocation to build with the same profile/features as the real `cargo test` that is run by `test.sh`. This avoids the `test.sh` invocation from having to build things -- all of the Rust compilation happens at `docker build` time now.

I didn't minimize the `runtests` docker image, as it's basically an ephemeral image to drive a test.

cjpatton (approved):
Thanks @branlwyd! I haven't gotten the end-to-end tests to run via docker-compose yet, but since this PR isn't a regression and it unblocks you, I will merge it now and address the issue later (hopefully today).

PR #97, opened 2022-08-23
Unblock DO requests in Daphne Workers

nakatsuka-y:
Previously we were defining and computing the `stub` within the `for` loop that iterates over the necessary variables.
`durable_post` was called within this loop while taking in the `stub` as an argument.
The aim of this change is to send concurrent DO requests within this loop by removing `await` from the `durable_post` within the `for` loop.
However, if we try to remove `await` from `durable_post`, we cannot compile to code because the defined `stub` goes out of scope even though `durable_post` is still using it.

The new changes precompute the `stub` and put it into an array where we can reference it when calling `durable_post`.

### DIFF
--- a/daphne_worker/src/dap.rs
+++ b/daphne_worker/src/dap.rs
@@ -156,23 +156,25 @@ impl<D> DapAggregator<BearerToken> for DaphneWorkerConfig<D> {
         // Check whether the request overlaps with previous requests. This is done by
         // checking the AggregateStore and seeing whether it requests for aggregate
         // shares that have already been marked collected.
-        let namespace = self.durable_object("DAP_AGGREGATE_STORE")?;
+        let mut requests = Vec::new();
         for window in (batch_interval.start..batch_interval.end())
             .step_by(task_config.min_batch_duration.try_into().unwrap())
         {
-            let stub = namespace
-                .id_from_name(&durable_agg_store_name(
-                    &task_config.version,
-                    &task_id.to_hex(),
-                    window,
-                ))?
-                .get_stub()?;
+            let durable_name =
+                durable_agg_store_name(&task_config.version, &task_id.to_hex(), window);
+            requests.push(self.durable_post(
+                BINDING_DAP_AGGREGATE_STORE,
+                DURABLE_AGGREGATE_STORE_GET,
+                durable_name,
+                &(),
+            ));
+        }
 
-            // TODO Don't block on DO requests (issue multiple requests simultaneously).
-            let mut resp = durable_post!(stub, DURABLE_AGGREGATE_STORE_GET, &()).await?;
+        let responses: Vec<AggregateStoreResult> = try_join_all(requests).await?;
 
+        for resp in responses {
             // If this agg share has been collected before, return BatchCollected error.
-            match resp.json().await? {
+            match resp {
                 AggregateStoreResult::Ok(_) => {
                     continue;
                 }
@@ -181,6 +183,7 @@ impl<D> DapAggregator<BearerToken> for DaphneWorkerConfig<D> {
                 }
             }
         }
+
         Ok(false)
     }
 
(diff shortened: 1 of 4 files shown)


daphne_worker/src/dap.rs
```diff
@@ -195,19 +204,31 @@ impl<D> DapAggregator<BearerToken> for DaphneWorkerConfig<D> {
...
+        let mut requests = Vec::with_capacity(stubs_and_shares.len());
+        for stub_and_share in &stubs_and_shares {
+            requests.push(durable_post!(
+                (*stub_and_share).0, // Stub
+                DURABLE_AGGREGATE_STORE_MERGE,
+                &(*stub_and_share).1 // AggregateShare
+            ));
         }
 
+        let _responses = try_join_all(requests).await?;
```
cjpatton, line 230: Let's make sure each response status is 200 and abort with internal error if not.

daphne_worker/src/dap.rs
```diff
@@ -157,20 +157,28 @@ impl<D> DapAggregator<BearerToken> for DaphneWorkerConfig<D> {
...
             .step_by(task_config.min_batch_duration.try_into().unwrap())
         {
-            let stub = namespace
-                .id_from_name(&durable_agg_store_name(
-                    &task_config.version,
-                    &task_id.to_hex(),
-                    window,
-                ))?
-                .get_stub()?;
+            stubs.push(
```
cjpatton, line 164: This approach might work, but it has the disadvantage of adding a lot of code complexity. If you we need to keep the stub in scope, you could have the async function you call create it. In fact, this is what we already do with `DaphneWorkerConfig::durable_post()`. (I think deleting this method is a regression.)

cjpatton (changes requested):

daphne_worker/src/dap.rs
nakatsuka-y: Fixed.

daphne_worker/src/dap.rs
nakatsuka-y: I've reverted the changes and brought back `DaphneWorkerConfig::durable_post()`.

daphne_worker/src/config.rs
```diff
@@ -329,17 +329,17 @@ impl<D> DaphneWorkerConfig<D> {
...
         durable_name: String,
         data: I,
-    ) -> Result<O> {
+    ) -> Result<Response> {
         let namespace = self.durable_object(durable_binding)?;
         let stub = namespace.id_from_name(&durable_name)?.get_stub()?;
-        let mut resp = durable_post!(stub, &durable_path, &data).await?;
-        resp.json().await
+        let resp = durable_post!(stub, &durable_path, &data).await?;
+        Ok(resp)
```
cjpatton, line 342: Can you explain why this change is needed? It seems to complicate the code below unnecessarily.

cjpatton (changes requested):

daphne_worker/src/dap.rs
cjpatton: I'm sorry, I appear to have mislead you here!! There is actually nothing to validate here: If the response was not 200 then parsing the JSON blob from the response would fail and we'd err here.

My bad. What this comment should have been is: There's nothing to validate here, so you can remove the `let` statement.

daphne_worker/src/config.rs
cjpatton: :( https://github.com/cloudflare/daphne/pull/97#discussion_r954391489

nakatsuka-y: Thanks for the feed back.
Apparently, we cannot remove the `let` statement or the Rust compiler will complain.
If we use `resp.json().await`, then we need to explicitly declare what we are expecting in the response `let _response: Vec<Something>`, which makes it very cumbersome when dealing with cases where DO responses are simply empty.
That's why responding with `Ok(resp)` is much simpler, because we can declare `let _response: Vec<Response>` and obtain the result if needed.

I agree that this change makes the code very messy, but I haven't yet found a better way to solve this. Of course, if we can simply get away without declaring a variable in response to `try_join_all()`, then that would be great.

daphne_worker/src/dap.rs
nakatsuka-y: Thanks for clarifying! Please see my response below.

daphne_worker/src/config.rs
cjpatton: Ohhh, I guess what we can do is make sure that each result is equal to the empty string in this case?
```rust
for result in try_join_all!(requests).await? {
    if result != "" {
        // abort
    }
}
```

cjpatton: Does this make sense?

nakatsuka-y: There is an issue in the implementation here:
In the DO AggregateStore, [some responses](https://github.com/cloudflare/daphne/blob/6228556c7b87a7fe85e414a3186a2511407896f0/daphne_worker/src/durable/aggregate_store.rs#L84) are encoded in JSON and [some](https://github.com/cloudflare/daphne/blob/6228556c7b87a7fe85e414a3186a2511407896f0/daphne_worker/src/durable/aggregate_store.rs#L74) return with an empty response.
In Cloudflare Workers, `from_json` sets the HTTP header to have `application/json` (see [here](https://github.com/cloudflare/workers-rs/blob/main/worker/src/response.rs#L41)) while the empty response does not populate the HTTP header with any content type.

When we run `resp.json().await`, `resp.json()` assumes that `resp` has the HTTP header set to `application/json` and errors out if it doesn't.

We have several ways to deal with responses that do and don't have this content-type:
1. return `Ok(resp)` instead of `resp.json().await` and have the caller decide whether to deserialize, or
2. change AggregateStore so that we always return a JSON body, but just an empty one

Which do you think is better?

cjpatton: Thanks for the very clear description of the problem, plus options. I think (2.) is better.

nakatsuka-y: Thank you for the suggestion! I've implemented this in the new commit.

daphne_worker/src/dap.rs
nakatsuka-y: This should be reflected in the latest commit.

cjpatton (commented):
Awesome, this looks good. The scope has creeped a little bit, but I think in a way that's ultimately useful. Go ahead and squash and I'll apprvoe.

nakatsuka-y:
Squashed!

cjpatton (approved):

PR #99, opened 2022-08-25
Trigger don't-clear-the-agg-job-queue behavior on a new variable.

branlwyd:
Previously, it was triggered if DAP_DEPLOYMENT was set to "dev".
However, DAP_DEPLOYMENT also controls whether various parts of the
codebase override hostnames to localhost. I think that controlling both
aspects with one variable did not capture all the use-cases:

* Local dev, based on running a miniflare process on a developer
  workstation, will want both the localhost-override & the
  don't-clear-agg-job-queue behavior.
* Container-based testing, as in the E2E tests, wants the
  don't-clear-agg-job-queue behavior but not the localhost-override.
* Production deployments want neither the localhost-override nor the
  don't-clear-agg-job-queue behavior.

### DIFF
--- a/daphne_worker/src/durable/report_store.rs
+++ b/daphne_worker/src/durable/report_store.rs
@@ -170,8 +170,16 @@ impl DurableObject for ReportStore {
                 //   processed. This way we can avoid storing "agg_job" here.
                 //
                 // In the meantime, to avoid flakyness in interop tests, supress the bug in
-                // development environments by never clearing the agg job queue.
-                if empty && self.env.var("DAP_DEPLOYMENT")?.to_string() != "dev" {
+                // development environments by never clearing the agg job queue if a workaround flag
+                // is set. Note that it is necessary to restart miniflare between test runs in order
+                // to reset the DO state.
+                if empty
+                    && self
+                        .env
+                        .var("DAP_ISSUE73_DISABLE_AGG_JOB_QUEUE_GARBAGE_COLLECTION")?
+                        .to_string()
+                        == "true"
+                {
                     let agg_job: Option<AggregationJob> = state_get(&self.state, "agg_job").await?;
                     if let Some(agg_job) = agg_job {
                         // NOTE There is only one agg job queue for now. In the future, work will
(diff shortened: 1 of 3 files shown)


branlwyd:
(another approach would be to introduce a new DAP_DEPLOYMENT value, e.g. `container_dev`, which applies the don't-clear-agg-job-queue behavior but not the override. IMO, the current approach is simpler, especially since the need for the don't-clear-agg-job-queue behavior will eventually disappear entirely -- with this implementation strategy, we'll just be able to rm the entire variable, rather than needing to coalesce DAP_DEPLOYMENT values)

daphne_worker/src/durable/report_store.rs
```diff
@@ -170,8 +170,9 @@ impl DurableObject for ReportStore {
                 //   processed. This way we can avoid storing "agg_job" here.
                 //
                 // In the meantime, to avoid flakyness in interop tests, supress the bug in
-                // development environments by never clearing the agg job queue.
-                if empty && self.env.var("DAP_DEPLOYMENT")?.to_string() != "dev" {
+                // development environments by never clearing the agg job queue if a workaround flag
+                // is set.
```
cjpatton, line 174:
```suggestion
                // is set. Note that it is necessary to restart miniflare between test runs in order to reset the DO state.
```

daphne_worker_test/helper.env
```diff
@@ -1,5 +1,6 @@
 DAP_DEPLOYMENT = "prod"
 DAP_AGGREGATOR_ROLE = "helper"
+DISABLE_JOB_QUEUE_FINISH = "true"
```
cjpatton, line 3: 1. The convention is to prefix variables with "DAP_" that impact Daphne-Worker's behavior.
2. Maybe we should give it a name that points to the fact that this hacks around a bug. Maybe `DAP_ISSUE73_DISABLE_AGG_JOB_QUEUE_GARBAGE_COLLECTION`. (being overly verbose is an asset here, I think.)

cjpatton (changes requested):

daphne_worker_test/helper.env
branlwyd: Good idea, I went with your suggestion of `DAP_ISSUE73_DISABLE_AGG_JOB_QUEUE_GARBAGE_COLLECTION`.

cjpatton (commented):
Nice! Plaease squash and I'll approve.

branlwyd:
Squashed!

cjpatton (approved):

PR #103, opened 2022-08-29
`daphne`: Remove `worker` dependency

cjpatton:
Based on #102 (merge that first).

Remove `From<worker::Error` for `DapError`, thereby eliminating the
`worker` dependency from `daphne` altogether. Instead, implement
conversion explicitly in the `daphne_worker` crate.

This results in more verbose code, but the upside is that we no longer
need to synchronize the version of `worker` to use across the crates.
This is useful while workers-rs is under active development.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -510,7 +510,6 @@ dependencies = [
  "thiserror",
  "tokio",
  "url",
- "worker",
 ]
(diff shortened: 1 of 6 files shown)


armfazh (approved):
good one!

PR #108, opened 2022-08-31
Prepare to release daphne_worker 0.1.1

### LINKED ISSUE #107: Compute DO alarm offset properly
Once this change lands, we'll need to update our code accordingly: https://github.com/cloudflare/workers-rs/pull/216

Specifically, here we need to use an offset instead of the current time plus the offset: https://github.com/cloudflare/daphne/blob/main/daphne_worker/src/durable/helper_state_store.rs#L66

cjpatton:
Closes #107.

The main change is upgrading to `worker` version 0.0.11, which is necessary in order to publish a new version of the `daphne_worker` crate. The new version of `worker` includes a bugfix for DO alarams, which allows us to also close #107.

In order to get `daphne_worker` to compile to WASM, we need to ensure we depend on a single version of the `worker` crate. But since `daphne_worker` uses `daphne` version 0.1.0, which still has the `worker` dependency, we end up building two version of the `worker`. To resolve this, a new version of `daphne` was cut with the `worker` dependency removed. This has already been published: https://crates.io/crates/daphne/0.1.1/dependencies

Next steps after merging this change:
* publish `daphne_worker` version 0.1.1 (try `cargo publish --dry-run` to make sure this will work)
* Yank `daphne` version 0.1.0 and `daphne_worker` version 0.1.0 from crates.io so that  no one tries to use them.


armfazh (approved):

cjpatton:
New crate published: https://crates.io/crates/daphne_worker/0.1.1
Old versions of daphne and daphne_worker yanked.

PR #111, opened 2022-09-02
Fix "type" field in Problem Details

### LINKED ISSUE #110: Problem details type field is not a URI
Currently, Daphne outputs the following in response to a DapAbort error:

```json
{"type":"batchMismatch"}
```

This is missing the `urn:ietf:params:ppm:dap:error:` prefix, [called for in the preamble before Table 1](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-01.html#section-3.1-2).

cjpatton:
Closes #110.

The spec calls for this field to be URI and to be prefixed by
"urn:ietf:params:ppm:dap:error:". This change corrects this.

While at it, bump crate versions in preparation for releasing a new
version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -490,7 +490,7 @@ dependencies = [
 
 [[package]]
 name = "daphne"
-version = "0.1.1"
+version = "0.1.2"
 dependencies = [
  "anyhow",
  "assert_matches",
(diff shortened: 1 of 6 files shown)


armfazh (approved):

cjpatton:
@divergentdave let me know when you need this!

divergentdave:
We don't depend on this yet, so no rush. I just noticed it while debugging.

Issue #109, opened 2022-09-02
Helper in daphne_worker_test may lose reports in AggregateStore


divergentdave:
We are testing Janus against Daphne in an integration test, and we have noticed an intermittent failure. (I can more reliably trigger the failure when my CPU has high load) We're testing against commit e1b503e, but I think the issue has not yet been addressed on main. When it receives a merge request, AggregateStore tries to read from the "agg_share" key, uses an empty aggregate share if it's not yet present, updates the share, and writes it back. I updated this method as follows with some logging, and I saw the following leading up to a test failure.

```rust
            (DURABLE_AGGREGATE_STORE_MERGE, Method::Post) => {
                let mut agg_share: DapAggregateShare =
                    state_get_or_default(&self.state, "agg_share").await?;
                let count_before = agg_share.report_count();
                let agg_share_delta = req.json().await?;
                agg_share.merge(agg_share_delta).map_err(int_err)?;
                let count_after = agg_share.report_count();
                self.state.storage().put("agg_share", agg_share).await?;
                console_log!(
                    "aggregate store DO merge, updated from {} reports to {} reports",
                    count_before,
                    count_after
                );
                Response::from_json(&String::new())
            }
```

```
aggregate store DO merge, updated from 0 reports to 12 reports
...
aggregate store DO merge, updated from 0 reports to 13 reports
...
aggregate store DO merge, updated from 12 reports to 33 reports
```

Later, the leader sends an aggregate share request for 46 reports, but the Daphne helper returns an error due to a batch mismatch, as it is expecting 33 reports.

Clearly there is some issue with the transactionality of this DO's storage accesses, but it's hard for me to say whether it's a bug in Daphne or Miniflare, i.e. whether the same would occur when running on Cloudflare's real runtime. The DO API documentation says that "a series of reads followed by a series of writes (with no other intervening I/O) are automatically atomic and behave like a transaction." When the DO tries to read "agg_share" and gets an error with "No such value in storage.", does that count as a read or not for the purposes of blocking other concurrent event? In other words, does the transactional storage API handle phantom reads correctly? There is a `transaction()` method available, which takes a callback, and runs it in an explicit transaction. It's possible the phantom read behavior is different with that. (again, this could differ on Miniflare and Cloudflare)

cjpatton:
[...]

PR #112, opened 2022-09-02
Reorder I/O calls in DURABLE_AGGREGATE_STORE_MERGE

divergentdave:
This reorders the I/O operations done by AggregateStore, in order to address #109. I tested this for a while locally, and I was no longer able to reproduce the same issue. It would be nice to add some sort of stress test that just created a bevy of AggregateStore DOs on top of miniflare and threw DURABLE_AGGREGATE_STORE_MERGE requests at them, but I'm not sure where that should go, and how much it would take to expose DO requests to an integration test.

### DIFF
--- a/daphne_worker/src/durable/aggregate_store.rs
+++ b/daphne_worker/src/durable/aggregate_store.rs
@@ -66,11 +66,17 @@ impl DurableObject for AggregateStore {
 
         match (req.path().as_ref(), req.method()) {
             (DURABLE_AGGREGATE_STORE_MERGE, Method::Post) => {
+                let agg_share_delta = req.json().await?;
+
+                // To keep this pair of get and put operations atomic, there should be no await
+                // points between them. See the note below `transaction()` on
+                // https://developers.cloudflare.com/workers/runtime-apis/durable-objects/#transactional-storage-api.
+                // See issue #109.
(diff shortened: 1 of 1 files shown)


daphne_worker/src/durable/aggregate_store.rs
```diff
@@ -66,11 +66,16 @@ impl DurableObject for AggregateStore {
 
         match (req.path().as_ref(), req.method()) {
             (DURABLE_AGGREGATE_STORE_MERGE, Method::Post) => {
+                let agg_share_delta = req.json().await?;
+
+                // To keep this pair of get and put operations atomic, there should be no await
+                // points between them. See the note below `transaction()` on
+                // https://developers.cloudflare.com/workers/runtime-apis/durable-objects/#transactional-storage-api.
```
cjpatton, line 73:
```suggestion
                // https://developers.cloudflare.com/workers/runtime-apis/durable-objects/#transactional-storage-api.
                // See issue #109.
```

cjpatton (commented):
Stress testing would be useful. I suppose they could live in `daphne_worker_test/test` behind a feature. This will have to wait, however.

cjpatton (approved):

cjpatton:
Please squash before merging!

PR #113, opened 2022-09-02
Remove HPKE Receiver Config from environment file

### LINKED ISSUE #98: Daphne-Worker: Store `HpkeReceiverConfig` list in KV
Currently the Aggregator's HPKE receiver config list is stored as a secret:
https://github.com/cloudflare/daphne/blob/main/daphne_worker/src/config.rs#L69-L73

What we want to do instead is store the list in KV. The HPKE secret keys should be encrypted under a symmetric key, which in turn is stored as a secret. Requirements:
* Encrypt each `HpkeReceiverConfig` individually. Ideally only the HPKE secret key would be encrypted. That way you can retrieve an HPKE config without decrypting.
* Store each `HpkeReceiverConfig` separately. Use config ID for the look-up key.
* When a Client requests an HPKE config, always return the first config in the list.
* We will use the same set of HPKE configs for many tasks.
* Use https://docs.rs/chacha20poly1305/0.2.1/chacha20poly1305/struct.XChaCha20Poly1305.html with a random nonce.
* `HpkeDecrypter::hpke_decrypt()` needs to be async. (And maybe other methods on this trait, too.)
* For now, we only need to generate and store one `HpkeRecevierConfig` at a time. If the config does not exist in KV, it should be created as needed.

nakatsuka-y:
Closes #98.

The intent of this PR is to remove HPKE Receiver Configs from environment files (i.e., `helper.env` and `leader.env`) and store them on Cloudflare KV.
The current implementation requires the Aggregator to generate a new HPKE Receiver Config when it detects that there is none in the KV, but in the future this would be created through `wrangler`.
This changes allows HPKE public and private keys easier to rotate.


nakatsuka-y:
One thing to note is that the algorithm that returns the first HPKE Receiver Config if multiple exists is not quite right, because the ordering of the KV keys is not temporal (https://developers.cloudflare.com/workers/runtime-apis/kv/#ordering).
I am thinking of way to achieve this correctly.

cjpatton (commented):
To reduce latency, let's cache HpkeReceiverConfig's that we fetch from KV. That way we don't have to make a request to KV each time we need a given config.

daphne_worker_test/wrangler.toml
```diff
@@ -28,6 +28,11 @@ bindings = [
   { name = "DAP_GARBAGE_COLLECTOR", class_name = "GarbageCollector" },
 ]
 
+[[kv_namespaces]]
+  binding = "DAP_HPKE_CONFIG_STORE"
```
cjpatton, line 32: Be sure to update `daphne_worker` accordingly.
```suggestion
  binding = "DAP_HPKE_RECEIVER_CONFIG_STORE"
```

daphne/dapf/src/lib.rs
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) 2022 Cloudflare, Inc. All rights reserved.
```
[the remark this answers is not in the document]
nakatsuka-y: I forgot to remove the `lib` section from the Cargo.toml. I've removed this and made `dapf` a binary.

Cargo.toml
```diff
@@ -6,6 +6,7 @@ members = [
     "daphne",
     "daphne_worker",
     "daphne_worker_test",
+    "daphne/dapf",
```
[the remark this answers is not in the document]
nakatsuka-y: Moved 👍

daphne/Cargo.toml
```diff
@@ -35,10 +31,7 @@ serde = { version = "^1.0", features = ["derive"] }
...
 thiserror = "1.0"
 url = { version = "2.2.2", features = ["serde"] }
-clap = { version = "3.2.12", features = ["derive"] } # Required for daph CLI
-reqwest = { version = "0.11.11", features = ["blocking"] } # Required for daph CLI
-anyhow = "1.0.58" # Required for daph CLI
 matchit = "0.6.0"
 
 [dev-dependencies]
-tokio = { version = "^1.19", features = ["macros","rt"] }
+tokio = { version = "^1.19", features = ["rt", "macros"] }
```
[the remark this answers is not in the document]
nakatsuka-y: Removed!

daphne/dapf/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+license = "BSD-3-Clause"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[[bin]]
+name = "dapf"
+path = "src/bin/dapf.rs"
+
+[dependencies]
```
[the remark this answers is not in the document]
nakatsuka-y: I've pruned the dependencies.

daphne/src/hpke.rs
```diff
@@ -210,18 +212,16 @@ impl HpkeReceiverConfig {
         }
     }
 
-    #[cfg(test)]
-    #[allow(dead_code)]
-    pub(crate) fn gen(id: u8, kem_id: HpkeKemId) -> Self {
+    pub fn gen(id: u8, kem_id: HpkeKemId) -> Self {
```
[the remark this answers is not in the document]
nakatsuka-y: I've added the documentation.

daphne/src/vdaf/mod.rs
```diff
@@ -211,9 +211,9 @@ impl VdafConfig {
     ///
     /// * `encrypted_input_share` is the encrypted input share.
     #[allow(clippy::too_many_arguments)]
-    pub(crate) fn consume_report_share<D: HpkeDecrypter>(
+    pub(crate) async fn consume_report_share(
         &self,
-        decrypter: &D,
+        decrypter: &impl HpkeDecrypter,
```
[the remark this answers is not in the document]
nakatsuka-y: `cargo build` complains that we don't have an `&` here.

daphne_worker/src/config.rs
```diff
@@ -57,7 +56,7 @@ pub(crate) struct DaphneWorkerConfig<D> {
 
 impl<D> DaphneWorkerConfig<D> {
     /// Fetch DAP parameters from environment variables.
-    pub(crate) fn from_worker_context(ctx: RouteContext<D>) -> Result<Self> {
+    pub(crate) async fn from_worker_context(ctx: RouteContext<D>) -> Result<Self> {
```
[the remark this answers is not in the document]
nakatsuka-y: You are absolutely right. Removed the unnecessary async.

daphne_worker/src/config.rs
```diff
@@ -264,16 +244,23 @@ impl<D> DaphneWorkerConfig<D> {
...
         &self,
         hpke_config_id: u8,
-    ) -> Option<&HpkeReceiverConfig> {
-        for hpke_receiver_config in self.hpke_receiver_config_list.iter() {
-            if hpke_config_id == hpke_receiver_config.config.id {
-                return Some(hpke_receiver_config);
-            }
-        }
-        None
+    ) -> Option<HpkeReceiverConfig> {
```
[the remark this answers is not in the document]
nakatsuka-y: I've changed this and others to propagate errors correctly.

daphne_worker/src/config.rs
```diff
@@ -264,16 +244,23 @@ impl<D> DaphneWorkerConfig<D> {
...
+        let name = format!("hpke_config/{}", hpke_config_id);
+        let kv_store = self
+            .kv("DAP_HPKE_CONFIG_STORE")
+            .expect("hpke config not found");
```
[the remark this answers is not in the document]
nakatsuka-y: Fixed!

daphne_worker/src/config.rs
```diff
@@ -264,16 +244,23 @@ impl<D> DaphneWorkerConfig<D> {
...
+        let builder = kv_store.get(&name);
+        builder.json().await.expect("kv store get failed")
```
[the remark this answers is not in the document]
nakatsuka-y: Ditto :)

daphne_worker/src/dap.rs
```diff
@@ -64,28 +65,56 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
     Ok(worker_resp)
 }
 
+#[async_trait(?Send)]
 impl<D> HpkeDecrypter for DaphneWorkerConfig<D> {
-    fn get_hpke_config_for(&self, _task_id: &Id) -> Option<&HpkeConfig> {
-        if self.hpke_receiver_config_list.is_empty() {
-            return None;
-        }
+    async fn get_hpke_config_for(&self, _task_id: &Id) -> Option<HpkeConfig> {
```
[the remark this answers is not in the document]
nakatsuka-y: Changed to propagate error.

daphne_worker/src/dap.rs
```diff
@@ -64,28 +65,56 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
...
+        let keys = kv_store.list().execute().await.expect("keys must exist");
+
+        if keys.keys.is_empty() {
+            // Generate a new random HPKE config ID.
+            let hpke_config_id: u8 = rand::thread_rng().gen();
+            let name = format!("hpke_config/{}", hpke_config_id);
+
+            // Generate a new HPKE receiver config.
+            let hpke_receiver_config =
+                HpkeReceiverConfig::gen(hpke_config_id, HpkeKemId::X25519HkdfSha256);
```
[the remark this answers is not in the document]
nakatsuka-y: The HPKE KEM has been added to the global config.

daphne_worker_test/wrangler.toml
nakatsuka-y: You're right. I've changed this and the KV name `hpke_config` to `HPKE_RECEIVER_CONFIG` and `hpke_receiver_config`

nakatsuka-y:
Clippy is complaining about a part of the code that holds onto a Mutex while calling an async function.
I'm currently working on that, but the remaining should be ready for review.

daphne/dapf/src/bin/dapf.rs
```diff
@@ -22,25 +22,25 @@ use url::Url;
 /// DAP Functions, a utility for interacting with DAP deployments.
 #[derive(Parser, Debug)]
 #[clap(version, about, long_about = None)]
-struct Cli {
+pub struct Cli {
```
cjpatton, line 25: This no longer needs to be public, right? (Here and below.)

daphne/dapf/src/bin/dapf.rs
```diff
@@ -74,7 +74,39 @@ enum Action {
     },
 }
 
-fn main() -> Result<()> {
+#[derive(Debug, Deserialize)]
```
cjpatton, line 77: lines 77 to 107 appear to have been moved without actually chnaging the code. Can you move this stuff back to where it was?

daphne/src/lib.rs
```diff
@@ -291,6 +292,9 @@ pub struct DapGlobalConfig {
     /// Batch intervals cannot end more than `max_batch_interval_end`
     /// apart from the current batch interval.
     pub max_batch_interval_end: u64,
+
+    /// The default HPKE KEM used to generate HPKE receiver configs.
+    pub hpke_kem_id: HpkeKemId,
```
cjpatton, line 297: Nice! Can we make this type `supported_hpke_kems: Vec<HpkeKemId>` so that, in the future, we can support multiple KEMs simultaneously if needed? Feel free to treat `supported_hpke_kems.len() != 1` as an internal error.

daphne_worker/src/config.rs
```diff
@@ -264,16 +253,21 @@ impl<D> DaphneWorkerConfig<D> {
         DurableConnector::new(&self.ctx.as_ref().expect("no route context configured").env)
     }
 
-    pub(crate) fn get_hpke_receiver_config_for(
+    pub(crate) fn kv(&self, binding: &str) -> Result<KvStore> {
+        self.ctx
+            .as_ref()
+            .ok_or_else(|| Error::RustError(format!("Unrecognized binding: {}", binding)))?
```
cjpatton, line 259: Is "unrecognized binding" the right feedback here? What might have happened if we got an error here?

daphne_worker/src/dap.rs
```diff
@@ -64,32 +65,117 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
     Ok(worker_resp)
 }
 
+#[async_trait(?Send)]
 impl<D> HpkeDecrypter for DaphneWorkerConfig<D> {
-    fn get_hpke_config_for(&self, _task_id: &Id) -> Option<&HpkeConfig> {
-        if self.hpke_receiver_config_list.is_empty() {
-            return None;
-        }
+    async fn get_hpke_config_for(
```
cjpatton, line 70: Please add a test to `e2e.sh` that confirms the receiver config is written to KV. It's sufficient to confirm that each call to `GET /hpke_config` returns the same config.

daphne_worker/src/dap.rs
```diff
@@ -64,32 +65,117 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
...
-        self.get_hpke_receiver_config_for(config_id).is_some()
+    // This is called by the Leader when handling the `/upload` request.
+    // Get HPKE receiver config from KV.
+    async fn can_hpke_decrypt(
+        &self,
+        _task_id: &Id,
+        config_id: u8,
+    ) -> std::result::Result<bool, DapError> {
+        Ok(self
+            .get_hpke_receiver_config_for(config_id)
```
cjpatton, line 134: Let's check that cache here before hitting KV. just like we do for hpke_decrypt(). Hint: Consider having get_hpke_receiver_config_for() check the cache and populate on a miss. That way you don't have to duplicate this code in multiple places.

cjpatton (changes requested):

daphne/dapf/src/bin/dapf.rs
nakatsuka-y: Yes, you are right.

daphne/dapf/src/bin/dapf.rs
nakatsuka-y: Oops, I totally forgot about this. The file structure should be the way it used to be now.

daphne/src/lib.rs
nakatsuka-y: Great suggestion, this is implemented now.

daphne_worker/src/config.rs
nakatsuka-y: Updated with the correct error message.

daphne_worker/src/dap.rs
nakatsuka-y: The requested test has been added :)

daphne_worker/src/dap.rs
nakatsuka-y: Great suggestion! This actually helped me to fix what Clippy was complaining about.

daphne_worker/src/config.rs
```diff
@@ -264,16 +254,52 @@ impl<D> DaphneWorkerConfig<D> {
...
+    }
+
+    pub(crate) async fn get_hpke_receiver_config_for(
         &self,
         hpke_config_id: u8,
-    ) -> Option<&HpkeReceiverConfig> {
-        for hpke_receiver_config in self.hpke_receiver_config_list.iter() {
-            if hpke_config_id == hpke_receiver_config.config.id {
-                return Some(hpke_receiver_config);
+    ) -> Result<Option<HpkeReceiverConfig>> {
```
cjpatton, line 267: Let's move the code for generating the receiver config (if it doesn't exist) here. Then the return type for this function can be `Result<HpkeReceiverConfig>`.

daphne_worker/src/dap.rs
```diff
@@ -64,29 +65,90 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
...
-    fn get_hpke_config_for(&self, _task_id: &Id) -> Option<&HpkeConfig> {
-        if self.hpke_receiver_config_list.is_empty() {
-            return None;
-        }
+    async fn get_hpke_config_for(
+        &self,
+        _task_id: &Id,
+    ) -> std::result::Result<Option<HpkeConfig>, DapError> {
+        let kv_store = self.kv("DAP_HPKE_RECEIVER_CONFIG_STORE").map_err(dap_err)?;
+        let keys = kv_store
```
cjpatton, line 75: Since we're only ever going to read the first key, it's a good idea to set a limit to the number of keys returned: https://docs.rs/worker-kv/latest/worker_kv/struct.ListOptionsBuilder.html#

While at it, it's probably a good idea to specify a prefix for the query (i.e., "hpke_receiver_config/") so that, if we ever have keys with a different key/value structure in the future ("count" or something), we avoid any bugs that would result.

daphne_worker/src/dap.rs
```diff
@@ -64,29 +65,90 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
...
+            .execute()
+            .await
+            .map_err(|e| DapError::Fatal(e.to_string()))?;
 
-        // Always advertise the first HPKE config in the list.
-        Some(&(self.hpke_receiver_config_list[0]).config)
+        if keys.keys.is_empty() {
+            // Generate a new random HPKE config ID.
+            let hpke_config_id: u8 = rand::thread_rng().gen();
+            let name = format!("hpke_receiver_config/{}", hpke_config_id);
```
cjpatton, line 84: Because "hpke_reciever_config" is used in multiple places, it's a good idea to declare it as a constant. At the top of this file:
```rust
const KV_KEY_PREFIX_HPKE_RECEIVER_CONFIG: &'str = "hpke_receiver_config";
```

Then here we would have
```rust
let name = format!("{}/{}", KV_KEY_PREFIX_HPKE_RECEIVER_CONFIG, hpke_config_id);
```

daphne_worker/src/dap.rs
```diff
@@ -64,29 +65,90 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
...
+
+            // Generate a new HPKE receiver config.
+            //
+            // NOTE(nakatsuka-y): The following logic should be changed when we support
```
cjpatton, line 88: Changed in what way?

daphne_worker/src/dap.rs
```diff
@@ -64,29 +65,90 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
...
+
+            // Store newly generated HPKE receiver config into KV.
+            kv_store
+                .put(&name, hpke_receiver_config.clone())
+                .map_err(|e| DapError::Fatal(e.to_string()))?
+                .execute()
+                .await
+                .map_err(|e| DapError::Fatal(e.to_string()))?;
+
+            // Add HPKE receiver config in list.
```
cjpatton, line 106: This comment doesn't seem relevant.
```suggestion
```

daphne_worker/src/dap.rs
```diff
@@ -64,29 +65,90 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
...
+            } else {
+                Ok(None)
+            }
+        }
     }
 
-    fn can_hpke_decrypt(&self, _task_id: &Id, config_id: u8) -> bool {
-        self.get_hpke_receiver_config_for(config_id).is_some()
+    // This is called by the Leader when handling the `/upload` request.
+    // Get HPKE receiver config from KV.
```
cjpatton, line 124: Function block comments have the format
```rust
///
/// 
```

*But*, this comment isn't super helpful, since this a trait method and the functionality of the method is documented by the trait. If you think additional context would improve readability, consider adding a comment in the function itself.
```suggestion
```

daphne_worker/src/dap.rs
```diff
@@ -64,29 +65,90 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
...
+            .cached_hpke_receiver_configs(config_id)
+            .await
+            .map_err(dap_err)?;
+        Ok(hpke_receiver_configs.get(&config_id).is_some())
     }
 
-    fn hpke_decrypt(
+    // This is called by the Leader and Helper during the aggregation phase and by the Collector
+    // during the collection phase. Get HPKE receiver config from KV if cache in DaphneWorkerConfig
+    // does not exist and get it from cache from next time.
```
cjpatton, line 139: Same comment as above
```suggestion
```

cjpatton (changes requested):

daphne_worker/src/config.rs
nakatsuka-y: It turned out that this does not work because `can_hpke_decrypt` will always return `Ok(true)` which is not correct when a client sends a report share encrypted using an expired HPKE public key.

daphne_worker/src/dap.rs
nakatsuka-y: Great idea, I've missed these functions.

daphne_worker/src/dap.rs
nakatsuka-y: Removed.

daphne_worker/src/dap.rs
nakatsuka-y: Ditto :)

daphne_worker/src/dap.rs
nakatsuka-y: Removed this one as well.

daphne_worker/src/dap.rs
```diff
@@ -64,29 +65,88 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
...
+    ) -> std::result::Result<Option<HpkeConfig>, DapError> {
+        let kv_store = self.kv("DAP_HPKE_RECEIVER_CONFIG_STORE").map_err(dap_err)?;
+        let num_kv_keys: u64 = 1; // Used to limit the number of keys we retreive from KV.
+        let keys = kv_store
+            .list()
+            .limit(num_kv_keys)
+            .prefix(KV_KEY_PREFIX_HPKE_RECEIVER_CONFIG.to_string())
+            .execute()
+            .await
+            .map_err(|e| DapError::Fatal(e.to_string()))?;
```
cjpatton, line 82:
```suggestion
            .map_err(dap_err)?;
```

daphne_worker/src/dap.rs
```diff
@@ -64,29 +65,88 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
...
 
-        // Always advertise the first HPKE config in the list.
-        Some(&(self.hpke_receiver_config_list[0]).config)
+        if keys.keys.is_empty() {
+            // Generate a new random HPKE config ID.
+            let hpke_config_id: u8 = rand::thread_rng().gen();
+            let name = format!("{}/{}", KV_KEY_PREFIX_HPKE_RECEIVER_CONFIG, hpke_config_id);
```
cjpatton, line 87:
```suggestion
            let new_key = format!("{}/{}", KV_KEY_PREFIX_HPKE_RECEIVER_CONFIG, hpke_config_id);
```

daphne_worker/src/dap.rs
```diff
@@ -64,29 +65,88 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
...
+                .map_err(|e| DapError::Fatal(e.to_string()))?;
+
+            Ok(Some(hpke_receiver_config.config))
+        } else {
+            // Return the first HPKE receiver config.
+            let builder = kv_store.get(&keys.keys[0].name);
+            if let Some(hpke_receiver_config) = builder
+                .json::<HpkeReceiverConfig>()
+                .await
+                .map_err(|e| DapError::Fatal(e.to_string()))?
```
cjpatton, line 117:
```suggestion
                .map_err(dap_err)?
```

daphne_worker_test/tests/e2e.rs
```diff
@@ -32,6 +32,22 @@ async fn e2e_helper_hpke_config() {
     t.helper_get_raw_hpke_config(&client).await;
 }
 
+#[tokio::test]
+#[cfg_attr(not(feature = "test_e2e"), ignore)]
+async fn e2e_check_hpke_configs_are_same() {
```
cjpatton, line 37:
```suggestion
async fn e2e_hpke_configs_are_cached() {
```

daphne_worker/src/dap.rs
nakatsuka-y: This doesn't work, because `dap_err` converts `worker::Error` into `daphne::DapError`.

daphne_worker/src/dap.rs
nakatsuka-y: Please see my comment above.

daphne_worker/src/dap.rs
```diff
@@ -64,29 +65,88 @@ pub(crate) fn dap_response_to_worker(resp: DapResponse) -> Result<Response> {
...
-        }
+    async fn get_hpke_config_for(
+        &self,
+        _task_id: &Id,
+    ) -> std::result::Result<Option<HpkeConfig>, DapError> {
+        let kv_store = self.kv("DAP_HPKE_RECEIVER_CONFIG_STORE").map_err(dap_err)?;
+        let num_kv_keys: u64 = 1; // Used to limit the number of keys we retreive from KV.
+        let keys = kv_store
+            .list()
+            .limit(num_kv_keys)
```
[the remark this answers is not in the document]
nakatsuka-y: Fixed.

daphne_worker/src/dap.rs
nakatsuka-y: Changed the variable name.

daphne_worker_test/tests/e2e.rs
nakatsuka-y: Changed the function name.

daphne_worker/src/dap.rs
cjpatton: Isn't that exactly what we want? What error type does `execute()` return? I would have thought it would be `worker::Error`.

nakatsuka-y: Oh sorry, I forgot to mention, but `execute()` returns `KvError`, not `worker::Error`.

nakatsuka-y: The latest push should address this.

daphne_worker/src/dap.rs
nakatsuka-y: See above.

daphne_worker/src/dap.rs
cjpatton: good catch

cjpatton (approved):
🦆 -- good work

Issue #82, opened 2022-08-09
Task provisioning


cjpatton:
* Each deployment has a bearer-token authenticated endpoint for configuring tasks:
  * `POST /add_task`: Stores task ID, task config, and bearer token to accept for the task.
     * Validate task config
     * (Optional) Encrypt secrets
  * `POST /delete_task?task_id=<task_id>`: Deletes a task.
 * Any changes should be a proper subset of David's interop test plan.

divergentdave:
The interop test interface defines a JSON object for setting up task parameters in [table 6](https://divergentdave.github.io/draft-dcook-ppm-dap-interop-test-design/draft-dcook-ppm-dap-interop-test-design.html#table-6). If you dropped the "verifyKey" attribute, that would work for these purposes, I think.

The VDAF and its parameters are covered in [table 1](https://divergentdave.github.io/draft-dcook-ppm-dap-interop-test-design/draft-dcook-ppm-dap-interop-test-design.html#table-1), though we plan to change that soon so that "bits" is a string, not a number. (The rationale is that some VDAF-related parameters, measurements or aggregate results cannot safely be transmitted as a number, since they could be 64-bit or 128-bit integers. Making all VDAF-related items strings simplifies reasoning about them and simplifies adding new VDAFs.)

cjpatton:
> (The rationale is that some VDAF-related parameters, measurements or aggregate results cannot safely be transmitted as a number, since they could be 64-bit or 128-bit integers. ...

Why is this not safe? Is the a limitation of JSON?

divergentdave:
Yes, JS's Number type is a 64-bit float, thus JSON implementations commonly use that for internal representation. `serde_json` can use any of a u64, i64, or f64. `jq` will round large numbers beyond 17 digits. I checked the JSON RFC, and it says that large JSON numbers "may indicate potential interoperability problems".

cjpatton:
Thanks David for pointing this out. Notably this is an issue for the VDAF test vectors as well: https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/116

PR #116, opened 2022-09-09
Allow rwlock-guarded data to be passed by reference

cjpatton:
Partially addresses issue #82. There are two commits in this PR (review them separately):

1. Allow `HpkeDecrypter::get_hpke_config_for` to handle mutex guard.

Change the return type of `HpkeDecrypter::get_hpke_config_for` to
`HpkeDecryptor::WrappedHpkeConfig: AsRef<HpkeConfig>`. This allows
Daphne-Worker to return a reference to an HPKE config protected by a
mutex.

For `DaphneWorkerConfig`, the return type is
`GuardedHpkeReceiverConfig<'a>`, which is required to live as long as
the mutex guard for the underlying data structure (a map from config IDs
to configs). The compiler does not know whether this object lives as
long as `DaphneWorkerConfig`, so it is necessary to add a lifetime to
the `HpkeDecrypter` trait.

Because `get_hpke_config_for()` is called at most once per HTTP request,
and because the HPKE config needs to be fetched from KV for each
request, this change does not save any overhead. (It doesn't add any
either.) However, since plan to move the DAP task configs to KV as part
of issue#82, this pattern will need to be replicated for
`DapAggregator::get_task_config_for()` as well.

2. Make `DapAggregator::get_task_config_for()` async.

Allows us to use KV to back task storage from behind a mutex, which will be
needed for issue#84. Copying each time we need to fetch a task config would
be very expensive.

### DIFF
--- a/daphne/src/lib.rs
+++ b/daphne/src/lib.rs
@@ -325,6 +325,12 @@ impl DapTaskConfig {
     }
 }
 
+impl AsRef<DapTaskConfig> for DapTaskConfig {
+    fn as_ref(&self) -> &Self {
+        self
+    }
+}
+
 #[derive(Deserialize, Serialize)]
 struct ShadowDapTaskConfig {
     version: DapVersion,
--- a/daphne_worker/src/config.rs
+++ b/daphne_worker/src/config.rs
@@ -17,7 +17,7 @@ use daphne::{
     auth::BearerToken,
     constants,
     hpke::HpkeReceiverConfig,
-    messages::{Id, Interval, Nonce},
+    messages::{HpkeConfig, Id, Interval, Nonce},
     DapError, DapGlobalConfig, DapRequest, DapTaskConfig, DapVersion,
 };
 use matchit::Router;
@@ -264,10 +264,10 @@ impl<D> DaphneWorkerConfig<D> {
 
     // Get a reference to the HPKE receiver configs, ensuring that the config indicated by
     // `hpke_config_id` is cached (if it exists).
-    pub(crate) async fn cached_hpke_receiver_configs(
+    pub(crate) async fn hpke_receiver_config(
         &self,
         hpke_config_id: u8,
-    ) -> Result<RwLockReadGuard<HashMap<u8, HpkeReceiverConfig>>> {
+    ) -> Result<Option<GuardedHpkeReceiverConfig>> {
         // If the config is cached, then return immediately.
         {
             let hpke_receiver_configs = self.hpke_receiver_configs.read().map_err(|e| {
@@ -278,7 +278,10 @@ impl<D> DaphneWorkerConfig<D> {
             })?;
 
             if hpke_receiver_configs.get(&hpke_config_id).is_some() {
-                return Ok(hpke_receiver_configs);
+                return Ok(Some(GuardedHpkeReceiverConfig {
+                    hpke_receiver_configs,
+                    hpke_config_id,
+                }));
             }
         }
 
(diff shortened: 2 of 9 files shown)


daphne/src/lib.rs
```diff
@@ -325,6 +325,12 @@ impl DapTaskConfig {
     }
 }
 
+impl AsRef<DapTaskConfig> for DapTaskConfig {
+    fn as_ref(&self) -> &Self {
+        self
+    }
+}
```
cjpatton, line 332: I believe this can be removed (similarly below in messages.rs).

tanyav2:
Given that these are configs with what appears to be significantly more reads than writes, any reason why you're using mutex and not a rwlock?

cjpatton:
> Given that these are configs with what appears to be significantly more reads than writes, any reason why you're using mutex and not a rwlock?

Ohhh, actually we are use `RwLock`, I just have been conflating these 😬

tanyav2:
> > Given that these are configs with what appears to be significantly more reads than writes, any reason why you're using mutex and not a rwlock?
> 
> Ohhh, actually we are use `RwLock`, I just have been conflating these grimacing

oops I'd just started looking at this, based it on the PR message

armfazh (approved):

daphne_worker/src/config.rs
```diff
@@ -264,10 +264,10 @@ impl<D> DaphneWorkerConfig<D> {
 
     // Get a reference to the HPKE receiver configs, ensuring that the config indicated by
     // `hpke_config_id` is cached (if it exists).
-    pub(crate) async fn cached_hpke_receiver_configs(
-        &self,
+    pub(crate) async fn hpke_receiver_config<'a>(
+        &'a self,
```
sudheesh001, line 268: I think the explicit lifetimes here could be elided turning the signature of this method into the following instead with no changes to functionality
```rust
pub(crate) async fn hpke_receiver_config(
        &self,
        hpke_config_id: u8,
    ) -> Result<Option<GuardedHpkeReceiverConfig>> {
```

cjpatton: Holy smokes, you're right! I was under the impression that, since it's part of the type, we'd have to name the parameter here. Maybe we don't have to name it in the trait? Let me give it a shot...

sudheesh001 (approved):

daphne_worker/src/config.rs
cjpatton: I've applied this change and played with ways we might remove the explicit lifetime from the traits. Still not clear if there's a way to do this.

cjpatton:
@armfazh asks (internally): 

> what I didn't understand so well is why to store references for that specific use case.

The high-level goal here is to plumb a reference to an HPKE config (or DAP task config) through the trait API when the underlying data structure is protected by an `RwLock`. I believe an explicit lifetime is needed to tell the compiler that the thing that's returned (i.e., an implementation of `AsRef<HpkeConfig>`) lives as long as the underlying data.

cjpatton:
[Here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c7c9e8b3438d0156ec4d2e972d5beb0c) is a minimal working example if anyone wants to play around with this. This makes me realize that we probably also want to add the lifetime to the `Id` reference as well, I'll update this PR accordingly.

cjpatton:
Since there are no objects, and no one has any better ideas here, I'm gonna merge as-is. Thanks all for the comments.

PR #119, opened 2022-09-14
Moving to hpke-rs

armfazh:
Moving away from `hpke` to `hpke-rs` crate.

### DIFF
--- a/daphne/src/hpke.rs
+++ b/daphne/src/hpke.rs
@@ -3,6 +3,14 @@
 
 //! Hybrid Public-Key Encryption ([HPKE](https://datatracker.ietf.org/doc/rfc9180/)).
 
+use hpke_rs::{Hpke, HpkeError, HpkePrivateKey, HpkePublicKey, Mode};
+use hpke_rs_crypto::{
+    error::Error,
+    types::{AeadAlgorithm, KdfAlgorithm, KemAlgorithm},
+    HpkeCrypto,
+};
+use hpke_rs_rust_crypto::HpkeRustCrypto as ImplHpkeCrypto;
+
 use crate::{
     messages::{
         decode_u16_bytes, encode_u16_bytes, HpkeAeadId, HpkeCiphertext, HpkeConfig, HpkeKdfId,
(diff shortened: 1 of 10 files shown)


bhalleycf (approved):

daphne/src/hpke.rs
```diff
@@ -113,66 +110,18 @@ pub struct HpkeReceiverConfig {
...
-                // We should never get here.
-                unreachable!("Unsupported KEM ({:?})", self.config.kem_id);
-            }
-        }
+        let sender: Hpke<ImplHpkeCrypto> =
+            check_suite(self.config.kem_id, self.config.kdf_id, self.config.aead_id)?;
+        let pk = HpkePublicKey::new(self.config.public_key.clone());
+        let (enc, mut ctx) = sender.setup_sender(&pk, info, None, None, None)?;
+        let ciphertext = ctx.seal(aad, plaintext)?;
+        Ok((enc, ciphertext))
```
cjpatton, line 124: :D
```suggestion
        self.config.encrypt(info, aad, plaintext)
```

daphne/src/hpke.rs
```diff
@@ -184,75 +133,39 @@ impl HpkeReceiverConfig {
...
-                }
-            }
-            _ => {
-                // We should never get here.
-                unreachable!("Unsupported KEM ({:?})", kem_id);
-            }
+        let kem = match kem_id {
+            HpkeKemId::P256HkdfSha256 => KemAlgorithm::DhKemP256,
+            HpkeKemId::X25519HkdfSha256 => KemAlgorithm::DhKem25519,
+            _ => unreachable!("Unsupported KEM ({:?})", kem_id),
```
cjpatton, line 149: Matching "_" for enumerated types is a bit of a footgun in my experience. Better to be explicit!

```suggestion
            NotImplemented(x) => unreachable!("Unsupported KEM ({:?})", x),
```

daphne/src/hpke.rs
```diff
@@ -184,75 +133,39 @@ impl HpkeReceiverConfig {
...
+                    id,
+                    kem_id,
+                    kdf_id: HpkeKdfId::HkdfSha256,
+                    aead_id: HpkeAeadId::Aes128Gcm,
+                    public_key: Vec::from(pk.as_slice()),
+                },
+                secret_key: Vec::from(sk.as_slice()),
+            };
+        } else {
+            unreachable!("bad key generation for KEM ({:?})", kem_id)
```
cjpatton, line 168: Is this actually "unreachable"? Why not propagate the error here?

cjpatton (changes requested):
Nice! Just a few comments.

daphne/src/hpke.rs
armfazh: so true,

daphne/src/hpke.rs
armfazh: done

daphne/src/hpke.rs
armfazh: take a look

daphne/src/lib.rs
```diff
@@ -378,7 +372,10 @@ impl DapGlobalConfig {
         let kem_ids = self.supported_hpke_kems.clone();
         kem_ids.into_iter().enumerate().map(move |(i, kem_id)| {
             let (config_id, _overflowed) = first_config_id.overflowing_add(i as u8);
-            HpkeReceiverConfig::gen(config_id, kem_id)
+            match HpkeReceiverConfig::gen(config_id, kem_id) {
+                Ok(cfg) => cfg,
+                Err(e) => panic!("{:?}", e),
```
armfazh, line 377: @cjpatton any recommendation here to avoid panic?

daphne/src/hpke.rs
```diff
@@ -38,44 +68,26 @@ impl HpkeConfig {
...
+    }
 
-                Ok((enc.to_bytes().to_vec(), sender.seal(plaintext, aad)?))
-            }
-            _ => {
-                // We should never get here.
-                panic!("Unsupported KEM ({:?})", self.kem_id);
-            }
-        }
+    pub fn decrypt(
```
cjpatton, line 78:
```suggestion
    pub(crate) fn decrypt(
```

cjpatton (approved):
Looks good, one minor change.

daphne/src/lib.rs
```diff
@@ -378,7 +372,10 @@ impl DapGlobalConfig {
         let kem_ids = self.supported_hpke_kems.clone();
         kem_ids.into_iter().enumerate().map(move |(i, kem_id)| {
             let (config_id, _overflowed) = first_config_id.overflowing_add(i as u8);
-            HpkeReceiverConfig::gen(config_id, kem_id)
+            match HpkeReceiverConfig::gen(config_id, kem_id) {
```
cjpatton, line 375: No need to match, you could just unwrap() or expect():
```rust
     HpkeReceiverconfig::gen(config_id, kem_id).expect("failed to generate HPKE receiver config")
```

daphne/src/lib.rs
cjpatton: I would probably just propagate the DapError here. I think it's fine to panic, as long as the only way to panic is an RNG failure?

daphne/src/hpke.rs
armfazh: done

daphne/src/lib.rs
armfazh: it is possible that kem_id is invalid, so better to propagate error.

cjpatton (approved):

Issue #100, opened 2022-08-26
Changes for DAP-02


cjpatton:
- [x] agg flow: Increase message lengths from 16 to 32 bits in various places
- [x] Remove Time from Nonce
- [x] Define ReportMetadata
- [x] Define Query and BatchSelector
- [x] Add report count to CollectResp
- [x] Add public share to Report and ReportShare
- [x] Upgrade to VDAF-03
- [x] Update AAD input to HPKE
- [x] "insufficientBatchSize" -> "invalidBatchSize"
- [x] Drop task ID from /hpke_config
- [x] Fixed-size queries.
- [x] Implement max task lifetime https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/304
- [x] Update references to DAP-02
- [x] https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/354
- [x] https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/356
- [x] Add test for leader rejecting expired reports: https://github.com/cloudflare/daphne/pull/137#discussion_r998657793
- [x] WON'T FIX (bumped to #165) Reject reports for epochs too far in the past or future, check for replays in all "active" epochs.
- [x] WON'T FIX (bumped to #177) Client: truncate the timestamp.
- [x] WON'T FIX "The leader MUST remove a collect job's results when the collector sends an HTTP DELETE request to the collect job  URI. The leader responds with HTTP status 204 No Content for requests to a collect job URI whose results have been removed."

Issue #15, opened 2022-06-23
Ideas for serialization rework


cjpatton:
We currently use [libprio](https://github.com/divviup/libprio-rs)'s `codec::Decode` trait for decoding all messages in the DAP protocol. Currently all fields are owned by the struct, i.e., no fields reference data somewhere else. This means that, when decoding a message from the wire, it is necessary to copy bytes into the struct fields. This is fine for short messages, but some messages are quite long. In particular, we would like to be able to decode aggregate requests and responses and have the corresponding structs hold a reference to the underlying data.

cjpatton:
While working on #100 I noticed a couple other nit-picky things with `Decode` and `Encode` that I would like to fix:
1. We have run into message length limits, which suggests that it would be useful if encoding could fail gracefully. The parallel nature of the aggregation flow gives one the impression that an arbitrary number of reports can be processed at once. In fact, this would breach size limits quite quickly in DAP-01.
2. The `Cursor` API is a bit awkward to work with, as you have to typecast usize to u64 in a few places.
3. It might be simpler for decoding to return `Option<_>` instead of `Error<_, CodecError>`. (We pretty much never interpret `CodecError`.)
4. Length-prefixed encoding is a bit awkward, since we have to allocate space for the prefix, then go back and write the prefix. I reallly like how Go solved this problem with the cryptobyte API -- I wonder if we could do the same thing here.

tgeoghegan:
Enabling zero-copy decoding is a good direction to take. I did the simplest thing that would work in the current iteration of the `codec` module, following the unofficial Rust programmer guidance of avoiding references and tricky lifetime problems until you have a real enough performance problem to motivate it. I'm also happy to move away from `Cursor`. Possibly that trait could be generic over `std::io::Read` (which `Cursor<Vec<u8>>` does implement), but I'm guessing I punted on the complex generic signature that would entail for the same reason I chickened out of dealing with references.

> It might be simpler for decoding to return `Option<_>` instead of `Error<_, CodecError>`. (We pretty much never interpret CodecError.)

Applications that don't want the error can use [`std::result::Result::ok`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) to convert to an Option. I don't think `prio` should force all its clients to discard a potentially interesting error string.

branlwyd:
[...]

PR #120, opened 2022-09-14
DAP-02: messages

cjpatton:
Partially addresses #100.

Changes in this PR (each corresponds to a commit, so best to review each commit individually):
1. Bump daphne to 0.2.0 and  lock daphne_worker to daphne 0.1.2 until we're ready to upgrade to 0.2.0. This allows us to move daphne to DAP-02 without impacting daphne_worker.
2. Rework encoding of length-prefixed fields. In preparation for DAP-02, generalize the encoding methods to support     different prefix lengths.
3.  Align various messages with DAP-02: move the timestamp out of the nonce; define the ReportMetadata structure; update length prefixes; and add "batch-saturated" transition failure type. While at it, add type defs for Time and Duration and fiddle with buffers for AAD computations. (*)
4. Define BatchSelector and Query structs.
5. Add public share to Report and ReportShare. (*)
6. add report count to CollectResp. (*)

(*) Indicates a wire-breaking change.


daphne/src/lib.rs
```diff
@@ -398,7 +401,7 @@ pub struct DapLeaderUncommitted {
 /// The Helper's state during the aggregation flow.
 #[derive(Clone, Debug, PartialEq)]
 pub struct DapHelperState {
-    pub(crate) seq: Vec<(VdafState, Nonce)>,
+    pub(crate) seq: Vec<(VdafState, Time, Nonce)>,
```
cjpatton, line 404: This change is due to moving the `time` field out of the `Nonce` struct.

daphne/src/messages.rs
```diff
@@ -4,7 +4,7 @@
 //! Messages in the DAP protocol.
```
cjpatton, line 4: Please double check that all of the following changes match the spec:
- Move time out of Nonce struct
- New ReportMetadata struct and corresponding changes to Report and ReportShare
- New field in AggregateInitialzeReq (we call it the BatchParameter here)
- All length prefixes are bumped properly (16 -> 32)
- New Query and BatchSelector structs (these are identical in fact).
- Changes to AggregateShareReq, CollectReq, and CollectResp.

daphne/src/messages.rs
```diff
@@ -688,15 +825,84 @@ pub(crate) fn constant_time_eq(left: &[u8], right: &[u8]) -> bool {
     r == 0
 }
```
cjpatton, line 827: I'm planning to rework message serialization in the future: see #15.

daphne/src/messages.rs
```diff
@@ -16,6 +16,8 @@ const KEM_ID_X25519_HKDF_SHA256: u16 = 0x0020;
 const KEM_ID_P256_HKDF_SHA256: u16 = 0x0010;
 const KDF_ID_HKDF_SHA256: u16 = 0x0001;
 const AEAD_ID_AES128GCM: u16 = 0x0001;
+const QUERY_TYPE_TIME_INTERVAL: u8 = 0x01;
+const QUERY_TYPE_FIXED_SIZE: u8 = 0x02;
```
chris-wood, line 20: I think these both should be `u16`s:

```
enum {
   reserved(0), /* Reserved for testing purposes */
   time_interval(1),
   fixed_size(2),
   (65535)
} QueryType;
```

daphne/src/messages.rs
```diff
@@ -389,36 +461,91 @@ impl Encode for Interval {
...
+    TimeInterval { batch_interval: Interval },
+    FixedSize { batch_id: Id },
+}
+
+impl Query {
+    pub(crate) fn unwrap_interval(&self) -> &Interval {
+        if let BatchSelector::TimeInterval { ref batch_interval } = self {
+            batch_interval
+        } else {
+            panic!("TOOD(issue #100)");
```
chris-wood, line 483:
```suggestion
            panic!("TODO(issue #100)");
```
Panic with attitude!

daphne/src/messages.rs
```diff
@@ -428,30 +555,40 @@ impl Decode for CollectReq {
 // TODO Add serialization tests.
 #[derive(Clone, Debug, Deserialize, PartialEq, Eq, Serialize)]
 pub struct CollectResp {
+    pub report_count: u64,
```
chris-wood, line 558: Isn't this missing the query-specific response parameters, like the batch ID for chunky queries?

chris-wood (changes requested):
Most of the DAP-02 message changes look good, though I think we missed a couple things.

daphne/src/messages.rs
cjpatton: Oh god, good catch. Why did we make this 2 bytes instead of just one?

daphne/src/messages.rs
cjpatton: Yes, this gets added later in the PR stack with a type called PartialBatchSelector. Should be clearer then.

chris-wood: Is that PR coming? I only see one PR in this stack.

cjpatton:
@chris-wood ready for another round. I updated the size of QueryType to match the spec. I also added a TODO to replace the message encoding stuff with what libprio will ship in 0.10.0.

daphne/src/messages.rs
cjpatton: Yes.

cjpatton: I haven't created PRs yet because there a number of them and I have to  keep them rebased.

chris-wood (approved):

PR #122, opened 2022-09-20
daphne: Upgrade prio to 0.10.0

cjpatton:
Partially addresses #100.

Contains wire-breaking changes.

DAP-02 uses VDAF-03, which is implemented by version 0.10.0 of the `prio` crate. Upgrade the dependency and plumb syntax changes into the protocol accordingly.

While at, it use prio's serialization cdoe isntead of rolling our own.

### DIFF
--- a/daphne/src/vdaf/prio2.rs
+++ b/daphne/src/vdaf/prio2.rs
@@ -23,7 +23,7 @@ pub(crate) fn prio2_shard(
     measurement: DapMeasurement,
 ) -> Result<Vec<Vec<u8>>, VdafError> {
     let vdaf = Prio2::new(dimension as usize)?;
-    let input_shares = match measurement {
+    let (_public_share, input_shares) = match measurement {
         DapMeasurement::U32Vec(ref data) => vdaf.shard(data)?,
         _ => panic!("prio2_shard: unexpected measurement type"),
     };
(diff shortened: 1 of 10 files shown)


daphne/src/vdaf/prio2.rs
```diff
@@ -23,7 +23,7 @@ pub(crate) fn prio2_shard(
     measurement: DapMeasurement,
 ) -> Result<Vec<Vec<u8>>, VdafError> {
     let vdaf = Prio2::new(dimension as usize)?;
-    let input_shares = match measurement {
+    let (_public_share, input_shares) = match measurement {
```
chris-wood, line 26: Why do we not need this?

daphne/src/vdaf/prio3_test.rs
```diff
@@ -76,6 +76,7 @@ fn test_prepare(
     // Unshard
     let agg_res = prio3_unshard(
         config,
+        1,
```
chris-wood, line 79: I'm probably misunderstanding something here -- shouldn't this be `2`?

chris-wood (commented):
Approving assuming I'm just misunderstanding `num_measurements`.

chris-wood (approved):

daphne/src/vdaf/prio2.rs
cjpatton: The public share is meaningful to Poplar1, but not Prio3. However, The VDAF API in libprio is designed to encompass all VDAFs, hence the extra output parameter that we just ignore here.

daphne/src/vdaf/prio3_test.rs
cjpatton: This is the "report count". Only one report is aggregated in this test.

PR #126, opened 2022-09-23
daphne_worker: Upgrade to DAP-02

cjpatton:
Partially addresses #100.
Based on #125 (merge that first).

Upgrade the daphne_worker crate to use the latest daphne crate.

### DIFF
--- a/daphne_worker/src/durable/report_store.rs
+++ b/daphne_worker/src/durable/report_store.rs
@@ -270,15 +270,15 @@ impl DurableObject for ReportStore {
 }
 
 pub(crate) fn nonce_hex_from_report(report_hex: &str) -> Option<&str> {
-    // task_id
-    if report_hex.len() < 64 {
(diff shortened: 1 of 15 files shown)


daphne_worker/src/durable/report_store.rs
```diff
@@ -270,15 +270,15 @@ impl DurableObject for ReportStore {
 }
 
 pub(crate) fn nonce_hex_from_report(report_hex: &str) -> Option<&str> {
```
cjpatton, line 272: Reviewer note: This function borrows the hex encoding of the nonce from a report. The definition of the nonce has changed to not encompass the timestamp, hence this change. Note that this change does not reflect renaming the nonce to "report ID" and moving it in the report structure. This change will come later in the PR stack.

chris-wood (approved):
There's not much changing here since the chunky queries come later, so 🦆 🚢

PR #128, opened 2022-10-03
Make task ID optional for HPKE config request

cjpatton:
Partially addresses #100.
Based on #127 (merge that first).

Per DAP-02, allow the task ID to be an optional query parameter for the /hpke_config endpoint.

### DIFF
--- a/daphne/src/roles.rs
+++ b/daphne/src/roles.rs
@@ -163,18 +163,13 @@ pub trait DapAggregator<'a, S>: HpkeDecrypter<'a> + Sized {
             if task_config.as_ref().version != req.version {
                 return Err(DapAbort::InvalidProtocolVersion);
             }
-
-            let hpke_config = self
-                .get_hpke_config_for(task_id)
-                .await?
-                .ok_or(DapAbort::UnrecognizedTask)?;
-            Ok(DapResponse {
-                media_type: Some(MEDIA_TYPE_HPKE_CONFIG),
-                payload: hpke_config.as_ref().get_encoded(),
-            })
-        } else {
-            Err(DapAbort::BadRequest("missing query parameter".into()))
         }
+
+        let hpke_config = self.get_hpke_config_for(id.as_ref()).await?;
+        Ok(DapResponse {
+            media_type: Some(MEDIA_TYPE_HPKE_CONFIG),
+            payload: hpke_config.as_ref().get_encoded(),
+        })
     }
 }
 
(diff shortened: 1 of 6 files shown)


daphne/src/roles.rs
```diff
@@ -163,18 +163,13 @@ pub trait DapAggregator<'a, S>: HpkeDecrypter<'a> + Sized {
...
-                .ok_or(DapAbort::UnrecognizedTask)?;
-            Ok(DapResponse {
-                media_type: Some(MEDIA_TYPE_HPKE_CONFIG),
-                payload: hpke_config.as_ref().get_encoded(),
-            })
-        } else {
-            Err(DapAbort::BadRequest("missing query parameter".into()))
         }
+
+        let hpke_config = self.get_hpke_config_for(id.as_ref()).await?;
```
cjpatton, line 168: `id` is type `Option<Id>`. It's set above from the query parameters of the URL.

daphne/src/testing.rs
```diff
@@ -198,14 +198,23 @@ impl<'a> HpkeDecrypter<'a> for MockAggregator {
...
+    ) -> Result<&'a HpkeConfig, DapError> {
         if self.hpke_receiver_config_list.is_empty() {
-            return Ok(None);
+            return Err(DapError::fatal("emtpy HPKE receiver config list"));
+        }
+
+        // Aggregators MAY abort if the HPKE config request does not specify a task ID. While not
+        // required for MockAggregator, we simulate this behavior for testing purposes.
+        //
+        // TODO(cjpatton) To make this clearer, have MockAggregator store a map from task IDs to
```
chris-wood, line 210: 👍

daphne_worker/src/dap.rs
```diff
@@ -74,8 +74,8 @@ impl<'a, D> HpkeDecrypter<'a> for DaphneWorkerConfig<D> {
 
     async fn get_hpke_config_for(
         &'a self,
-        _task_id: &Id,
-    ) -> std::result::Result<Option<GuardedHpkeReceiverConfig<'a>>, DapError> {
+        _task_id: Option<&Id>,
```
chris-wood, line 77: It's not obvious (to me) that this is ignored. Can we document that?

chris-wood (approved):

daphne_worker/src/dap.rs
cjpatton: That what is ignored, the task ID? This is denoted by the "_" in front. This is a fairly common pattern for Rust, and you'll find it throughout the code base. IMO this is sufficiently "self-documenting".

PR #129, opened 2022-10-03
daphne: Add query type to task config

cjpatton:
Based on #128 (merge that first).

Add the query type to the task configuration. For fixed-size queries it is necessary to handle batch sizes that exceed the maximum batch size, so do this as well.

While at it, change min_batch_duration to time_precision to match DAP-02.


daphne/src/lib.rs
```diff
@@ -304,18 +304,80 @@ pub struct DapGlobalConfig {
...
+    /// roughly the same size.
+    FixedSize {
+        min_batch_size: u64,
+        max_batch_size: u64,
+    },
+}
+
+impl DapQueryConfig {
+    /// Check if the batch is ready to aggregate based on the report count. Returns an error if the
+    /// batch size is too large.
```
chris-wood, line 327: Isn't the error returned if the _report count_ is too large (`report_count > *max_batch_size`)?

daphne/src/roles.rs
```diff
@@ -424,8 +424,13 @@ pub trait DapLeader<'a, S>: DapAuthorizedSender<S> + DapAggregator<'a, S> {
             .get_agg_share(&collect_req.task_id, &collect_req.query)
             .await?;
 
-        // Check that the minimum batch size is met.
-        if leader_agg_share.report_count < task_config.min_batch_size {
+        // Check the batch size. If not not ready, then return early.
+        //
+        // TODO Consider logging this error, as it shouuld never happen.
```
chris-wood, line 429: 👍

chris-wood (approved):
I was a bit confused at first since I expected the new query types to be supported, but this just moves the query config around. 

Pinning a reminder for myself to ask if we have tests for the `InvalidBatchSIze` error for chunky queries. I assume we'll get to that at some point later?

daphne/src/lib.rs
cjpatton: Yup! Done.

cjpatton:
> Pinning a reminder for myself to ask if we have tests for the InvalidBatchSIze error for chunky queries. I assume we'll get to that at some point later?

+1, I'm not sure we do. This should come up in #133.

PR #130, opened 2022-10-03
dapne: Refactor roles test framework

cjpatton:
Partially addresses #100.
Based on #129 (merge that first).

In preparation for adding support for "fixed_size" queries, refactor the test framework in roles_test.rs to make it easier to define per-test task configurations. The main limitation of the current test code is that the task list is hard-coded. By generating it at runtime we can more easily define a set tasks that we want our code to cover.

This change also addresses another important limitation. All code used to simulate interactions between the various parties is implemented on `MockAggregator`. This makes testing quite awkward, particularly when the parties need to share artifacts from dynamically generated tasks. To address this, a struct `Test` is defined which stores all of the artifacts associated with a logical "deployment".

### DIFF
--- a/daphne/src/lib.rs
+++ b/daphne/src/lib.rs
@@ -29,6 +29,7 @@
 //! on [`VdafConfig`](crate::VdafConfig) for producing reports and consuming aggregate results.
 
 use crate::{
+    hpke::HpkeReceiverConfig,
     messages::{CollectResp, Duration, HpkeConfig, Nonce, Time, TransitionFailure},
     vdaf::{
         prio2::prio2_decode_prepare_state,
@@ -283,7 +284,7 @@ impl AsRef<str> for DapVersion {
 //
 // NOTE: Consider whether the following parameters should be included
 // in the spec.
-#[derive(Deserialize, Serialize)]
+#[derive(Clone, Deserialize, Serialize)]
 pub struct DapGlobalConfig {
     /// Maximum interval duration permitted in CollectReq.
     /// Prevents Collectors from requesting wide range or reports.
@@ -302,6 +303,23 @@ pub struct DapGlobalConfig {
     /// HPKE KEM types that are supported. Used when generating HPKE
     /// receiver config.
     pub supported_hpke_kems: Vec<HpkeKemId>,
+}
+
+impl DapGlobalConfig {
+    /// Generate a list of HPKE receiver configurations, one for each element of supported KEM
+    /// algorithm. `first_config_id` is used as the first config ID; subsequent IDs are chosen by
+    /// incrementing `first_config_id`.
+    pub fn gen_hpke_receiver_config_list(
+        &self,
+        first_config_id: u8,
+    ) -> impl IntoIterator<Item = HpkeReceiverConfig> {
+        assert!(self.supported_hpke_kems.len() <= 256);
+        let kem_ids = self.supported_hpke_kems.clone();
+        kem_ids.into_iter().enumerate().map(move |(i, kem_id)| {
+            let (config_id, _overflowed) = first_config_id.overflowing_add(i as u8);
+            HpkeReceiverConfig::gen(config_id, kem_id)
+        })
+    }
 }
 
 /// DAP Query configuration.
@@ -351,7 +369,7 @@ impl DapQueryConfig {
 }
 
 /// Per-task DAP parameters.
-#[derive(Deserialize, Serialize)]
+#[derive(Clone, Deserialize, Serialize)]
 #[serde(try_from = "ShadowDapTaskConfig")]
 pub struct DapTaskConfig {
     /// The protocol version (i.e., which draft).
--- a/daphne/src/roles_test.rs
+++ b/daphne/src/roles_test.rs
@@ -7,34 +7,32 @@ use crate::{
         MEDIA_TYPE_AGG_CONT_REQ, MEDIA_TYPE_AGG_INIT_REQ, MEDIA_TYPE_AGG_SHARE_REQ,
         MEDIA_TYPE_COLLECT_REQ, MEDIA_TYPE_HPKE_CONFIG, MEDIA_TYPE_REPORT,
     },
-    hpke::HpkeReceiverConfig,
+    hpke::{HpkeDecrypter, HpkeReceiverConfig},
     messages::{
         AggregateContinueReq, AggregateInitializeReq, AggregateResp, AggregateShareReq,
         AggregateShareResp, BatchParameter, BatchSelector, CollectReq, CollectResp, HpkeCiphertext,
-        Id, Interval, Nonce, Query, Report, ReportMetadata, ReportShare, Transition,
-        TransitionFailure, TransitionVar,
+        HpkeKemId, Id, Interval, Query, Report, ReportShare, Time, Transition, TransitionFailure,
+        TransitionVar,
     },
-    roles::{DapAggregator, DapHelper, DapLeader},
-    testing::{
-        AggStoreState, BucketInfo, MockAggregateInfo, MockAggregator, ReportStore,
-        COLLECTOR_BEARER_TOKEN, HPKE_RECEIVER_CONFIG_LIST, LEADER_BEARER_TOKEN,
-    },
-    DapAbort, DapAggregateShare, DapCollectJob, DapLeaderTransition, DapMeasurement,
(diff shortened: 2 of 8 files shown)


daphne/src/lib.rs
```diff
@@ -304,6 +305,23 @@ pub struct DapGlobalConfig {
     pub supported_hpke_kems: Vec<HpkeKemId>,
 }
 
+impl DapGlobalConfig {
+    /// Generate a list of HPKE receiver configurations, one for each element of supported KEM
+    /// algorithm. `first_config_id` is used as the first config ID; subsequent IDs are chosen by
+    /// incrementing `first_config_id`.
```
cjpatton, line 311: Reviewer note: The reason to choose the initial id at random is to prevent peers from ossifying on fixed ids. We could consider choosing random, distinct IDs, but I don't think this is necessary.

daphne/src/lib.rs
```diff
@@ -304,6 +305,23 @@ pub struct DapGlobalConfig {
     pub supported_hpke_kems: Vec<HpkeKemId>,
 }
```
cjpatton, line 307: Reviewer note: This change allows us to reuse some code between daphne and daphne_worker.

daphne_worker/src/dap.rs
```diff
@@ -86,37 +85,48 @@ impl<'a, D> HpkeDecrypter<'a> for DaphneWorkerConfig<D> {
             .map_err(|e| DapError::Fatal(format!("kv_store: {}", e)))?;
 
         let hpke_config_id = if keys.keys.is_empty() {
```
cjpatton, line 87: Reviewer note: The main change here is to use gen_hpke_receiver_config_list to generate the HPKE keys, thereby increasing code reuse. We still only support a single KEM, but it'll now be eassier to add support for multiple KEMs in the future.

daphne/src/testing.rs
```diff
@@ -26,121 +26,27 @@ use std::{
 };
 use url::Url;
```
cjpatton, line 28: Reviewer note: These static JSON blobs are replaced with dynamic configurations generated by `Test::new()`.

daphne/src/roles_test.rs
```diff
@@ -7,34 +7,32 @@ use crate::{
         MEDIA_TYPE_AGG_CONT_REQ, MEDIA_TYPE_AGG_INIT_REQ, MEDIA_TYPE_AGG_SHARE_REQ,
```
cjpatton, line 7: Reviewer note: The bulk of the changes here are due to using `Test` to drive tests instead of `MockAggregator`. The test cases should not have changed.

daphne_worker/src/dap.rs
```diff
@@ -86,37 +85,48 @@ impl<'a, D> HpkeDecrypter<'a> for DaphneWorkerConfig<D> {
...
+            let mut hpke_config_id = None;
+            for hpke_receiver_config in self
+                .global_config
+                .gen_hpke_receiver_config_list(rand::random())
+                .into_iter()
+            {
+                if hpke_config_id.is_none() {
+                    hpke_config_id = Some(hpke_receiver_config.config.id);
+                }
+                let new_key = format!(
```
chris-wood, line 107:
```suggestion
                let new_config_key = format!(
```
nit: since `key` here means "the key for KV" and not "the HPKE key".

daphne/src/roles_test.rs
```diff
@@ -172,160 +277,220 @@ impl MockAggregator {
...
+        let collect_job = self.leader.poll_collect_job(&task_id, &collect_id).await?;
+        assert_matches!(collect_job, DapCollectJob::Done(..));
+
+        Ok(())
+    }
 
-        self.finish_collect_job(task_id, collect_id, &collect_resp)
-            .await
-            .unwrap();
+    async fn leader_authorized_req<M: Encode>(
```
chris-wood, line 428: Nice 👍

daphne/src/roles_test.rs
```diff
@@ -361,92 +526,106 @@ async fn http_get_hpke_config_missing_task_id() {
...
     req.sender_auth = Some(BearerToken::from("incorrect auth token!".to_string()));
     assert_matches!(
-        leader.http_post_collect(&req).await,
+        t.leader.http_post_collect(&req).await,
         Err(DapAbort::UnauthorizedRequest)
     );
 }
 
 #[tokio::test]
 async fn http_post_aggregate_failure_hpke_decrypt_error() {
```
chris-wood, line 608: Unrelated: do we have a test for sending reports with an unknown config ID?

daphne/src/roles_test.rs
```diff
@@ -43,10 +41,131 @@ macro_rules! get_reports {
...
+            max_batch_interval_end: 259200,
+            supported_hpke_kems: vec![HpkeKemId::X25519HkdfSha256],
+        };
+
+        // Task Parameters that the Leader and Helper must agree on.
+        let vdaf_config = VdafConfig::Prio3(Prio3Config::Count);
+        let leader_url = Url::parse("https://leader.biz/v01/").unwrap();
+        let helper_url = Url::parse("http://helper.com:8788/v01/").unwrap();
+        let time_precision = 3600;
+        let version = DapVersion::Draft01;
```
chris-wood, line 79:
```suggestion
        let version = DapVersion::Draft02;
```
Missing update?

daphne/src/roles_test.rs
```diff
@@ -43,10 +41,131 @@ macro_rules! get_reports {
...
+        let collector_hpke_receiver_config =
+            HpkeReceiverConfig::gen(rng.gen(), HpkeKemId::X25519HkdfSha256);
+
+        // Create the task list.
+        let time_interval_task_id = Id(rng.gen());
```
chris-wood, line 84: These are now randomly generated, as opposed to being fixed in a JSON test file, right? Should they be deterministically derived to help with test reproducibility?

chris-wood (changes requested):
The test refactor is really quite nice and removes a lot of redundant code. There's only one potential issue I spotted regarding the draft version wired into the singleton test config.

daphne_worker/src/dap.rs
cjpatton: oh good call. I went with new_kv_config_key to make it even more specific.

daphne/src/roles_test.rs
cjpatton: Thanks!

daphne/src/roles_test.rs
cjpatton: Not yet, this change is still planned.

daphne/src/roles_test.rs
cjpatton: How would this help with reproducibility? The upside of generating it randomly is we avoid writing test code that hardcodes this. (Hardcoding saves us the pain of having to change the value in multiple places when we need to update it. This is already a problem for daphne_worker_test.)

daphne/src/roles_test.rs
cjpatton: Nope, and in fact we weren't enforcing this! I added tests to daphne/src/vdaf/mod_test.rs and added the check to the implementation of HpkeDecrypter for HpkeReceiverConfig.

cjpatton: Good catch.

cjpatton:
> The test refactor is really quite nice and removes a lot of redundant code. There's only one potential issue I spotted regarding the draft version wired into the singleton test config.

The version enum only definfes the draft01 variant for now. We'll replace this with draft02 before #100 is done.

daphne/src/roles_test.rs
chris-wood: Say, for example, a test fails but only for a specific config ID. It'd be a shame if we lost that into the abyss because we never write down the config ID anywhere. (Not a blocking comment -- more of a question.)

chris-wood (approved):

chris-wood:
> The version enum only definfes the draft01 variant for now. We'll replace this with draft02 before https://github.com/cloudflare/daphne/issues/100 is done.

ACK 👍 Flipping to approve.

daphne/src/roles_test.rs
cjpatton: Yeah I get what you're saying. Of course it's possible to fix constants whenever we want a regression test for a behavior that dependent constant. But for the "baseline" test I don't think it's necessary.

PR #131, opened 2022-10-03
daphne: Check for query mismatch

cjpatton:
Partially addresses #100.
Based on #130 (merge that first).

Ensure all requests that encode a query type assert that the correct query type is indicated. Along the way, refactor the batch validation logic so that the Leader and Helper use more of the same code.

In addition, rename "invalidBatchInterval" to "batchInvalid".

### DIFF
--- a/daphne/src/messages.rs
+++ b/daphne/src/messages.rs
@@ -3,7 +3,6 @@
 
 //! Messages in the DAP protocol.
 
-use crate::DapTaskConfig;
 use prio::codec::{
     decode_u16_items, decode_u32_items, encode_u16_items, encode_u32_items, CodecError, Decode,
     Encode,
@@ -442,17 +441,6 @@ impl Interval {
     /// Return the end of the interval, i.e., `self.start + self.duration`.
     pub fn end(&self) -> Time {
         self.start + self.duration
-    }
-
-    /// Check that the batch interval is valid for the given task configuration.
-    pub fn is_valid_for(&self, task_config: &DapTaskConfig) -> bool {
-        if self.start % task_config.time_precision != 0
-            || self.duration % task_config.time_precision != 0
-            || self.duration < task_config.time_precision
-        {
-            return false;
-        }
-        true
     }
 }
 
--- a/daphne/src/roles.rs
+++ b/daphne/src/roles.rs
@@ -12,60 +12,18 @@ use crate::{
     hpke::HpkeDecrypter,
     messages::{
         constant_time_eq, AggregateContinueReq, AggregateInitializeReq, AggregateResp,
-        AggregateShareReq, AggregateShareResp, BatchSelector, CollectReq, CollectResp, Id, Nonce,
-        Report, ReportShare, TransitionFailure, TransitionVar,
+        AggregateShareReq, AggregateShareResp, BatchParameter, BatchSelector, CollectReq,
+        CollectResp, Id, Nonce, Report, ReportShare, Time, TransitionFailure, TransitionVar,
     },
     DapAbort, DapAggregateShare, DapCollectJob, DapError, DapGlobalConfig, DapHelperState,
     DapHelperTransition, DapLeaderProcessTelemetry, DapLeaderTransition, DapOutputShare,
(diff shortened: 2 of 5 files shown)


daphne/src/messages.rs
```diff
@@ -443,17 +442,6 @@ impl Interval {
     pub fn end(&self) -> Time {
         self.start + self.duration
     }
-
```
cjpatton, line 446: Reviewer note: This check is now implemented here: https://github.com/cloudflare/daphne/pull/131/files#diff-6edcf4a9c3adc1ac1e8061cd876a931a9dedf381bdf3ce49d70b34b95a697205R771-R777

daphne/src/roles.rs
```diff
@@ -12,61 +12,19 @@ use crate::{
...
     DapAbort, DapAggregateShare, DapCollectJob, DapError, DapGlobalConfig, DapHelperState,
     DapHelperTransition, DapLeaderProcessTelemetry, DapLeaderTransition, DapOutputShare,
-    DapRequest, DapResponse, DapTaskConfig, DapVersion,
+    DapQueryConfig, DapRequest, DapResponse, DapTaskConfig, DapVersion,
 };
 use async_trait::async_trait;
 use prio::codec::{Decode, Encode};
 use rand::prelude::*;
 use std::collections::HashMap;
 use url::Url;
```
cjpatton, line 27: Reviewer note: Now implemented by https://github.com/cloudflare/daphne/pull/131/files#diff-6edcf4a9c3adc1ac1e8061cd876a931a9dedf381bdf3ce49d70b34b95a697205R752.

daphne/src/roles.rs
```diff
@@ -12,61 +12,19 @@ use crate::{
...
-        }
-
-        // Check that the aggreation parameter is suitable for the given VDAF.
-        if !$task_config.vdaf.is_valid_agg_param(&$agg_param) {
-            // TODO spec: Define this behavior.
-            return Err(DapAbort::UnrecognizedMessage);
-        }
-    }};
-}
-
```
cjpatton, line 46: Reviewer note: This has been moved to https://github.com/cloudflare/daphne/pull/131/files#diff-6edcf4a9c3adc1ac1e8061cd876a931a9dedf381bdf3ce49d70b34b95a697205R779-R793.

daphne/src/roles.rs
```diff
@@ -773,3 +728,78 @@ pub trait DapHelper<'a, S>: DapAggregator<'a, S> {
         })
     }
 }
+
+fn check_part_batch(
```
chris-wood, line 732: It wasn't clear to me what this function is actually doing based on the name, but I don't have a better name suggestion :-\  The previous code used `check_batch_param` to cover the VDAF config check. Was that the intent here?

daphne/src/roles.rs
```diff
@@ -773,3 +728,78 @@ pub trait DapHelper<'a, S>: DapAggregator<'a, S> {
...
+                ));
+            }
+
+            if now.abs_diff(batch_interval.end()) > global_config.max_batch_interval_end {
+                return Err(DapAbort::BadRequest(
+                    "batch interval too far into future".to_string(),
+                ));
+            }
+        }
+        (DapQueryConfig::FixedSize { .. }, BatchSelector::FixedSize { .. }) => (),
```
chris-wood, line 795: TODO!

chris-wood (approved):

daphne/src/roles.rs
cjpatton: This will become slightly less cryptic when we change BatchParameter to PartialBatchSelector up the PR stack. Bascially:

1. check_part_batch: Checks if the aggregation parameter is valid and that the PartialBatchSelector indicated by the Leader AggregateInitializeReq matches the query type for the task.
2. check_batch: Checks if the aggregation parameter is valid and that the BatchSelector indicated by the Leader in its AggregateShareReq (resp. the Collector in its CollectReq) matches the query type for the task.

Thus the terminology will at least be consistent. Still, it can be improved. Not super critical though given that this is an internal API. How about a comment explaining what the thing does, either here or where it's called?

chris-wood (approved):

PR #132, opened 2022-10-03
Prepare `MockAggregator` for fixed-size query tasks

cjpatton:
Partially addresses #100.
Based on #131 (merge that first).

Extend `MockAggregator`'s data stores so that data can be bucketed by batch ID and not just by batch window. Along the way, simplify some of the mutex code.

### DIFF
--- a/daphne/src/lib.rs
+++ b/daphne/src/lib.rs
@@ -404,10 +404,10 @@ pub struct DapTaskConfig {
     pub(crate) collector_hpke_config: HpkeConfig,
 }
 
-#[cfg(test)]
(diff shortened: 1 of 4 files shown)


daphne/src/lib.rs
```diff
@@ -417,6 +417,10 @@ impl DapTaskConfig {
             },
         }
     }
+
```
cjpatton, line 420: Reviewer note: This is increases code reuse a bit.

chris-wood (approved):
TL;DR: Make AggStore take BatchSelectors as input 👍

PR #133, opened 2022-10-03
daphne: Implement fixed-size queries

cjpatton:
Partially addresses #100.
Based on #132 (merge that first).

Complete the implementation of fixed-size queries and add tests for MockAggregator. While at it, add the PartialBatchSelector struct to CollectResp.

PR #134, opened 2022-10-04
daphne_worker: Implement fixed-size queries

### LINKED ISSUE #67: Leader: Check for replays/collected before scheduling report aggregation
Currently our Leader checks if a report pertains to a collected batch when deciding whether to accept an upload request. The following sequence of events would lead to the leader attempting to process reports in an already collected batch, which is illegal:
1. min_batch_size+1 reports are uploaded
2. min_batch_size reports are aggregated
3. Aggregate is collected
4. 1 report is aggregated (for the same batch)

cjpatton:
Based on #133 (merge that first).
Closes #67.

Implement DAP-02's stricter anti-replay requirements (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/362). While at it, add support for fixed-size queries.

The current schema for ReportStore instances encodes the batch window of the reports and report metadata it contains:

    <version>/task/<task_id>/window/<window>/bucket/<shard>

This is no longer sufficient because DAP-02 requires anti-replay enforcement for the report ID, whereas DAP-01 only requires enforcement for the report ID and timestamp.

To implement the new requirements, we instead shard reports across batch boundaries so that anti-replay can be enforced with a constant number of DO queries:

    <version>/task/<task_id>/epoch/<epoch>/shard/<shard>

where <epoch> is something like a day or a week. (We would then reject reports automatically that all in epochs that are too far in the past or the future.)

This creates another problem: Because we have to shard across batch boundaries, we can no longer use ReportStore track if a bucket has been collected. We must instead rely on AggregateStore. Because we cannot shard aggregate shares the way we can shard reports, it is necessary to take this check off the critical path of the upload flow and defer it to the aggregation flow.

Implementing this change resulted in a major refactor of Daphne-Worker. Once done, implementing fixed-size queries semantics was straightforward. All told, this commit consists of the following changes to the daphne_worker crate:

- Split ReportStore into ReportsPending and ReportsStore. Both use the new DO instance naming schema above.
- Refactor and unify queuing logic across DOs.
- [Fixed-size] Add a new LeaderBatchQueue DO, which is responsible for assigning reports to batches for fixed-size tasks.
- [Fixed-size] Extend schema AggregateStore to to support batch IDs.

This commit also makes the following changes to the daphne crate:

- Replace DapHelper::mark_aggregated() with DapAggregator::check_early_reject(), which is used to filter out reports that need to be rejected early. The same functionality is now implemented by both the Leader and Helper. The Leader performs this check during the aggregation flow rather than the upload flow. This allows us to implement this check with latency-sensitive primitives (i.e., Durable Objects).
- Have DapLeader::get_reports() partition reports by PartialBatchSelector, thereby allowing the implementation to assign reports to multiple batch IDs.
- Replace batch_bucket_for_meta with batch_span_for_meta, which allows us to more efficiently coalesce requests to durable storage.
- For consistency with batch_span_for_out_shares and batch_span_for_meta, have batch_span_for_sel return a HashSet.


cjpatton:
Reviewer note: Initially the goal of this PR was to just implement fixed-size queries for Daphne-Worker. Along the way, I ended up wanting to clean up a few things (described in the top comment), the big one being implementing the stronger anti-replay requirements. It would likely make review easier if I split this into a few PRs, but at least one of them (anti-replay) is bound to be fairly large no matter what we do.

bhalleycf (approved):

PR #136, opened 2022-10-11
Complete wire-breaking changes for DAP-02

cjpatton:
Partially addresses #100.
Based on #135 (merge that first).

- Rename and move Nonce -> ReportID
- Reduce QueryType enum from 2 bytes to 1
- Add TaskExpired transition failure type


daphne/src/vdaf/mod.rs
```diff
@@ -579,21 +581,21 @@ impl VdafConfig {
...
+            // If the report ID is not recognized, then respond with a transition failure.
             //
             // TODO spec: Having to enforce this is awkward because, in order to disambiguate the
             // trigger condition from the leader skipping a report that can't be processed, we have
             // to make two passes of the request. (The first step is to compute `recognized`). It
             // would be nice if we didn't have to keep track of the set of processed reports. One
             // way to avoid this would be to require the leader to send the reports in a well-known
-            // order, say, in ascending order by nonce.
-            if !recognized.contains(&leader.nonce) || processed.contains(&leader.nonce) {
+            // order, say, in ascending order by ID.
```
armfazh, line 591: ascending order considering ID interpreted as big- or little-? endian 16-byte number.

armfazh (approved):

daphne/src/vdaf/mod.rs
cjpatton: I don't think it much matters. In any case, we'd have to work out the details in the spec :)

PR #137, opened 2022-10-11
Implement the task expiration mechanism for DAP-02

cjpatton:
Partially addresses #100.
Based on #136 (merge that first).


daphne_worker_test/helper.env
```diff
@@ -15,7 +15,7 @@ DAP_HELPER_STATE_STORE_GARBAGE_COLLECT_AFTER_SECS = '10'
...
+DAP_TASK_LIST = '{"f285be3caf948fcfc36b7d32181c14db95c55f04f55a2db2ee439c5879264e1f":{"version":"v02","leader_url":"http://leader:8787/v02/","helper_url":"http://helper:8788/v02/","collector_hpke_config":{"id":23,"kem_id":"X25519HkdfSha256","kdf_id":"HkdfSha256","aead_id":"Aes128Gcm","public_key":"ec6427a49c8e9245307cc757dbdcf5d287c7a74075141af9fa566c293a52ee7c"},"time_precision":3600,"expiration": 30695925405,"query":{"time_interval":{"min_batch_size":10}},"vdaf":{"prio3":{"sum":{"bits":10}}},"vdaf_verify_key":"1fd8d30dc0e0b7ac81f0050fcab0782d"},"6d2ef497d369fd2e72d96c47e5c28496800f5e546e912554d0df83bf24a25378":{"version":"v02","leader_url":"http://leader:8787/v02/","helper_url":"http://helper:8788/v02/","collector_hpke_config":{"id":23,"kem_id":"X25519HkdfSha256","kdf_id":"HkdfSha256","aead_id":"Aes128Gcm","public_key":"ec6427a49c8e9245307cc757dbdcf5d287c7a74075141af9fa566c293a52ee7c"},"time_precision":3600,"expiration": 30695925405,"query":{"fixed_size":{"min_batch_size":10,"max_batch_size":12}},"vdaf":{"prio3":{"sum":{"bits":10}}},"vdaf_verify_key":"d4c0fd634e80005dc514eb4c9538fbf7"}}'
```
cjpatton, line 18: Reviewer note: Expiration dates for tasks are set to 1000 years from today. This is a hack: Once we implement #138 , /internal/test/add_task, we'll be able to replace these hardcoded configs with configs generated on the fly.

daphne/src/roles.rs
```diff
@@ -321,7 +326,10 @@ pub trait DapLeader<'a, S>: DapAuthorizedSender<S> + DapAggregator<'a, S> {
             .await?;
         let reports = reports
             .into_iter()
-            .filter(|report| early_rejects.get(&report.metadata.id).is_none())
+            .filter(|report| {
+                task_config.expiration >= report.metadata.time
+                    || early_rejects.get(&report.metadata.id).is_none()
```
bhalleycf, line 331: We want the reports that are not expired AND not early rejects, but we have "||" here instead of "&&".  We should add unit test coverage for this filtering too.

cjpatton: Great catch. I checked to make sure the Helper enforces this correctly. In fact, we already have a test to verify the Helper (see http_post_aggregate_init_expired_task() in daphne/src/roles_test.rs). We don't have a corresponding test for the Leader because the test framework for this crate (daphne/src/testing.rs) currently doesn't allow us to run run_agg_job() directly. (Basically we need a way to mock HTTP requests made by the Leader.)  It is feasible to add such a test to daphne_worker_test/tests/e2e.rs. I think we should add this, but I'd like to do so in a separate PR, as there is some e2e.rs refactoring that happens higher in the PR stack.

bhalleycf (approved):
looks good now

daphne/src/roles.rs
cjpatton: Noted in issue https://github.com/cloudflare/daphne/issues/100.

Issue #138, opened 2022-10-11
Implement draft-dcook-ppm-dap-interop-test-design-02


cjpatton:
The draft we're targeting is not quiet ready: See https://github.com/divergentdave/draft-dcook-ppm-dap-interop-test-design.

- [x] [/internal/test/ready](https://divergentdave.github.io/draft-dcook-ppm-dap-interop-test-design/draft-dcook-ppm-dap-interop-test-design.html#name-internal-test-ready-2)
- [x] [/internal/test/endpoint_for_task](https://divergentdave.github.io/draft-dcook-ppm-dap-interop-test-design/draft-dcook-ppm-dap-interop-test-design.html#name-internal-test-endpoint_for_)
- [x] [/internal/test/add_task](https://divergentdave.github.io/draft-dcook-ppm-dap-interop-test-design/draft-dcook-ppm-dap-interop-test-design.html#name-internal-test-add_task)
- [ ] (Leader) [/internal/test/fetch_batch_ids](https://divergentdave.github.io/draft-dcook-ppm-dap-interop-test-design/draft-dcook-ppm-dap-interop-test-design.html#name-internal-test-fetch_batch_i)

fyi/ @divergentdave

PR #141, opened 2022-10-13
daphne_worker: Move bearer tokens to KV

cjpatton:
Partially addresses #138. (We need to be able to generate bearer tokens dynamically in tests.)
Based on #140 (merge that first).

Defines a new endpoint, /internal/test/add_authentication_token, for
storing bearer tokens per-task in KV. This obsoletes the variables
DAP_LEADER_BEARER_TOKEN_LIST and DAP_COLLECTOR_BEARER_TOKEN_LIST.

The map of task IDs to tokens is cached in memory. This table is guarded
by an RwLock. To avoid copying the task ID, we pass a reference to the
already parsed task ID to DapAggregator::authorized() by adding the task
ID field to the DapRequest structure. We still parse the task ID twice
we should consider changing the DAP spec by moving the task IDs out of
message to make it easier to authorize the request before parsing it.


daphne/src/auth.rs
```diff
@@ -12,60 +12,71 @@ use crate::{
     DapError, DapRequest,
 };
 use async_trait::async_trait;
-use prio::codec::Decode;
 use serde::{Deserialize, Serialize};
-use std::io::Cursor;
 
 /// A bearer token used for authorizing DAP requests as specified in draft-ietf-ppm-dap-02.
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct BearerToken(String);
```
cjpatton, line 21: Reviewer note: For whatever reason, the manner in which serde encodes this struct triggers undefined behavior in miniflare. The new format is tolerated.

daphne_worker/src/config.rs
```diff
@@ -366,8 +432,21 @@ impl<D> DaphneWorkerConfig<D> {
             .ok_or_else(|| Error::RustError(format!("Failed to parse path: {}", path)))?;
 
         let payload = req.bytes().await?;
+
+        // Parse the task ID from the front of the request payload and use it to look up the
+        // epxected bearer token.
```
armfazh, line 437:
```suggestion
        // expected bearer token.
```

daphne_worker_test/tests/test_runner.rs
```diff
@@ -347,6 +353,44 @@ impl TestRunner {
...
+                "token": t.leader_bearer_token.clone()
+            }),
+        )
+        .await;
+
+        // Configure the Helper with the Leader's bearer token.
+        t.helper_post_internal(
+            "/internal/test/add_authentication_token",
+            &json!({
+                "role": "leader",
```
armfazh, line 387: it shouldn't be helper? 
-- I understand the semantic here is setting the Leader's token that is used by the Helper. But as the code as it reads looks like something different.

armfazh (approved):

daphne_worker_test/tests/test_runner.rs
cjpatton: Hmm, yeah good point. I don't really have any bright ideas for making this clearer. We could maybe change "role" to "bearer_role"? However this would conflict with one of David's goals for the interop test runner: https://github.com/divergentdave/draft-dcook-ppm-dap-interop-test-design/pull/23/files#r996215019

PR #142, opened 2022-10-14
daphne_worker: Move task list to KV

cjpatton:
Partially addresses #138. (The "add task" endpoint is implemented, but the schema does not match the spec precisely. See https://github.com/divergentdave/draft-dcook-ppm-dap-interop-test-design/issues/17.)
Based on #141 (merge that first).

Defines a new endpoint, /internal/test/add_task, for storing a task config in KV. This obsoletes enviornment variable DAP_TASK_LIST.

In order to better align the JSON schema with the interop draft (draft-dcook-ppm-dap-interop-test-design), move the min_batch_size parameter of DapQueryConfig into DapTaskConfig.

We also take this opportunity to update the serialization of VdafVerifyKey so that it is easier to deserialize. In particular, there is no longer a need for the ShadowDapTaskConfig. The new format is more verbose, but it is only relevant to our implementation.

Finally, change the DapAggregator::get_task_config() API so that we can cache the task list in memory. To make this zoro-copy, we need to add an explicit lifetime to indicate that the value returned lives as long as the request.


daphne/src/roles.rs
```diff
@@ -40,9 +41,12 @@ pub trait DapAuthorizedSender<S> {
...
 #[async_trait(?Send)]
-pub trait DapAggregator<'a, S>: HpkeDecrypter<'a> + Sized {
-    /// Return type of `get_task_config_for()`, wraps a reference to task configuration.
-    type WrappedDapTaskConfig: AsRef<DapTaskConfig>;
+pub trait DapAggregator<'srv, 'req, S>: HpkeDecrypter<'srv> + Sized
+where
+    'srv: 'req,
+{
+    /// A refernce to a task configuration stored by the Aggregator.
+    type WrappedDapTaskConfig: AsRef<DapTaskConfig> + 'req;
```
cjpatton, line 49: Reviewer note: I don't think this bound is strictly necessary.
```suggestion
    type WrappedDapTaskConfig: AsRef<DapTaskConfig>;
```

daphne_worker_test/tests/test_runner.rs
```diff
@@ -634,6 +421,8 @@ impl JanusServerHandle {
     }
 }
 
+// TODO(issue #138) Re-enable Janus interop tests.
+/*
```
armfazh, line 425: I think is cleaner to use a cfg directive.
`#[cfg( feature= "janus" )]`

armfazh (approved):

daphne/src/roles.rs
cjpatton: It does not. cargo seems to be able to infer that the lifetime of the output needs to live at least as long as `'req`. Interestingly, clippy will usually warn you about lifetimes that can be elided, but doesn't here.

daphne_worker_test/tests/test_runner.rs
cjpatton: I agree. But even with the directive, cargo will try to compile this function, but it doesn't currently compile. I could just remove the janus stuff altogether, then add it back when we're ready. That is, I would delete everything that is commented here and delete tests/janus.rs entirely. Would that be preferable you think?

bhalleycf (approved):
I liked Armando's suggestions too.

daphne/src/roles.rs
cjpatton: Removed.

daphne/src/lib.rs
```diff
@@ -593,6 +564,21 @@ impl DapTaskConfig {
 
         Ok(span)
     }
+
+    /// Check if the batch is ready to aggregate based on the report count. Returns an error if the
+    /// report count is too large.
+    pub(crate) fn check_batch_size(&self, report_count: u64) -> Result<bool, DapAbort> {
```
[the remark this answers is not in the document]
cjpatton: Done.

daphne_worker_test/tests/test_runner.rs
cjpatton: Ohhh that worked. Thanks! Done.

Issue #139, opened 2022-10-11
Implement draft-wang-ppm-dap-taskprov-00


cjpatton:
See https://github.com/wangshan/draft-wang-ppm-dap-taskprov.

- [x] Compute task ID (#146)
- [x] daphne::roles: Figure out how to store the pre-shared secret and compute the VDAF verify key (what changes are needed to the traits)
- [x] daphne::roles: Figure out how to store the collector's HPKE config
- [x] daphne::messages: Add Taskprov variant to Extension enum.
- [x] daphne: Add a flag to DapGlobalConfig for enabling taskprov. We should safely ignore the extension if when disabled.
- [x] daphne::vdaf: Use taskprov to compute task config while processing a set reports.
- [x] daphne_worker: Implement taskprov
- [x] Add callback for checking deciding whether to "opt-out"

cc/ @wangshan

bhalleycf:
This is done.

PR #148, opened 2022-10-20
taskprov work

bhalleycf:
Partially addresses #139.

This is a snapshot of work on taskprov.  The TaskConfig structure is not complete yet, I have more fields to add.  Then I'll add a function to compute a task id from the serialized form.

### DIFF
--- a/daphne/src/messages/taskprov.rs
+++ b/daphne/src/messages/taskprov.rs
@@ -1,0 +1,296 @@
+// Copyright (c) 2022 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+//! Messages in the taskprov extension to the DAP protocol, as
+//! defined in draft-wang-ppm-dap-taskprov-00.
+
+use crate::messages::{
+    decode_u16_bytes, encode_u16_bytes, Duration, Id, Time, QUERY_TYPE_FIXED_SIZE,
+    QUERY_TYPE_TIME_INTERVAL,
+};
+use prio::codec::{
+    decode_u16_items, decode_u24_items, decode_u8_items, encode_u16_items, encode_u24_items,
+    encode_u8_items, CodecError, Decode, Encode,
+};
+use ring::{
+    hmac::HMAC_SHA256,
+    hmac::{sign, Key},
+};
+use serde::{Deserialize, Serialize};
+use std::io::Cursor;
+
+// VDAF type codes.
+const VDAF_TYPE_PRIO3_AES128_COUNT: u32 = 0x00000000;
+const VDAF_TYPE_PRIO3_AES128_SUM: u32 = 0x00000001;
+const VDAF_TYPE_PRIO3_AES128_HISTOGRAM: u32 = 0x00000002;
+const VDAF_TYPE_POPLAR1_AES128: u32 = 0x00001000; // The gap from the previous constant is intentional
+
+// Differential privacy mechanism types.
+const DP_MECHANISM_RESERVED: u8 = 0x00;
+const DP_MECHANISM_NONE: u8 = 0x01;
+
+/// A VDAF type along with its type-specific data.
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq, Eq)]
+pub enum VdafTypeVar {
+    Prio3Aes128Count,
+    Prio3Aes128Sum { bit_length: u8 },
+    Prio3Aes128Histogram { buckets: Vec<u64> },
+    Poplar1Aes128 { bit_length: u16 },
+    NotImplemented(u32),
+}
+
+impl Encode for VdafTypeVar {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        match &self {
+            VdafTypeVar::Prio3Aes128Count => VDAF_TYPE_PRIO3_AES128_COUNT.encode(bytes),
+            VdafTypeVar::Prio3Aes128Sum { bit_length } => {
+                VDAF_TYPE_PRIO3_AES128_SUM.encode(bytes);
+                bit_length.encode(bytes);
+            }
+            VdafTypeVar::Prio3Aes128Histogram { buckets } => {
+                VDAF_TYPE_PRIO3_AES128_HISTOGRAM.encode(bytes);
+                encode_u24_items(bytes, &(), buckets);
+            }
+            VdafTypeVar::Poplar1Aes128 { bit_length } => {
+                VDAF_TYPE_POPLAR1_AES128.encode(bytes);
+                bit_length.encode(bytes);
+            }
+            VdafTypeVar::NotImplemented(x) => {
+                x.encode(bytes);
+            }
+        }
+    }
+}
+
+impl Decode for VdafTypeVar {
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
+        let x = u32::decode(bytes)?;
+        match x {
+            VDAF_TYPE_PRIO3_AES128_COUNT => Ok(Self::Prio3Aes128Count),
+            VDAF_TYPE_PRIO3_AES128_SUM => Ok(Self::Prio3Aes128Sum {
+                bit_length: u8::decode(bytes)?,
+            }),
+            VDAF_TYPE_PRIO3_AES128_HISTOGRAM => Ok(Self::Prio3Aes128Histogram {
+                buckets: decode_u24_items(&(), bytes)?,
+            }),
+            VDAF_TYPE_POPLAR1_AES128 => Ok(Self::Poplar1Aes128 {
+                bit_length: u16::decode(bytes)?,
+            }),
+            _ => Err(CodecError::UnexpectedValue),
+        }
+    }
+}
+
+/// A differential privacy mechanism.
+#[derive(Clone, Copy, Deserialize, Serialize, Debug, PartialEq, Eq)]
+pub enum DpMechanism {
+    Reserved,
+    None,
+    NotImplemented(u8),
+}
+
+impl From<DpMechanism> for u8 {
+    fn from(dp_mech: DpMechanism) -> Self {
+        match dp_mech {
+            DpMechanism::Reserved => DP_MECHANISM_RESERVED,
+            DpMechanism::None => DP_MECHANISM_NONE,
+            DpMechanism::NotImplemented(x) => x,
+        }
+    }
+}
+
+impl Encode for DpMechanism {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        u8::from(*self).encode(bytes);
+    }
+}
+
+impl Decode for DpMechanism {
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
+        match u8::decode(bytes)? {
+            x if x == DP_MECHANISM_RESERVED => Ok(Self::Reserved),
+            x if x == DP_MECHANISM_NONE => Ok(Self::None),
+            _ => Err(CodecError::UnexpectedValue),
+        }
+    }
+}
+
+/// A differential privacy configuration.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct DpConfig {
+    pub mechanism: DpMechanism,
+}
+
+impl Encode for DpConfig {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        self.mechanism.encode(bytes);
+    }
+}
+
+impl Decode for DpConfig {
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
+        Ok(Self {
+            mechanism: DpMechanism::decode(bytes)?,
+        })
+    }
+}
+
+/// A VDAF configuration, made up from a differential privacy configuration,
+/// a VDAF type, and type-specific configuration.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct VdafConfig {
+    pub dp_config: DpConfig,
+    pub var: VdafTypeVar,
+}
+
+impl Encode for VdafConfig {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        self.dp_config.encode(bytes);
+        self.var.encode(bytes);
+    }
+}
+
+impl Decode for VdafConfig {
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
+        Ok(Self {
+            dp_config: DpConfig::decode(bytes)?,
+            var: VdafTypeVar::decode(bytes)?,
+        })
+    }
+}
+
+/// A URL encode / decode helper struct, essentially a box for
+/// a Vec<u8>.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct UrlBytes {
+    pub bytes: Vec<u8>,
+}
+
+impl Encode for UrlBytes {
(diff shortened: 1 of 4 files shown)


daphne/src/messages.rs
```diff
@@ -809,6 +817,218 @@ impl Decode for HpkeCiphertext {
     }
 }
```
cjpatton, line 819: Can you add a comment noting that this message is defined in draft-wang-ppm-dap-taskprov-00? (Here and below.)

daphne/src/messages.rs
```diff
@@ -809,6 +817,218 @@ impl Decode for HpkeCiphertext {
...
 
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq, Eq)]
+pub enum VdafType {
+    Prio3AES128Count,
```
cjpatton, line 822: rust naming convention (clippy might complain about this, I'm not sure)
```suggestion
    Prio3Aes128Count,
```

daphne/src/messages.rs
```diff
@@ -809,6 +817,218 @@ impl Decode for HpkeCiphertext {
...
+}
+
+// The VDAF configuration of draft-wang-ppm-dap-taskprov
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct VdafConfig {
+    pub dp_config: DpConfig,
+    pub vdaf_type: VdafType,
+}
+
+impl AsRef<VdafConfig> for VdafConfig {
```
cjpatton, line 937: Is this needed yet? Generally I like to delay implementing traits until we need them. This file has a few cases where we implement AsRef, but in each case we use it elsewhere.

daphne/src/messages.rs
```diff
@@ -4,22 +4,30 @@
...
 };
+use url::Url;
 
 const KEM_ID_X25519_HKDF_SHA256: u16 = 0x0020;
 const KEM_ID_P256_HKDF_SHA256: u16 = 0x0010;
 const KDF_ID_HKDF_SHA256: u16 = 0x0001;
 const AEAD_ID_AES128GCM: u16 = 0x0001;
 const QUERY_TYPE_TIME_INTERVAL: u8 = 0x01;
 const QUERY_TYPE_FIXED_SIZE: u8 = 0x02;
+const PRIO3_AES128_COUNT: u32 = 0x00000000;
```
cjpatton, line 25: Hrm, I think we want to try avoid jamming all of these constants into the same list. This PR might be a good time to get in the habit of grouping constants by type.
```suggestion

// VDAFs (draft-wang-ppm-dap-taskprov-00)
const PRIO3_AES128_COUNT: u32 = 0x00000000;
```

daphne/src/messages.rs
```diff
@@ -4,22 +4,30 @@
...
+const PRIO3_AES128_SUM: u32 = 0x00000001;
+const PRIO3_AES128_HISTOGRAM: u32 = 0x00000002;
+const POPLAR1_AES128: u32 = 0x00001000;
+const DP_MECHANISM_RESERVED: u8 = 0x00;
```
cjpatton, line 29: I don't think we need to define this constant, since the reserved codepoint (0x00) will never be used for a real DpConfig. We could just handle this as DpConfig::NotImplemented, I suppose?

daphne/src/messages.rs
```diff
@@ -809,6 +817,218 @@ impl Decode for HpkeCiphertext {
     }
 }
 
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq, Eq)]
+pub enum VdafType {
+    Prio3AES128Count,
+    Prio3AES128Sum { bit_length: u8 },
+    Prio3AES128Histogram { buckets: Vec<u64> },
+    Poplar1AES128 { bit_length: u16 },
+    NotImplemented(u32),
```
cjpatton, line 826: Typically we would want to store the payload as well, like we do for Extension::Unhandled. However, this makes me realize there is a problem with the spec: How does the decoder know how long the rest of VdafConfig is supposed to be if it gets an unrecognized VdafType? This seems like a bug in the spec. (I'll raise an issue.)

The fix would be to length-prefix the remainder of the VdafConfig message. This would be a breaking change, so it'll have to wait for taskprov-01; I think the only way to handle this situation in taskprov-00 is to hard-fail if we don't recognized the VdafType.

daphne/src/messages.rs
```diff
@@ -809,6 +817,218 @@ impl Decode for HpkeCiphertext {
...
+            bytes: decode_u8_items(&(), bytes)?,
+        })
+    }
+}
+
+// The Task configuration of draft-wang-ppm-dap-taskprov
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct TaskConfig {
+    pub task_info: Vec<u8>,
+    pub aggregator_endpoints: Vec<Url>,
```
cjpatton, line 989: Hmm, so this would mean handling a malformed URL as a CodeError, right? I don't think this is quite what we want. In the actual protocol, the helper would end up aborting with "unrecognizedMessage", which isn't a helpful feedback for the leader.

I would treat this as opaque bytes here and leave it to the application to parse the bytes as a URL.
```suggestion
    pub aggregator_endpoints: Vec<Vec<u8>>,
```

cjpatton (commented):
This is looking pretty good so far! Serialization looks to be correct so far. One comment on readability: Here, "VdafType" actually means `VdafType` in the spec, plus the remainder of the `VdafConfig` struct. This name collision can add to confusion.

One way around this is to follow the `Transition` / `TransitionVar` pattern we have in this file. This would look something like this:

```rust
pub enum VdafConfigVar {
    Prio3Aes128Count,
    ...
}

pub struct VdafConfig {
    pub dp_config: DpConfig,
    pub var: VdafConfigVar,
}
```

daphne/src/messages.rs
bhalleycf: I wasn't sure about that one.  The draft says it is "reserved for testing purposes" but I didn't know if that meant it would be instantiated for testing or not, so I gave it a definition.  Happy to remove it though!

daphne/src/messages.rs
bhalleycf: It's ok at the moment as it is last, but it is a problem if the task config is ever changed in the future.  Though you could argue if you were going to change it, then you should allocate a different extension code anyway.  I'm not objecting to clarifying this part of the spec though!

daphne/src/messages.rs
bhalleycf: ok, much easier for me!

daphne/src/messages.rs
bhalleycf: not needed, will remove until we need it.

daphne/src/messages.rs
bhalleycf: fixed in next push

daphne/src/messages.rs
bhalleycf: fixed in next push

daphne/src/messages.rs
bhalleycf: I grouped these together and also added a VDAF_TYPE_ prefix

daphne/src/messages.rs
bhalleycf: Url simplified to Vec<u8>

daphne/src/messages.rs
cjpatton: Fine to keep it I think :)

daphne/src/messages.rs
cjpatton: > Though you could argue if you were going to change it, then you should allocate a different extension code anyway.

that sounds right, but we're expecting there to be lots of VDAFs in the wild for which there is no formal specification. At least, for the foreseeable future.

daphne/src/messages.rs
```diff
@@ -809,6 +822,266 @@ impl Decode for HpkeCiphertext {
...
+            VDAF_TYPE_PRIO3_AES128_SUM => Ok(Self::Prio3Aes128Sum {
+                bit_length: u8::decode(bytes)?,
+            }),
+            VDAF_TYPE_PRIO3_AES128_HISTOGRAM => Ok(Self::Prio3Aes128Histogram {
+                buckets: decode_u24_items(&(), bytes)?,
+            }),
+            VDAF_TYPE_POPLAR1_AES128 => Ok(Self::Poplar1Aes128 {
+                bit_length: u16::decode(bytes)?,
+            }),
+            _ => Ok(Self::NotImplemented(x)),
```
cjpatton, line 872: Consider the following situation: We're decoding a message like

```
struct {
   ...
   VdafConfig vdaf_config;
   Foo foo;
} Bar;
```

If we don't recognize the VdafType in VdafConfig, then we don't know where vdaf_config ends and foo begins. In my opinion, the only thing we can do here is fail.
```suggestion
            _ => Err(CodecError::UnexpectedValue),
```


I think we should err here instead. (And we should definitely fix this in the spec.)

daphne/src/messages.rs
```diff
@@ -809,6 +822,266 @@ impl Decode for HpkeCiphertext {
...
+        }
+    }
+}
+
+// The Differential Privacy mechanism type defined in draft-wang-ppm-dap-taskprov-00
```
cjpatton, line 877: In rust, docucomments begin with "///" instead of "//". See https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/documentation.html#documenting-source-code
```suggestion
/// The Differential Privacy mechanism type defined in draft-wang-ppm-dap-taskprov-00.
```

daphne/src/messages.rs
```diff
@@ -809,6 +822,266 @@ impl Decode for HpkeCiphertext {
...
+        u8::from(*self).encode(bytes);
+    }
+}
+
+impl Decode for DpMechanism {
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
+        match u8::decode(bytes)? {
+            x if x == DP_MECHANISM_RESERVED => Ok(Self::Reserved),
+            x if x == DP_MECHANISM_NONE => Ok(Self::None),
+            x => Ok(Self::NotImplemented(x)),
```
cjpatton, line 906: Similarly here
```suggestion
            x => Err(CodecError::UnexpectedValue),
```

daphne/src/messages.rs
```diff
@@ -809,6 +822,266 @@ impl Decode for HpkeCiphertext {
...
+}
+
+// URL encode / decode helper struct
+pub(crate) struct UrlBytes {
+    bytes: Vec<u8>,
+}
+
+impl Encode for UrlBytes {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        encode_u16_items(bytes, &(), &self.bytes)
```
cjpatton, line 961: Slightly simpler: encode_u16_bytes()

daphne/src/messages.rs
```diff
@@ -809,6 +822,266 @@ impl Decode for HpkeCiphertext {
...
+    pub time_precision: Duration,
+    pub max_batch_query_count: u16,
+    pub min_batch_size: u32,
+    pub var: QueryConfigVar,
+}
+
+impl Encode for QueryConfig {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        match &self.var {
+            QueryConfigVar::TimeInterval { .. } => {
```
cjpatton, line 996:
```suggestion
            QueryConfigVar::TimeInterval => {
```

daphne/src/messages.rs
```diff
@@ -809,6 +822,266 @@ impl Decode for HpkeCiphertext {
...
+                QUERY_TYPE_FIXED_SIZE.encode(bytes);
+            }
+        }
+        self.time_precision.encode(bytes);
+        self.max_batch_query_count.encode(bytes);
+        self.min_batch_size.encode(bytes);
+        match &self.var {
+            QueryConfigVar::TimeInterval { .. } => {
+                // Nothing to do.
+            }
```
cjpatton, line 1009:
```suggestion
            QueryConfigVar::TimeInterval => (),
```

daphne/src/messages.rs
```diff
@@ -809,6 +822,266 @@ impl Decode for HpkeCiphertext {
...
+pub struct QueryConfig {
+    pub time_precision: Duration,
+    pub max_batch_query_count: u16,
+    pub min_batch_size: u32,
+    pub var: QueryConfigVar,
+}
+
+impl Encode for QueryConfig {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        match &self.var {
```
cjpatton, line 995: Yeah it's super awkward that we have to match on `var` twice. Consider filing an issue against the spec to fix this.

daphne/src/messages.rs
```diff
@@ -809,6 +822,266 @@ impl Decode for HpkeCiphertext {
...
+}
+
+impl Encode for TaskConfig {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        encode_u8_items(bytes, &(), &self.task_info);
+        let url_bytes: Vec<UrlBytes> = self
+            .aggregator_endpoints
+            .iter()
+            .map(|v| UrlBytes { bytes: v.clone() })
+            .collect();
```
cjpatton, line 1056: We wouldn't need to construct this if we used `Vec<UrlBytes>` as the type of `aggreator_endpoints`. This would be slightlier easier to read, and would avoid the clones.

cjpatton (commented):
Encoding looks correct 👍 

Readability comment (feel free to ignore, if you think this would be worse): I would align VdafConfig with QueryConfig by naming the type-dependent part "var". So VdafTypeVar would become VdafConfigVar and the field would be renamed from "vdaf_type" to "var".

daphne/src/messages.rs
bhalleycf: If we do fix the spec, e.g. by making the VdafConfig a length-prefixed thing, then we can come back and change this again.  But I agree that what you suggest is better until then.

daphne/src/messages.rs
cjpatton: This will be addressed in taskprov-01 by https://github.com/wangshan/draft-wang-ppm-dap-taskprov/pull/28.

daphne/src/messages/taskprov.rs
```diff
@@ -0,0 +1,293 @@
...
+use serde::{Deserialize, Serialize};
+use std::io::Cursor;
+
+// VDAF types defined in draft-wang-ppm-dap-taskprov-00
+const VDAF_TYPE_PRIO3_AES128_COUNT: u32 = 0x00000000;
+const VDAF_TYPE_PRIO3_AES128_SUM: u32 = 0x00000001;
+const VDAF_TYPE_PRIO3_AES128_HISTOGRAM: u32 = 0x00000002;
+const VDAF_TYPE_POPLAR1_AES128: u32 = 0x00001000; // The gap from the previous constant is intentional
+
+// Differential Privacy mechanism types defined in draft-wang-ppm-dap-taskprov-00
```
cjpatton, line 28: (here and below) Now that taskprov things are in their own module, and the module refers to the draft, there's no reason to mention the draft in a comment.

daphne/src/messages/taskprov.rs
```diff
@@ -0,0 +1,293 @@
...
+const DP_MECHANISM_RESERVED: u8 = 0x00;
+const DP_MECHANISM_NONE: u8 = 0x01;
+
+/// VdafType plus the variable-length portions associated with it, from draft-wang-ppm-dap-taskprov-00
```
cjpatton, line 32: (here and below) Docucomments should be complete sentences.

cjpatton (approved):
Just a couple of nits regarding comment conventions, otherwise 🦆

PR #152, opened 2022-10-25
taskprov support

bhalleycf:


daphne/src/taskprov.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+/// Compute the VDAF verify key for task_id and the specified VDAF type.
+pub fn compute_vdaf_verify_key(
+    prk: &Prk,
+    task_id: &Id,
+    vdaf_type: VdafType,
+) -> Result<VdafVerifyKey, Unspecified> {
+    let info = [task_id.as_ref()];
+    let okm = prk.expand(&info, vdaf_type.clone())?;
```
cjpatton, line 26: Perhaps I'm reading this wrong, but this doesn't look quite right. [The spec says](https://datatracker.ietf.org/doc/html/draft-wang-ppm-dap-taskprov-00#section-3.2):

```
   verify_key = HKDF-Expand(
       HKDF-Extract(
           task_prov_salt,  # salt
           verify_key_init, # IKM
       ),
       task_id,             # info
       VERIFY_KEY_SIZE,     # L
   )
```

So I would expect to see a call to [prk = extract(task_prov_salt, verify_key_init)](https://docs.rs/ring/0.14.6/ring/hkdf/fn.extract.html) then to expand(prk, task_id, vdafy_type.len()).

daphne/src/taskprov.rs
```diff
@@ -0,0 +1,35 @@
...
+    match &vdaf_type {
+        VdafType::Prio3Aes128Count | VdafType::Prio3Aes128Sum | VdafType::Prio3Aes128Histogram => {
+            let mut bytes = [0u8; 16];
+            okm.fill(&mut bytes[..])?;
```
cjpatton, line 30: Unless the error here depends on the input, I don't think propagating this error is all that useful. Eventually we'd need to handle this as a `DapAbort`.
```
            okm.fill(&mut bytes[..]).expect("HKDF expand failed");
```

daphne/src/taskprov.rs
bhalleycf: The HKDF docs say that making a salt is kind of expensive, and since we use the same extracted value over and over, I split the "extract" part from the expand part.  The extract() is done when making the PRK in compute_vdaf_verify_prk(), and then we use that prk as a parameter in all the calls to compute_vdaf_verify_key().

bhalleycf: Also, the API is funky in that expand() doesn't actually expand, it merely configures expansion and returns an Okm object, on which you then call fill() to actually generate bytes.

cjpatton: OH yes, thanks I missed that bit. I was also thrown by passing vdafy_type to expand. This looks right.

daphne/src/messages/taskprov.rs
```diff
@@ -255,21 +279,25 @@ pub struct TaskConfig {
...
 }
 
+/// SHA-256 of "dap-taskprov-00"
+pub const TASK_PROV_SALT: [u8; 32] = [
+    0x4d, 0x63, 0x1a, 0xeb, 0xa8, 0xdf, 0xe0, 0x1b, 0x34, 0x4c, 0x29, 0x2d, 0x17, 0xba, 0x34, 0x9a,
+    0x78, 0x97, 0xbf, 0x64, 0x88, 0x00, 0x55, 0x1c, 0x0d, 0x75, 0x32, 0xab, 0x61, 0x4b, 0xe2, 0x21,
+];
+
+/// Compute the task id of a serialized task config.
+pub fn compute_task_id(serialized: &[u8]) -> Id {
```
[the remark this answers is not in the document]
bhalleycf: I considered moving it but didn't to avoid circularity but I'll see what else I can do because I too would prefer it elsewhere.

daphne/src/roles.rs
```diff
@@ -233,10 +243,12 @@ where
...
         let report = Report::get_decoded(req.payload.as_ref())?;
         let task_config = self
-            .get_task_config_for(Cow::Borrowed(req.task_id()?))
+            .get_task_config_considering_taskprov(
+                Cow::Borrowed(req.task_id()?),
+                Some(&report.metadata),
+            )
             .await?
             .ok_or(DapAbort::UnrecognizedTask)?;
-
```
cjpatton, line 239: nit: I typically prefer a space before a block comment.

daphne/src/roles.rs
```diff
@@ -497,16 +509,15 @@ where
...
             for (part_batch_sel, reports) in reports.into_iter() {
                 // TODO Consider splitting reports into smaller chunks.
                 // TODO Consider handling tasks in parallel.
                 telem.reports_processed += reports.len() as u64;
                 if !reports.is_empty() {
+                    let task_config = self
+                        .get_task_config_for(Cow::Owned(task_id.clone()))
+                        .await?
+                        .ok_or(DapAbort::UnrecognizedTask)?;
```
cjpatton, line 520: What's the reason for moving this task_config lookup into the inner for-loop?

daphne/src/roles.rs
```diff
@@ -547,7 +558,7 @@ where
         helper_state: &DapHelperState,
     ) -> Result<(), DapError>;
 
-    /// Fetch the Helper's aggregation-flow state. `None` is returned if the Helper has no state
+    /// Fetch the Helper's aggregation-flow state. `None` is returned if the Helper has no statef
```
cjpatton, line 561:
```suggestion
    /// Fetch the Helper's aggregation-flow state. `None` is returned if the Helper has no state
```

daphne/src/roles.rs
```diff
@@ -575,8 +586,15 @@ where
         match req.media_type {
             Some(MEDIA_TYPE_AGG_INIT_REQ) => {
                 let agg_init_req = AggregateInitializeReq::get_decoded(&req.payload)?;
+                let mut first_metadata: Option<&ReportMetadata> = None;
+                if !agg_init_req.report_shares.is_empty() {
+                    first_metadata = Some(&agg_init_req.report_shares[0].metadata);
+                }
```
cjpatton, line 592: [first()](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.first) is more idiomatic (not sure if the syntax is precisely correct)

```suggestion
                let first_metadata = agg_init_req.report_shares.first().map(|report_share| report_share.metadata);
```

daphne/src/taskprov.rs
```diff
@@ -0,0 +1,171 @@
+use crate::{
```
cjpatton, line 1: All files need copyright info. (Feel free to copy-paste from any other Rust source file, they should all be identical at this point.)

daphne/src/taskprov.rs
```diff
@@ -0,0 +1,171 @@
...
+
+/// SHA-256 of "dap-taskprov-00"
+#[allow(dead_code)]
+pub(crate) const TASK_PROV_SALT: [u8; 32] = [
+    0x4d, 0x63, 0x1a, 0xeb, 0xa8, 0xdf, 0xe0, 0x1b, 0x34, 0x4c, 0x29, 0x2d, 0x17, 0xba, 0x34, 0x9a,
+    0x78, 0x97, 0xbf, 0x64, 0x88, 0x00, 0x55, 0x1c, 0x0d, 0x75, 0x32, 0xab, 0x61, 0x4b, 0xe2, 0x21,
+];
+
+/// Compute the task id of a serialized task config.
+pub fn compute_task_id(serialized: &[u8]) -> Id {
```
cjpatton, line 26: Can/should this be `pub(crate)`?

daphne/src/taskprov.rs
```diff
@@ -0,0 +1,171 @@
...
+    let key = Key::new(HMAC_SHA256, &TASK_PROV_SALT);
```
cjpatton, line 27: I would add a comment here indicating that we're re-implementing HKDF-Expand().

daphne/src/taskprov.rs
```diff
@@ -0,0 +1,171 @@
...
+    let digest = sign(&key, serialized);
+    let mut b: [u8; 32] = [0; 32];
+    let d = digest.as_ref();
+    b[..32].copy_from_slice(&d[..32]);
+    Id(b)
+}
+
+/// Compute the VDAF pseudorandom key for the pre-shared secret verify_key_init.
+pub(crate) fn compute_vdaf_verify_prk(verify_key_init: &[u8]) -> Prk {
```
cjpatton, line 36: I would name this in a way that's more indicative of how it's used in the protocol.
```suggestion
/// Extract pseudorandom key from the pre-shared secret used for the "taskprov" extension.
pub(crate) fn extract_prk_from_verify_key_init(verify_key_init: &[u8]) -> Prk {
```

daphne/src/taskprov.rs
```diff
@@ -0,0 +1,171 @@
...
+    // The documentation says computing the Salt is expensive, and we use the same PRK all the
+    // time, so we compute it once.
+    Salt::new(HKDF_SHA256, &TASK_PROV_SALT).extract(verify_key_init)
+}
+
+/// Compute the VDAF verify key for task_id and the specified VDAF type using the
+/// specified pseudorandom key.
+pub(crate) fn compute_vdaf_verify_key_from_prk(
```
cjpatton, line 44:
```suggestion
/// Expand a pseudorandom key into the VDAF verification key for a given task.
pub(crate) fn expand_prk_into_verify_key(
```

daphne/src/messages/taskprov.rs
```diff
@@ -29,6 +26,29 @@ const VDAF_TYPE_POPLAR1_AES128: u32 = 0x00001000; // The gap from the previous c
 const DP_MECHANISM_RESERVED: u8 = 0x00;
 const DP_MECHANISM_NONE: u8 = 0x01;
 
+/// A VDAF type.
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum VdafType {
```
cjpatton, line 32: Can this be `pub(crate)`?

daphne/src/taskprov.rs
```diff
@@ -0,0 +1,171 @@
...
+    let digest = sign(&key, serialized);
+    let mut b: [u8; 32] = [0; 32];
+    let d = digest.as_ref();
+    b[..32].copy_from_slice(&d[..32]);
+    Id(b)
+}
+
+/// Compute the VDAF pseudorandom key for the pre-shared secret verify_key_init.
+pub(crate) fn compute_vdaf_verify_prk(verify_key_init: &[u8]) -> Prk {
+    // The documentation says computing the Salt is expensive, and we use the same PRK all the
```
cjpatton, line 37: Which documentation? In any case, are we actually making use of the precomputation of the PRK? I'm fine to leave as-is, I'm just curious.

daphne/src/taskprov.rs
```diff
@@ -0,0 +1,171 @@
...
+        }
+        _ => panic!("Unknown VDAF type"),
+    }
+}
+
+/// Compute the VDAF verify key for task_id and the specified VDAF type using the
+/// pre-shared secret verify_key_init.
+///
+/// This is a convenience function to call compute_vdaf_verify_prk() and
+/// compute_vdaf_verify_key_from_prk().  Callers reusing the same PRK frequently
```
cjpatton, line 67:
```suggestion
/// compute_vdaf_verify_key_from_prk(). Callers reusing the same PRK frequently
```

daphne/src/taskprov.rs
```diff
@@ -0,0 +1,171 @@
...
+        task_id,
+        vdaf_type,
+    )
+}
+
+pub fn bad_request(detail: &str) -> DapError {
+    DapError::Abort(DapAbort::BadRequest(detail.to_string()))
+}
+
+/// Check for a taskprov extension in the report, and return it if found.
```
cjpatton, line 87: Parsing issues should generally be handled as abort "unrecognizedMessage". In particular, I think it would be better to parse the extension in the implementation of `Decode` for `Extension` (messages/mod.rs)`.

However you have hit on another point here that Tim G. noticed today: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/375. We'll  have to decide how to address this in 03, but for now I think it makes perfect sense to just treat duplicate extensions as decoding failures. In particular, we should lift this de-duplication step to the implementation of `Decode` on `ReportMetadata`.

daphne/src/taskprov.rs
```diff
@@ -0,0 +1,171 @@
...
+        },
+        _ => Err(DapError::Abort(DapAbort::BadRequest(
+            "more than one taskprov extension".to_string(),
+        ))),
+    }
+}
+
+fn url_from_bytes(bytes: &[u8]) -> Result<Url, DapError> {
+    let s = str::from_utf8(bytes).map_err(|_| bad_request("bad URL UTF8"))?;
+    Url::parse(s).map_err(|_| bad_request("bad URL syntax"))
```
cjpatton, line 120: Ah yes! We definitely need to handle this case by aborting, but we want to make sure we properly ["opt-out"](https://datatracker.ietf.org/doc/html/draft-wang-ppm-dap-taskprov-00#section-3.3). The protocol behavior depends somewhat the situation, but should be spelled out in the Leader Behavior and Helper Behavior sections of the taskprov spec.

cjpatton (commented):
I'm still working through the changes to the daphne crate. Everything overall looks good, but I have enough comments that I think it would be best to pause here and wait for changes. 

One high-level comment: In vdaf/mod.rs, the Helper needs to make sure the taskprov extension is used consistently as described in [Section 6](https://datatracker.ietf.org/doc/html/draft-wang-ppm-dap-taskprov-00#section-6).

daphne/src/lib.rs
```diff
@@ -347,6 +347,9 @@ pub struct DapGlobalConfig {
     /// HPKE KEM types that are supported. Used when generating HPKE
     /// receiver config.
     pub supported_hpke_kems: Vec<HpkeKemId>,
+
+    /// Is the taskprov extension allowed?
+    pub taskprov_allowed: bool,
```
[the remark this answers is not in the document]
bhalleycf: good idea, will do

daphne/src/messages/taskprov.rs
bhalleycf: It can.

daphne/src/taskprov.rs
bhalleycf: the documentation for ring::hkdf.  I will improve the comment.  The code is structured so the optimization could be made relatively easily if it turned out to be needful, but I haven't made the optimization.

daphne/src/taskprov.rs
bhalleycf: good idea, will do

daphne/src/taskprov.rs
bhalleycf: Will work on this.

daphne/src/messages/taskprov.rs
```diff
@@ -294,3 +308,10 @@ impl Decode for TaskConfig {
         })
     }
 }
+
+impl TaskConfig {
+    pub fn serialize(&self) -> Vec<u8> {
```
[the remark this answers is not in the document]
bhalleycf: Yes, it seemed like an implementation detail that callers (e.g. daphne_worker_test's e2e tests) shouldn't have to know, so I wrapped it.

daphne/src/roles.rs
bhalleycf: me too!

daphne/src/roles.rs
bhalleycf: Good catch; this was an artifact of a change I undid, but not fully.  Originally I was calling get_task_config_considering_taskprov() here, but it was too late so I moved the call to the right earlier place but forgot to undo this change.  I will fix it.  (The way things are now, the original code works fine as taskprov config instantiation will already have happened and the normal get_task_config_for() will find the task.)

daphne/src/taskprov.rs
bhalleycf: No, because daphne_worker_test needs it.

daphne/src/roles.rs
bhalleycf: it was almost right, needed an & :). fixed, thanks!

daphne/src/messages/mod.rs
```diff
@@ -101,12 +104,17 @@ impl AsRef<[u8]> for ReportId {
 #[derive(Clone, Debug, Deserialize, PartialEq, Eq, Serialize)]
 #[serde(rename_all = "snake_case")]
 pub enum Extension {
+    Taskprov { payload: Vec<u8> }, // Not a TaskConfig to avoid circularity and also for efficiency if taskprov is off.
```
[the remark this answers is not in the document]
bhalleycf: no circularity but it works better the way it is after trying it the other way

daphne_worker/src/config.rs
```diff
@@ -84,6 +84,18 @@ pub(crate) struct DaphneWorkerConfig<D> {
 
     /// Indicates if DaphneWorker is used as the Leader.
     is_leader: bool,
+
```
cjpatton, line 87: Sugestion: Instead of having each of these be individually optional, It might make more sense to group them into one struct and make that struct optional:
```rust
pub(crate) taskprov_config: Option<TaskprovConfig>,
```

That way you don't have to deal with edge cases where the hpke_collector_config is present but the vdaf_verify_key_init is missing.

daphne_worker/src/dap.rs
```diff
@@ -251,11 +269,78 @@ where
...
+                &taskprov_task_id,
+                taskprov_task_config.unwrap(),
+                self.taskprov_vdaf_verify_key_init.as_ref().unwrap(),
+                self.taskprov_hpke_collector_config.as_ref().unwrap(),
+            )?;
+
+            // Define a leader bearer token.  We do this so authorize_with_bearer_token()
+            // finds something.  XXXRTH Note that this is generating KV garbage that will
+            // need collection at some point.
+            self.set_leader_bearer_token(
```
cjpatton, line 324: Why do we need to store the taskprov bearer tokens in KV? Couldn't we just grab it from the DaphneWorkerConfig as needed?

daphne_worker/src/dap.rs
```diff
@@ -251,11 +269,78 @@ where
         &self.global_config
     }
 
-    async fn get_task_config_for(
+    async fn get_task_config_considering_taskprov(
         &'srv self,
         task_id: Cow<'req, Id>,
+        metadata: Option<&ReportMetadata>,
     ) -> std::result::Result<Option<GuardedDapTaskConfig<'req>>, DapError> {
-        self.get_task_config(task_id).await.map_err(dap_err)
```
cjpatton, line 258: I was a bit confused by the control flow because it seems more natural to check KV for an existing task config only if taskprov is disallowed. In fact, what we're doing here is detecting if we have already configured a taskprov task. Maybe add a comment to clarify?

daphne/src/messages/mod.rs
```diff
@@ -143,11 +164,19 @@ impl Encode for ReportMetadata {
...
             id: ReportId::decode(bytes)?,
             time: Time::decode(bytes)?,
             extensions: decode_u16_items(&(), bytes)?,
-        })
+        };
+        // Check for duplicate extensions.
+        let mut seen: HashSet<u16> = HashSet::new();
+        for extension in &metadata.extensions {
+            if !seen.insert(extension.type_code()) {
+                return Err(CodecError::UnexpectedValue);
```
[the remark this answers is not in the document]
bhalleycf: will do

daphne/src/roles.rs
```diff
@@ -575,8 +588,31 @@ where
...
+                    let using_taskprov = agg_init_req
+                        .report_shares
+                        .iter()
+                        .filter(|share| share.metadata.is_taskprov())
+                        .count();
+
+                    if using_taskprov != 0 && using_taskprov != agg_init_req.report_shares.len() {
+                        // It's not all taskprov or no taskprov, so it's an error.
+                        return Err(DapAbort::UnrecognizedMessage);
+                    }
```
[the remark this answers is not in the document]
bhalleycf: My bad

daphne/src/roles.rs
```diff
@@ -575,8 +588,31 @@ where
...
+
+                let first_metadata = agg_init_req
+                    .report_shares
+                    .first()
+                    .map(|report_share| &report_share.metadata);
                 let wrapped_task_config = self
-                    .get_task_config_for(Cow::Borrowed(req.task_id()?))
+                    .get_task_config_considering_taskprov(
+                        Cow::Borrowed(req.task_id()?),
+                        first_metadata,
```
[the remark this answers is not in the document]
bhalleycf: good idea

daphne_worker/src/config.rs
bhalleycf: good idea

cjpatton (commented):
Very nice! Just a few more small things.

daphne_worker/src/dap.rs
```diff
@@ -251,11 +268,78 @@ where
...
+                Some(config) => config,
+                None => return Err(bad_request("taskprov configuration not found")),
+            };
+
+            // We currently always opt-in; any additional checks that would lead to
+            // us opting out should be done here.
+
+            let taskprov_task_id = task_id.as_ref().clone();
+            let task_config = DapTaskConfig::try_from_taskprov(
+                DapVersion::Draft02,
```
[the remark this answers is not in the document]
cjpatton: #161

daphne_worker/src/dap.rs
```diff
@@ -251,11 +268,78 @@ where
...
+        if taskprov_task_config.is_some() {
+            let global_config = self.get_global_config();
+            if !global_config.allow_taskprov {
+                // TODO(bhalleycf) if DAP gets a generic denied error, we should use it here.
+                return Err(bad_request("taskprov is not allowed"));
+            }
+            let taskprov_config = match &self.taskprov_config {
+                Some(config) => config,
+                None => return Err(bad_request("taskprov configuration not found")),
+            };
```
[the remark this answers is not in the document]
bhalleycf: will adopt this with a small change as the as_ref() is in the wrong place above

cjpatton (approved):
🦆

PR #156, opened 2022-10-28
Update Dockerfile, fix tests in docker-compose

divergentdave:
This updates the Dockerfile to accommodate the upgrade to Wrangler v2, and fixes the hostnames used in tests in a few places. I fixed the tests by deleting a few things that were overwriting URL hosts with `127.0.0.1`. As part of this, I deleted the last use of `DAP_DEPLOYMENT` in the aggregator. Now, `DAP_DEPLOYMENT` is only used as a plain environment variable in the host-targeted side of end-to-end tests.

I checked that the main test suite passes and the end-to-end test suite passes when run either locally or in docker-compose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,7 @@ The `daphne` crate relies on unit tests. The `daphne_worker` crate relies mostly
 on integration tests implemented in `daphne_worker_test`. See the README in that
 directory for instructions on running Daphne-Worker locally.
 
-> NOTE Integration tests can be run via docker-compose, but this is not working
-> at the moment.
+Integration tests can be run via docker-compose.
 
 ```
 docker-compose up --build --abort-on-container-exit --exit-code-from test
--- a/daphne_worker/src/config.rs
+++ b/daphne_worker/src/config.rs
@@ -68,6 +68,8 @@ pub(crate) struct DaphneWorkerConfig<D> {
     tasks: Arc<RwLock<HashMap<Id, DapTaskConfig>>>,
 
     /// Deployment type. This controls certain behavior overrides relevant to specific deployments.
+    #[allow(unused)]
+    // No overrides have been defined yet, but this will be used in the future.
     pub(crate) deployment: DaphneWorkerDeployment,
 
     /// Sharding key, used to compute the ReportsPending or ReportsProcessed shard to map a report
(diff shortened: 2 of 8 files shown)


cjpatton:
Yes, I think I know what needs to be fixed. However, stepping back: One feature of our docker-compose is that we use a "virtual" network instead of the host. Do you think this is useful? https://github.com/cloudflare/daphne/pull/156/files#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9R5-R7

daphne_worker_test/docker/miniflare.Dockerfile
```diff
@@ -1,27 +1,30 @@
...
+RUN npm install -g n wasm-pack && \
     n 18.4.0
-RUN npm install -g \
-        @cloudflare/wrangler \
-        wasm-pack \
-        miniflare@2.5.1
+
+WORKDIR /tmp/wrangler2
+RUN git clone https://github.com/cloudflare/wrangler2 -b wrangler@2.1.13 . && \
+    npm install -g wrangler
```
armfazh, line 8: does this is enough here?
```suggestion
RUN npm install -g wrangler@2.1.13
```

divergentdave: Oh wow, that's much simpler. I missed this because the wrangler2 README only linked to the NPM page for v1.

README.md
```diff
@@ -36,8 +36,7 @@ The `daphne` crate relies on unit tests. The `daphne_worker` crate relies mostly
 on integration tests implemented in `daphne_worker_test`. See the README in that
 directory for instructions on running Daphne-Worker locally.
 
-> NOTE Integration tests can be run via docker-compose, but this is not working
-> at the moment.
+> NOTE Integration tests can be run via docker-compose.
```
cjpatton, line 39: The "> NOTE" bit was meant to call out that this is a bug

```suggestion
Integration tests can be run via docker-compose.
```

daphne_worker/src/config.rs
```diff
@@ -538,16 +516,3 @@ impl AsRef<DapTaskConfig> for GuardedDapTaskConfig<'_> {
         self.value()
     }
 }
-
-/// Deployment types for Daphne-Worker. This defines overrides used to control inter-Aggregator
```
cjpatton, line 542: Hmm, I'm definitely OK with removing `Dev` (and simplifying accordingly, as you have done), but removing this enum altogether is a regression, as I plan to have other deployment types, such as one that allows both Aggregators to run on Cloudflare infra.

For the sake of unblocking interop testing, let's leave as-is. I'll just leave a comment here to link to an issue.

cjpatton (commented):
I'm getting "connection refused" when the test runner tries /internal/delete_all. I'll get to debugging.

cjpatton (approved):
Please squash, and I'll merge.

cjpatton (approved):

PR #158, opened 2022-11-01
/internal/test/add_task: Return status in response

divergentdave:
This returns a success status in the response object returned from the `/internal/test/add_task` handler, and fleshes out uses of response objects in the end-to-end tests. (see [Table 8](https://divergentdave.github.io/draft-dcook-ppm-dap-interop-test-design/draft-dcook-ppm-dap-interop-test-design.html#table-8))


cjpatton (approved):
Boy, that was a little sloppy of me. Thanks for sending a PR -- in the future, you should feel free to file an issue to get something in the interop test endpoints fixed.

Issue #162, opened 2022-11-07
Implement draft-wang-ppm-dap-taskprov-01


cjpatton:
The first drat, 00, was implemented in #139. The new draft has a [few minor changes](https://www.ietf.org/rfcdiff?difftype=--hwdiff&url2=draft-wang-ppm-dap-taskprov-01.txt).

bhalleycf:
This is done.

Issue #161, opened 2022-11-04
Clean up: Avoid hard-coding draft numbers in protocol logic.


cjpatton:
This came up in the implementation of taskprov: https://github.com/cloudflare/daphne/pull/152#discussion_r1014508424

bhalleycf:
Hardcoding in protocol logic has been removed as part of #171.

PR #164, opened 2022-11-09
Implement draft-wang-ppm-dap-taskprov-01.

bhalleycf:
This implements the changes in draft-wang-ppm-dap-taskprov-01, [#162] while continuing to support draft-wang-ppm-dap-taskprov-00.  The taskprov extension is not versioned, and we are currently using the same extension id for both versions, so which version we implement is controlled by global server configuration.  We may wish to considering using a different number, e.g. 0xff01, as that would allow the server to support both versions concurrently.

This PR also addresses [#161] as the work fit in will with this work.  We now pass the DAP version obtained from the UI when appropriate.  For example, a DAP draft02 API invocation using taskprov will create a DAP draft02 task.

I'm leaving this as a draft for the moment as I'd like to add end-to-end testing for taskprov-01.


daphne/src/lib.rs
```diff
@@ -213,7 +218,8 @@ impl DapAbort {
             | Self::UnrecognizedAggregationJob
             | Self::UnrecognizedHpkeConfig
             | Self::UnrecognizedMessage
-            | Self::UnrecognizedTask => (self.to_string(), None),
+            | Self::UnrecognizedTask
+            | Self::InvalidTask => (self.to_string(), None),
```
cjpatton, line 222: This list should be kept alphabetized.

daphne/src/roles.rs
```diff
@@ -55,18 +55,29 @@ where
...
 
     /// Look up the DAP task configuration for the given task ID.
     async fn get_task_config_for(
         &'srv self,
         task_id: Cow<'req, Id>,
     ) -> Result<Option<Self::WrappedDapTaskConfig>, DapError> {
-        self.get_task_config_considering_taskprov(task_id, None)
+        // We use DapVersion::Unknown here as we don't know it and we don't need to
+        // know it as we will not be doing any taskprov task creation.
+        self.get_task_config_considering_taskprov(DapVersion::Unknown, task_id, None, |_| Ok(false))
```
cjpatton, line 80: In fact, I think the Helper should always know which version was indicated, via the version field of the `DapRequest`. I suppose the Leader would also know based on its version preference (typically the latest draft).

daphne/src/roles.rs
```diff
@@ -55,18 +55,29 @@ where
...
     /// Look up the DAP task configuration for the given task ID.
-    async fn get_task_config_considering_taskprov(
+    ///
+    /// If a `report` has been provided, then look for the draft-wang-ppm-dap-taskprov-<nn> extension
+    /// in the report.  If a taskprov task configuration is successfully read from the report,
+    /// the `opt_in_callback` will be called, and if it returns Ok(true) the server will opt-in to the task.
+    /// if it returns Ok(false) or an error then the server will opt-out or return an appropriate error.
+    async fn get_task_config_considering_taskprov<F>(
         &'srv self,
+        version: DapVersion,
```
cjpatton, line 65: Why is the version needed?

daphne/src/roles.rs
```diff
@@ -244,8 +255,10 @@ where
         let report = Report::get_decoded(req.payload.as_ref())?;
         let task_config = self
             .get_task_config_considering_taskprov(
+                req.version,
                 Cow::Borrowed(req.task_id()?),
                 Some(&report.metadata),
+                |_| Ok(true),
```
cjpatton, line 261: Worth commenting here that the behavior is "always opt-in". (Similarly below.)

daphne/src/roles.rs
```diff
@@ -55,18 +55,29 @@ where
...
+    /// in the report.  If a taskprov task configuration is successfully read from the report,
+    /// the `opt_in_callback` will be called, and if it returns Ok(true) the server will opt-in to the task.
+    /// if it returns Ok(false) or an error then the server will opt-out or return an appropriate error.
+    async fn get_task_config_considering_taskprov<F>(
         &'srv self,
+        version: DapVersion,
         task_id: Cow<'req, Id>,
         report: Option<&ReportMetadata>,
-    ) -> Result<Option<Self::WrappedDapTaskConfig>, DapError>;
+        opt_in_callback: F,
```
cjpatton, line 68: When we do eventually want to validate the task config (and "opt-out" if we don't like it), where do you imagine this callback coming from? I wonder if it would make more sense to define this as a method on the DapAggregator trait.

cjpatton (commented):
Overall this change looks great! I would say let's keep the versioning infra as-is (we'll want want to replicate this pattern when we have to support multiple versions), but I would say let's drop support for draft-00. It's essentially a dead draft that we don't intend to deploy.

daphne/src/roles.rs
bhalleycf: I considered doing that but went with the callback since the issue called for it, and I wasn't sure if I was missing something or not where the more general callback concept would be useful.  For the code as it is, I think a method on the trait would do, as presumably the Leader DapAggregator would have the comprehensive check and the Helper would just be OK with it.  Happy to do it that way if you don't think we need the generality later.

daphne/src/roles.rs
bhalleycf: The Helper and the Leader do pass the API entry point version used in the cases where taskprov might create a task config.  The "Unknown" version is used in cases where we don't have a report (and thus won't be creating any tasks) and are instead either going to find an existing task (taskprov or not) or find nothing and complain.   The unknown and the false callback are just to adapt the the API so we don't duplicate code.

daphne/src/roles.rs
cjpatton: Makes sense to me. Would you mind adding this context to the docucomment?

daphne/src/roles.rs
cjpatton: I think the generality will be useful. Besides, as of now the callback is hard-coded: To make it more programmable, we'll eventually have to pass the callback through the functions that use it. It would fit more clearly into the current architecture to have this simply be a trait method.

daphne/src/roles.rs
```diff
@@ -54,9 +54,24 @@ where
...
+    ///
+    /// Returning `Ok(true)` opts in, returning `Ok(false)` opts out, and any error is
+    /// also an opt out, but that error code is used instead of InvalidTask.
+    fn taskprov_opt_in_decision(&self, task_config: &DapTaskConfig) -> Result<bool, DapError>;
+
     /// Look up the DAP task configuration for the given task ID.
+    ///
+    /// If a `report` has been provided, then look for the draft-wang-ppm-dap-taskprov-<nn> extension
+    /// in the report.  If a taskprov task configuration is successfully read from the report,
+    /// the `opt_in_callback` will be called, and if it returns Ok(true) the server will opt-in to the task.
```
cjpatton, line 67:
```suggestion
    /// [`DapAggregator::taskprov_opt_in_decision`] will be called, and if it returns Ok(true) the server will opt-in to the task.
```

daphne/src/roles.rs
```diff
@@ -54,9 +54,24 @@ where
...
+    /// if it returns Ok(false) or an error then the server will opt-out or return an appropriate error.
+    ///
+    /// A DapVersion must be specified because we may create a DapTaskConfig via taskprov, and we want it
```
cjpatton, line 70:
```suggestion
    /// The DAP version must be specified because we may create a DapTaskConfig via taskprov, and we want it
```

daphne/src/testing.rs
```diff
@@ -289,8 +290,14 @@ where
         &self.global_config
     }
 
+    fn taskprov_opt_in_decision(&self, _task_config: &DapTaskConfig) -> Result<bool, DapError> {
+        // For now we always opt-in.
+        Ok(true)
```
cjpatton, line 295: Since MockAggregator doesn't properly implement taskprov, I would panic here instead.

cjpatton (approved):
LGTM modulo nits.

Issue #81, opened 2022-08-09
Clean up: Awkward tension between task config version and leader endpoint


cjpatton:
We want that the serialized `DapTaskConfig` matches more or less what is shared between the Aggregators. In particular, our endpoint URL should include the DAP draft in the path. However, the leader also needs the version itself in the task config so that it can produce a `DapRequest` with a version properly.

One way to resolve this might be to make the Version field in the DapRequest optional.

bhalleycf:
This was done as part of #171.

PR #171, opened 2022-11-28
DAP-03 Implementation

bhalleycf:
This is the DAP-03 implementation.

### DIFF
--- a/daphne/src/auth.rs
+++ b/daphne/src/auth.rs
@@ -14,7 +14,7 @@ use crate::{
 use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 
-/// A bearer token used for authorizing DAP requests as specified in draft-ietf-ppm-dap-02.
+/// A bearer token used for authorizing DAP requests.
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct BearerToken {
     raw: String,
--- a/daphne/src/messages/mod.rs
+++ b/daphne/src/messages/mod.rs
@@ -3,9 +3,10 @@
 
 //! Messages in the DAP protocol.
 
+use crate::{DapError, DapVersion};
 use prio::codec::{
     decode_u16_items, decode_u32_items, encode_u16_items, encode_u32_items, CodecError, Decode,
-    Encode,
+    Encode, ParameterizedDecode, ParameterizedEncode,
 };
 use serde::{Deserialize, Serialize};
 use std::{
@@ -24,6 +25,10 @@ const AEAD_ID_AES128GCM: u16 = 0x0001;
 // Query types
 const QUERY_TYPE_TIME_INTERVAL: u8 = 0x01;
 const QUERY_TYPE_FIXED_SIZE: u8 = 0x02;
+
+// FixedSize query subtypes
+const FIXED_SIZE_QUERY_TYPE_BY_BATCH_ID: u8 = 0x00;
+const FIXED_SIZE_QUERY_TYPE_CURRENT_BATCH: u8 = 0x01;
 
 // Known extension types.
 const EXTENSION_TASKPROV: u16 = 0xff00;
@@ -169,10 +174,14 @@ impl Decode for ReportMetadata {
             time: Time::decode(bytes)?,
             extensions: decode_u16_items(&(), bytes)?,
         };
-        // Check for duplicate extensions.
+        // Check for duplicate extensions and unknown extensions.
         let mut seen: HashSet<u16> = HashSet::new();
         for extension in &metadata.extensions {
             if !seen.insert(extension.type_code()) {
+                return Err(CodecError::UnexpectedValue);
+            }
+            if matches!(extension, Extension::Unhandled { .. }) {
+                // Unrecognized extensions are an error.
                 return Err(CodecError::UnexpectedValue);
             }
         }
@@ -238,20 +247,20 @@ impl Decode for ReportShare {
     }
 }
 
-/// Batch parameter conveyed to the Helper by the Leader in the aggreagtion sub-protocol. Used to
+/// Batch parameter conveyed to the Helper by the Leader in the aggregation sub-protocol. Used to
 /// identify which batch the reports in the [`AggregateInitializeReq`] are intended for.
 #[derive(Clone, Debug, Eq, Deserialize, Hash, PartialEq, Serialize)]
 #[serde(rename_all = "snake_case")]
 pub enum PartialBatchSelector {
     TimeInterval,
-    FixedSize { batch_id: Id },
+    FixedSizeByBatchId { batch_id: Id },
 }
 
 impl From<BatchSelector> for PartialBatchSelector {
     fn from(batch_sel: BatchSelector) -> Self {
         match batch_sel {
             BatchSelector::TimeInterval { .. } => Self::TimeInterval,
-            BatchSelector::FixedSize { batch_id } => Self::FixedSize { batch_id },
+            BatchSelector::FixedSizeByBatchId { batch_id } => Self::FixedSizeByBatchId { batch_id },
         }
     }
 }
@@ -260,7 +269,7 @@ impl Encode for PartialBatchSelector {
     fn encode(&self, bytes: &mut Vec<u8>) {
         match self {
             Self::TimeInterval => QUERY_TYPE_TIME_INTERVAL.encode(bytes),
-            Self::FixedSize { batch_id } => {
+            Self::FixedSizeByBatchId { batch_id } => {
                 QUERY_TYPE_FIXED_SIZE.encode(bytes);
                 batch_id.encode(bytes);
             }
@@ -272,10 +281,69 @@ impl Decode for PartialBatchSelector {
     fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
         match u8::decode(bytes)? {
             QUERY_TYPE_TIME_INTERVAL => Ok(Self::TimeInterval),
-            QUERY_TYPE_FIXED_SIZE => Ok(Self::FixedSize {
+            QUERY_TYPE_FIXED_SIZE => Ok(Self::FixedSizeByBatchId {
                 batch_id: Id::decode(bytes)?,
             }),
             _ => Err(CodecError::UnexpectedValue),
+        }
+    }
+}
+
+/// A batch selector issued by the Leader in an aggregate-share request.
+#[derive(Clone, Debug, Deserialize, Hash, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum BatchSelector {
+    TimeInterval { batch_interval: Interval },
+    FixedSizeByBatchId { batch_id: Id },
+}
+
+impl Encode for BatchSelector {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        match self {
+            Self::TimeInterval { batch_interval } => {
+                QUERY_TYPE_TIME_INTERVAL.encode(bytes);
+                batch_interval.encode(bytes);
+            }
+            Self::FixedSizeByBatchId { batch_id } => {
+                QUERY_TYPE_FIXED_SIZE.encode(bytes);
+                batch_id.encode(bytes);
+            }
+        }
+    }
+}
+
+impl Decode for BatchSelector {
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
+        match u8::decode(bytes)? {
+            QUERY_TYPE_TIME_INTERVAL => Ok(Self::TimeInterval {
+                batch_interval: Interval::decode(bytes)?,
+            }),
+            QUERY_TYPE_FIXED_SIZE => Ok(Self::FixedSizeByBatchId {
+                batch_id: Id::decode(bytes)?,
+            }),
+            _ => Err(CodecError::UnexpectedValue),
+        }
+    }
+}
+
+impl Default for BatchSelector {
+    fn default() -> Self {
+        Self::TimeInterval {
+            batch_interval: Interval::default(),
+        }
+    }
+}
+
+impl TryFrom<Query> for BatchSelector {
+    type Error = DapError;
+
+    fn try_from(query: Query) -> Result<Self, DapError> {
+        match query {
+            Query::TimeInterval { batch_interval } => Ok(Self::TimeInterval { batch_interval }),
+            Query::FixedSizeByBatchId { batch_id } => Ok(Self::FixedSizeByBatchId { batch_id }),
+            Query::FixedSizeCurrentBatch => Err(DapError::Fatal(
+                "tried to make a BatchSelector from a FixedSizeCurrentBatch query".to_string(),
+            )),
         }
     }
 }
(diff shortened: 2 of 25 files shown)


daphne/src/auth.rs
```diff
@@ -14,7 +14,7 @@ use crate::{
 use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 
-/// A bearer token used for authorizing DAP requests as specified in draft-ietf-ppm-dap-02.
+/// A bearer token used for authorizing DAP requests as specified in draft-ietf-ppm-dap-03.
```
cjpatton, line 17: Hmm, actually this comment is a bit out-of-date. The bearer token thing goes back to draft-01, but is no longer specified in the current draft.

daphne/src/messages/mod.rs
```diff
@@ -15,6 +15,8 @@ use std::{
     io::{Cursor, Read},
 };
 
+use crate::{DapError, DapVersion};
```
cjpatton, line 18: nit: If you remove the line between this `use` and the previous one, then your linter should automatically put it in alphabetical order.

daphne/src/messages/mod.rs
```diff
@@ -169,12 +175,17 @@ impl Decode for ReportMetadata {
...
-        // Check for duplicate extensions.
+        // Check for duplicate extensions and unknown extensions.
         let mut seen: HashSet<u16> = HashSet::new();
         for extension in &metadata.extensions {
             if !seen.insert(extension.type_code()) {
                 return Err(CodecError::UnexpectedValue);
             }
+            if matches!(extension, Extension::Unhandled { .. }) {
+                // TODO(bhalleycf)  Draft03 Early Input Share Validation says that unrecognized
+                // extensions are an error.  Might ignoring them be better?
```
cjpatton, line 186: This was settled in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/369. (I agree, FWIW, but it was decided to abort.)

daphne/src/messages/mod.rs
```diff
@@ -290,22 +360,33 @@ pub struct AggregateInitializeReq {
...
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
         Ok(Self {
             task_id: Id::decode(bytes)?,
             agg_job_id: Id::decode(bytes)?,
-            agg_param: decode_u16_bytes(bytes)?,
+            agg_param: match decoding_parameter {
+                DapVersion::Draft02 => decode_u16_bytes(bytes)?,
+                DapVersion::Draft03 => decode_u32_bytes(bytes)?,
+                _ => panic!("unimplemented DapVersion"),
```
cjpatton, line 388: ?
```suggestion
                _ => unreachable!("unimplemented DapVersion"),
```

daphne/src/messages/mod.rs
```diff
@@ -290,22 +360,33 @@ pub struct AggregateInitializeReq {
     pub report_shares: Vec<ReportShare>,
 }
 
-impl Encode for AggregateInitializeReq {
-    fn encode(&self, bytes: &mut Vec<u8>) {
+impl ParameterizedEncode<DapVersion> for AggregateInitializeReq {
+    fn encode_with_param(&self, encoding_parameter: &DapVersion, bytes: &mut Vec<u8>) {
```
cjpatton, line 364: Explicit names are better (here and below).
```suggestion
    fn encode_with_param(&self, version: &DapVersion, bytes: &mut Vec<u8>) {
```

daphne/src/messages/mod.rs
```diff
@@ -518,33 +599,60 @@ impl Decode for Interval {
...
         match self {
             Self::TimeInterval { batch_interval } => {
                 QUERY_TYPE_TIME_INTERVAL.encode(bytes);
                 batch_interval.encode(bytes);
             }
-            Self::FixedSize { batch_id } => {
+            Self::FixedSizeByBatchId { batch_id } => {
                 QUERY_TYPE_FIXED_SIZE.encode(bytes);
+                if *encoding_parameter != DapVersion::Draft02 {
+                    FIXEDSIZE_QUERY_TYPE_BY_BATCH_ID.encode(bytes);
```
cjpatton, line 616: For consistency
```suggestion
                    FIXED_SIZE_QUERY_TYPE_BY_BATCH_ID.encode(bytes);
```

daphne/src/messages/mod.rs
```diff
@@ -518,33 +599,60 @@ impl Decode for Interval {
...
+                }
                 batch_id.encode(bytes);
             }
+            Self::FixedSizeCurrentBatch => {
+                if *encoding_parameter == DapVersion::Draft02 {
```
cjpatton, line 621: ?
```suggestion
                if *encoding_parameter != DapVersion::Draft03 {
```

daphne/src/roles.rs
```diff
@@ -326,13 +329,29 @@ where
...
 
+        if collect_req.query == Query::FixedSizeCurrentBatch {
+            // This is where we assign the current batch, and convert the
+            // Query::FixedSizeCurrentBatch into a Query::FixedSizeByBatchId.
+            //
+            // TODO(bhalleycf) Note that currently we are just looking at the
+            // head of the uncollected batch queue, so there is no parallelism
+            // possible for collectors on a given task.  To allow multiple
+            // batches for a task to be collected concurrently for the same task,
+            // we'd need a more complex DO state that allowed us have batch
```
cjpatton, line 340:
```suggestion
            // we'd need a more complex DO state that allowed us to have batch
```

daphne/src/vdaf/mod.rs
```diff
@@ -175,10 +180,17 @@ impl VdafConfig {
...
         }
 
-        const N: usize = CTX_INPUT_SHARE.len();
-        let mut info = [0; N + 2];
-        info[..N].copy_from_slice(CTX_INPUT_SHARE);
-        info[N] = CTX_ROLE_CLIENT; // Sender role (receiver role set below)
+        let input_share_text = match version {
+            DapVersion::Draft02 => CTX_INPUT_SHARE_DRAFT02,
+            DapVersion::Draft03 => CTX_INPUT_SHARE_DRAFT03,
+            _ => panic!("unimplemented DAP version"),
```
cjpatton, line 186: I wouldn't panic here, since the condition is triggered by user error (`version` is unhandled variant).

daphne/src/vdaf/mod.rs
```diff
@@ -802,19 +844,27 @@ impl VdafConfig {
...
+        version: DapVersion,
     ) -> Result<DapAggregateResult, DapError> {
-        const N: usize = CTX_AGG_SHARE.len();
-        let mut info = [0; N + 2];
-        info[..N].copy_from_slice(CTX_AGG_SHARE);
-        info[N + 1] = CTX_ROLE_COLLECTOR; // Receiver role (sender role set below)
+        let agg_share_text = match version {
+            DapVersion::Draft02 => CTX_AGG_SHARE_DRAFT02,
+            DapVersion::Draft03 => CTX_AGG_SHARE_DRAFT03,
+            _ => panic!("unimplemented DAP version"),
```
cjpatton, line 852: I would also not panic here, for the same reason.

daphne/src/testing.rs
```diff
@@ -444,6 +444,10 @@ where
 
         Ok(())
     }
+
+    async fn current_batch(&self, _task_id: &Id) -> std::result::Result<Id, DapError> {
+        panic!("current_batch not implemented yet");
```
cjpatton, line 449: In fact, we already implement something like this: search `current_batch` in this file. I bet it already provides what you need.

daphne/src/roles.rs
```diff
@@ -326,13 +329,29 @@ where
...
+            // state go from unassigned -> in-progress -> complete.
+            collect_req.query = Query::FixedSizeByBatchId {
+                batch_id: self.current_batch(req.task_id()?).await?,
+            };
+        }
+
         // Ensure the batch boundaries are valid and that the batch doesn't overlap with previosuly
         // collected batches.
+        let batch_selector = BatchSelector::try_from(collect_req.query.clone())?;
```
cjpatton, line 349: Can we avoid the `clone()` here?

daphne/src/vdaf/mod.rs
```diff
@@ -175,10 +180,17 @@ impl VdafConfig {
...
-        info[..N].copy_from_slice(CTX_INPUT_SHARE);
-        info[N] = CTX_ROLE_CLIENT; // Sender role (receiver role set below)
+        let input_share_text = match version {
+            DapVersion::Draft02 => CTX_INPUT_SHARE_DRAFT02,
+            DapVersion::Draft03 => CTX_INPUT_SHARE_DRAFT03,
+            _ => panic!("unimplemented DAP version"),
+        };
+        let n: usize = input_share_text.len();
+        let mut info = Vec::new();
+        info.reserve(n + 2);
```
cjpatton, line 190: Ugh, nice idea :)

daphne/src/vdaf/mod.rs
```diff
@@ -850,22 +900,29 @@ fn produce_encrypted_agg_share(
...
         .get_encoded();
 
-    const N: usize = CTX_AGG_SHARE.len();
-    let mut info = [0; N + 2];
-    info[..N].copy_from_slice(CTX_AGG_SHARE);
-    info[N] = if is_leader {
+    let agg_share_text = match version {
+        DapVersion::Draft02 => CTX_AGG_SHARE_DRAFT02,
+        DapVersion::Draft03 => CTX_AGG_SHARE_DRAFT03,
+        _ => panic!("unimplemented DAP version"),
```
cjpatton, line 914: I would not panic here.

daphne/src/vdaf/mod_test.rs
```diff
@@ -138,9 +140,18 @@ async fn roundtrip_report() {
     }
 }
 
-#[test]
-fn roundtrip_report_unsupported_hpke_suite() {
-    let t = Test::new(TEST_VDAF);
+#[tokio::test]
+async fn roundtrip_report_draft02() {
```
cjpatton, line 144: Ok, now I see what you were getting at! Go ahead and de-dup the mod_test.rs tests as you suggested.

daphne_worker/src/dap.rs
```diff
@@ -589,6 +589,10 @@ where
         try_join_all(requests).await.map_err(dap_err)?;
         Ok(())
     }
+
+    async fn current_batch(&self, task_id: &Id) -> std::result::Result<Id, DapError> {
+        self.internal_current_batch(task_id).await
```
cjpatton, line 594: Do you think that we still want `internal_current_batch()`?

daphne_worker/src/lib.rs
```diff
@@ -352,17 +352,42 @@ impl DaphneWorkerRouter {
...
                         let config = DaphneWorkerConfig::from_worker_context(ctx)?;
                         let cmd: InternalTestEndpointForTask = req.json().await?;
-                        config.internal_endpoint_for_task(cmd).await
+                        config
+                            .internal_endpoint_for_task(config.default_version, cmd)
+                            .await
+                    },
+                )
+                .post_async(
+                    "/:version/internal/test/endpoint_for_task",
```
cjpatton, line 361: Why not replace the old endpoint with the new one? The advantage of having one is less code. (Similarly below.)

daphne_worker_test/wrangler.toml
```diff
@@ -47,6 +47,7 @@ DAP_TASKPROV_HPKE_COLLECTOR_CONFIG = """{
 DAP_TASKPROV_VDAF_VERIFY_KEY_INIT = "0074a5dd6e9dac501f73f7a961193b2b" # SECRET
 DAP_TASKPROV_LEADER_BEARER_TOKEN = "I am the leader!" # SECRET
 DAP_TASKPROV_COLLECTOR_BEARER_TOKEN = "I am the collector!" # SECRET
+DAP_DEFAULT_VERSION = "v02"
```
cjpatton, line 50: Perhaps we should default to the latest (v03)?

cjpatton (commented):
Looks great overall! Having  looked at the changes that were needed to the tests, I think de-duplicating the code using macros as you suggested is a great idea.

High level comments/questions:
1. This appears to resolve https://github.com/cloudflare/daphne/issues/81, do you agree?
2. I expect there to be more changes for DAP-03, so be sure to keep the issue open.
3. I think we should add regression tests for DAP-02. At a minimum, I would suggest adding test vectors to vdaf/mod_test.rs for the aggregation flow that check for successful parsing and handling of the messages. (AggregateInitializeReq -> AggregateResp -> AggregateContinueReq -> AggregateResp.) I would also consider doing the same for the collect flow, but this might be harder to implement.
4. It might be useful to exercise some "version negotiation failure" cases, e.g., a DAP-02 client talking to a DAP-03 server. What do you think?

bhalleycf:
I agree it resolves #81.  I also agree about #4 and will ask you about #3 tomorrow.

daphne_worker/src/lib.rs
bhalleycf: I was trying to not break any code relaying on the old endpoint, just in case

daphne_worker/src/dap.rs
bhalleycf: I did this because it means I can leave the config stuff in config, and the trait implementation can call into it.

daphne/src/roles.rs
bhalleycf: Maybe, but we end up needing to clone an Interval or a batch id anyway.  I suggest we leave this as it is until future optimization work, and then only optimize it if it is hot.

daphne/src/messages/mod.rs
```diff
@@ -361,13 +359,13 @@ pub struct AggregateInitializeReq {
...
-    fn encode_with_param(&self, encoding_parameter: &DapVersion, bytes: &mut Vec<u8>) {
+    fn encode_with_param(&self, version: &DapVersion, bytes: &mut Vec<u8>) {
         self.task_id.encode(bytes);
         self.agg_job_id.encode(bytes);
-        match encoding_parameter {
+        match version {
             DapVersion::Draft02 => encode_u16_bytes(bytes, &self.agg_param),
             DapVersion::Draft03 => encode_u32_bytes(bytes, &self.agg_param),
-            _ => panic!("unimplemented DapVersion"),
+            _ => unreachable!(),
```
cjpatton, line 368: I believe you can still output a message, which might be helpful if in fact we manage to panic here.
```suggestion
            _ => unreachable!("unimplemented version of DAP"),
```

daphne/src/roles_test.rs
```diff
@@ -1614,15 +1424,7 @@ async fn http_post_fail_wrong_dap_version(version: DapVersion) {
...
-#[tokio::test]
-async fn http_post_fail_wrong_dap_version_draft02() {
-    http_post_fail_wrong_dap_version(DapVersion::Draft02).await
-}
-
-#[tokio::test]
-async fn http_post_fail_wrong_dap_version_draft03() {
-    http_post_fail_wrong_dap_version(DapVersion::Draft03).await
-}
+async_test_versions! { http_post_fail_wrong_dap_version }
```
cjpatton, line 1427: That's pretty cool :)

cjpatton (commented):
Oh man, this change is looking so snazzy. I'm pretty much ready to approve, but I want to run the end-to-end tests first. Can you rebase?

cjpatton (approved):
🦆

One nit: commit message should say "begin implementation", as DAP-03 is not complete as of this change.

PR #172, opened 2022-11-29
Implement taskprov for MockAggregator

cjpatton:
MockAggregator stores all tasks in the same pool, regardless of how the tasks are configured.

To allow dynamic task configuration, MockAggregator::tasks is now guraded by a Mutex. For convenience, we revise the type of MockAggregator::WrappedDapTaskConfig to force cloning of the task config as needed.


bhalleycf (approved):

cjpatton:
Squashed.

PR #174, opened 2022-12-05
daphne_worker: Add /task endpoint

### LINKED ISSUE #82: Task provisioning

cjpatton:
Closes #82.

Like /internal/test/add_task, but requires a bearer token. The header for the token is X-Daphne-Worker-Admin-Bearer-Token; on the server side, the token is stored in DAP_ADMIN_BEARER_TOKEN. This allows us to manually configure a task in a live deployment without exposing the internal test endpoints.

### DIFF
--- a/daphne_worker_test/wrangler.toml
+++ b/daphne_worker_test/wrangler.toml
@@ -77,6 +77,7 @@ preview_id = "<ignored>" # TODO(cjpatton) Figure out how to pick this.
 # production. In particular, they will not be passed as environment variables
 # as they are here. See
 # https://developers.cloudflare.com/workers/wrangler/commands/#secret.
+DAP_ADMIN_BEARER_TOKEN = "administrator bearer token" # SECRET
 DAP_AGGREGATOR_ROLE = "helper"
 DAP_BASE_URL = "http://127.0.0.1:8788/"
 DAP_ISSUE73_DISABLE_AGG_JOB_QUEUE_GARBAGE_COLLECTION = "true"
--- a/daphne_worker/src/config.rs
+++ b/daphne_worker/src/config.rs
@@ -105,6 +105,9 @@ pub(crate) struct DaphneWorkerConfig<D> {
 
     /// Default DAP version to use if not specified by the API URL
     pub(crate) default_version: DapVersion,
+
+    /// Admin bearer token. If configured, it is used to authorize requests from the administrator.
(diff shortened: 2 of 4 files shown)


chris-wood:
Can I suggest that this endpoint be named simply `/task`, and that new tasks are created by POSTing to that endpoint with appropriate credentials?

daphne_worker_test/wrangler.toml
```diff
@@ -77,6 +77,7 @@ preview_id = "<ignored>" # TODO(cjpatton) Figure out how to pick this.
 # production. In particular, they will not be passed as environment variables
 # as they are here. See
 # https://developers.cloudflare.com/workers/wrangler/commands/#secret.
+DAP_ADMIN_BEARER_TOKEN = "administrator bearer token" # SECRET
```
bhalleycf, line 80: Do we need to set this for the leader too?

cjpatton: This feature is optional: If not set, then tasks can't be configured this way.

bhalleycf (approved):
looks ok to me other than my question about the wrangler config

cjpatton:
The new feature is optional: If the bearer token is missing from the config, then tasks can't be configured this way.

cjpatton:
Do you think we should make it mandatory instead?

cjpatton:
> Can I suggest that this endpoint be named simply `/task`, and that new tasks are created by POSTing to that endpoint with appropriate credentials?

Whoops, sorry I missed this @chris-wood! 

I'm fine with calling it `/task`. 

> ... and that new tasks are created by POSTing to that endpoint with appropriate credentials?

I'm confused: What's the difference between your suggestion and what the PR currently does?

chris-wood:
Nothing beyond the name!

bhalleycf:
I don't think it needs to be mandatory, I was just confused by it being supported by both leader and helper, but only configured in the helper in wrangler.toml.

cjpatton:
Endpoint changed to /task.

PR #176, opened 2022-12-07
Make keys objects to hpke_rs crate

### LINKED ISSUE #160: Clean up: Validate public key pairs
See https://github.com/cloudflare/daphne/pull/152#discussion_r1014292934.

armfazh:
fixes #160

### DIFF
--- a/daphne/src/hpke.rs
+++ b/daphne/src/hpke.rs
@@ -69,23 +69,21 @@ impl HpkeConfig {
         plaintext: &[u8],
     ) -> Result<(Vec<u8>, Vec<u8>), DapError> {
         let sender: Hpke<ImplHpkeCrypto> = check_suite(self.kem_id, self.kdf_id, self.aead_id)?;
-        let pk = HpkePublicKey::new(self.public_key.clone());
-        let (enc, mut ctx) = sender.setup_sender(&pk, info, None, None, None)?;
+        let (enc, mut ctx) = sender.setup_sender(&self.public_key, info, None, None, None)?;
         let ciphertext = ctx.seal(aad, plaintext)?;
         Ok((enc, ciphertext))
     }
 
     pub(crate) fn decrypt(
         &self,
-        secret_key: &[u8],
+        private_key: &HpkePrivateKey,
         info: &[u8],
         aad: &[u8],
         enc: &[u8],
         ciphertext: &[u8],
     ) -> Result<Vec<u8>, DapError> {
         let receiver: Hpke<ImplHpkeCrypto> = check_suite(self.kem_id, self.kdf_id, self.aead_id)?;
-        let sk = HpkePrivateKey::new(secret_key.to_vec());
-        let mut ctx = receiver.setup_receiver(enc, &sk, info, None, None, None)?;
+        let mut ctx = receiver.setup_receiver(enc, private_key, info, None, None, None)?;
         let plaintext = ctx.open(aad, ciphertext)?;
         Ok(plaintext)
     }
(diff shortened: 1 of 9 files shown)


daphne_worker_test/wrangler.toml
```diff
@@ -40,7 +40,14 @@ DAP_TASKPROV_HPKE_COLLECTOR_CONFIG = """{
   "kem_id": "p256_hkdf_sha256",
   "kdf_id": "hkdf_sha256",
   "aead_id": "aes128_gcm",
-  "public_key": "047dab625e0d269abcc28c611bebf5a60987ddf7e23df0e0aa343e5774ad81a1d0160d9252b82b4b5c52354205f5ec945645cb79facff8d85c9c31b490cdf35466"
+  "public_key": { "value": [
```
cjpatton, line 43: This isn't going to work for us unfortunately; we need to be able to encode the public key in hex. There should be some serde magic that can help: Search for `#[serde(with = "hex")]` in the repo. You may need to add a wrapper object: see https://serde.rs/container-attrs.html#try_from.

daphne_worker_test/wrangler.toml
```diff
@@ -3,7 +3,7 @@ main = "build/worker/shim.mjs"
 compatibility_date = "2022-10-14"
 
 [build]
-command = "cargo install --git https://github.com/cloudflare/workers-rs && worker-build"
+command = "cargo install --git https://github.com/cloudflare/workers-rs --version 0.0.8 worker-build && worker-build"
```
cjpatton, line 6: What's the reasoning for this change?

cjpatton (commented):
Looking pretty good! One blocking comment (change in HpkeConfig JSON serialization) and one question so far.

daphne/src/hpke.rs
```diff
@@ -181,9 +177,23 @@ impl HpkeReceiverConfig {
         }
     }
 
-    /// Create a new HPKE receiver context given an HpkeConfig and a corresponding secret key.
-    pub fn new(config: HpkeConfig, secret_key: Vec<u8>) -> Self {
-        HpkeReceiverConfig { config, secret_key }
+    /// Create a new HPKE receiver context given an HpkeConfig and a corresponding private key.
+    /// Returns an error if the public key does not correspond to the private_key.
+    pub fn try_from(config: HpkeConfig, private_key: HpkePrivateKey) -> Result<Self, DapError> {
```
cjpatton, line 182: readability: Rust best practice is to not overload the names of common trait methods (i.e., TryFrom::try_from() in this case). Could we call this something like try_from_parts(), or maybe try_from_config_and_key()? Alternatively, we could implement `TryFrom<(HpkeConfig, HpkePrivateKey)> for `HpkeReceiverConfig`.

armfazh: done

daphne_worker_test/wrangler.toml
armfazh: it compiles only worker-build with specific version since workers-rs is a set of crates.
and does the same in the Dockerfile. 
By doing so, worker-build is fetched and compiled once.
But these are local changes, and they are not related to this pr.

daphne_worker_test/wrangler.toml
armfazh: done

daphne_worker_test/wrangler.toml
cjpatton: In that case, could you revert? (1) I'd like to keep this change minimal, and (2) in the past we have needed to track particular commits in workers-rs, and when we do this, we need to make sure to build the same version of worker-build.

daphne/src/hpke.rs
```diff
@@ -239,3 +253,37 @@ impl std::str::FromStr for HpkeReceiverConfig {
...
+struct HpkePrivateKeySerde(#[serde(getter = "HpkePrivateKeySerde::to_vec", with = "hex")] Vec<u8>);
+
+impl HpkePrivateKeySerde {
+    fn to_vec(x: &HpkePrivateKey) -> Vec<u8> {
+        x.as_slice().into()
+    }
+}
+
+impl From<HpkePrivateKeySerde> for HpkePrivateKey {
+    fn from(k: HpkePrivateKeySerde) -> HpkePrivateKey {
```
cjpatton, line 286:
```suggestion
    fn from(private_key_serde: HpkePrivateKeySerde) -> HpkePrivateKey {
```

cjpatton (approved):
Looks great, thank you!

PR #179, opened 2022-12-08
Move extensions into input shares per DAP Draft03.

bhalleycf:


daphne/src/messages/mod.rs
```diff
@@ -157,23 +157,33 @@ impl Decode for Extension {
...
-    fn encode(&self, bytes: &mut Vec<u8>) {
+impl ParameterizedEncode<DapVersion> for ReportMetadata {
+    fn encode_with_param(&self, version: &DapVersion, bytes: &mut Vec<u8>) {
         self.id.encode(bytes);
         self.time.encode(bytes);
-        encode_u16_items(bytes, &(), &self.extensions);
+        if matches!(version, DapVersion::Draft02) {
+            encode_u16_items(bytes, &(), &self.extensions);
+        } else if !self.extensions.is_empty() {
+            panic!("tried to encode extensions in the ReportMetadata for DAP > Draft02")
```
cjpatton, line 170: nit
```suggestion
            panic!("tried to encode extensions in the ReportMetadata for DAP > draft-02")
```

daphne/src/messages/mod.rs
```diff
@@ -971,6 +987,43 @@ impl Decode for HpkeCiphertext {
     }
 }
 
+/// A plaintext input share.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
```
cjpatton, line 991: Hmm, do we need `Serialize, Deserialize` for this struct?

daphne/src/messages/mod.rs
```diff
@@ -971,6 +987,43 @@ impl Decode for HpkeCiphertext {
...
+            payload: decode_u32_bytes(bytes)?,
+        };
+        // Check for duplicate extensions and unknown extensions.
+        let mut seen: HashSet<u16> = HashSet::new();
+        for extension in &share.extensions {
+            if !seen.insert(extension.type_code()) {
+                return Err(CodecError::UnexpectedValue);
+            }
+            if matches!(extension, Extension::Unhandled { .. }) {
+                // Unrecognized extensions are an error.
```
cjpatton, line 1019:
```suggestion
```

daphne/src/vdaf/mod.rs
```diff
@@ -172,17 +173,32 @@ impl VdafConfig {
...
-        let encoded_input_shares = match self {
+        let mut encoded_input_shares = match self {
             Self::Prio3(prio3_config) => prio3_shard(prio3_config, measurement)?,
             Self::Prio2 { dimension } => prio2_shard(*dimension, measurement)?,
         };
+        if version != DapVersion::Draft02 {
+            let mut encoded: Vec<Vec<u8>> = Vec::new();
+            for share in encoded_input_shares.drain(..) {
+                let input_share = PlaintextInputShare {
+                    extensions: extensions.clone(),
```
cjpatton, line 195: If I'm reading this code correctly, it should be possible to avoid this clone.
```suggestion
                    extensions,
```

daphne/src/vdaf/mod.rs
```diff
@@ -172,17 +173,32 @@ impl VdafConfig {
...
 
         let public_share = Vec::new();
-        let encoded_input_shares = match self {
+        let mut encoded_input_shares = match self {
             Self::Prio3(prio3_config) => prio3_shard(prio3_config, measurement)?,
             Self::Prio2 { dimension } => prio2_shard(*dimension, measurement)?,
         };
+        if version != DapVersion::Draft02 {
+            let mut encoded: Vec<Vec<u8>> = Vec::new();
+            for share in encoded_input_shares.drain(..) {
```
cjpatton, line 193: I think into_iter() is more conventional here.
```suggestion
            for share in encoded_input_shares.into_iter() {
```

daphne_worker/src/dap.rs
```diff
@@ -595,6 +595,19 @@ where
     }
 }
 
+fn task_id_from_report(report: &[u8]) -> std::result::Result<Id, DapError> {
+    // The task id is the first 32 bytes of the serialized report.
```
cjpatton, line 599: This is true for drafts 02 and 03, but it may change in the future. To prevent ourselves from shooting ourselves in the foot later, it might be useful to fence this method to those versions somehow. Maybe pass the version here and err unless we call on a compatible version?

daphne_worker/src/dap.rs
```diff
@@ -595,6 +595,19 @@ where
...
+    if report.len() < 32 {
+        // Should never happen!
+        return Err(DapError::Fatal(
+            "serialized report is too short".to_string(),
+        ));
+    }
+    let mut b: [u8; 32] = [0; 32];
+    b[..32].copy_from_slice(&report[..32]);
```
cjpatton, line 607: Slightly simpler (I'm pretty sure this is correct, you'll have to see if cargo disagrees!)
```suggestion
    let id = Id(&report[..33].try_into().map_err(|_| DapErr::fatal("serialize report is too short"))?);
```

daphne_worker/src/dap.rs
```diff
@@ -672,10 +685,13 @@ where
                 .map_err(dap_err)?;
 
             for report_hex in reports_from_durable {
-                let report = Report::get_decoded(&hex::decode(&report_hex).map_err(|_| {
+                let dehexed = hex::decode(&report_hex).map_err(|_| {
```
cjpatton, line 688: nit: Maybe `report_bytes` so that readers don't accuse us of witchcraft.

daphne_worker/src/dap.rs
```diff
@@ -672,10 +685,13 @@ where
...
                     DapError::fatal("response from ReportsPending is not valid hex")
-                })?)?;
-
+                })?;
+                let task_id = task_id_from_report(&dehexed)?;
```
cjpatton, line 691: I was thinking about whether we can avoid this copy here, and I don't think there's a way, sadly. `get_task_config_for()` takes in a `Cow`, which let's us use a reference to an `Id` instead of an owned `Id`, but I don't know if we can coerce a slice into a `&[u8; 32]` without a copy.

cjpatton (commented):
This is in great shape, just a few nit-picky things.

daphne_worker/src/dap.rs
cjpatton: Also it's more consistent with naming elsewhere in the crate :)

daphne/src/messages/mod.rs
bhalleycf: we don't, I'll remove it

daphne/src/vdaf/mod.rs
bhalleycf: We're added the passed in extensions into every report, and they are not references in the reports, so we have to clone().  If I take away the clone, it complains that I've moved the value in a previous iteration of the loop.

daphne_worker/src/dap.rs
bhalleycf: We can't do that, as we're calling this function to get the task id, so we can find the task, and thus learn the version.  I strengthened the comment to note that if we want to change the layout in the future, we have to change the serialization form of the DO so that it has an explicit version.

daphne_worker/src/dap.rs
bhalleycf: I don't know how to do it either.

daphne_worker/src/dap.rs
```diff
@@ -595,6 +595,16 @@ where
     }
 }
 
+fn task_id_from_report(report: &[u8]) -> std::result::Result<Id, DapError> {
+    // The task id MUST BE the first 32 bytes of the serialized report; if this
+    // needs to change in the future, then we must change the DO serialization
+    // format to contain a version.
```
cjpatton, line 601: 👍

cjpatton (approved):

PR #181, opened 2022-12-15
Fix clippy lint from Rust 1.66.0.

bhalleycf:


daphne/src/vdaf/mod.rs
```diff
@@ -352,7 +352,7 @@ impl VdafConfig {
             _ => PlaintextInputShare::get_decoded(&encoded_input_share)?,
         };
 
-        let agg_id = if is_leader { 0 } else { 1 };
+        let agg_id = usize::from(!is_leader);
```
cjpatton, line 355: Thanks for that one clippy 👍

cjpatton (approved):

PR #183, opened 2022-12-15
Add plumbing for landing page

cjpatton:
Thanks to @cdrubin for the initial implementation.


cjpatton:
I tested manually but didn't write end-to-end tests. The reason is that enabling the landing page requires disabling 404-by-default, which has been helpful for debugging in the past.

bhalleycf (approved):

PR #194, opened 2023-01-11
Separate the early metadata check from the worker implementation.

bhalleycf:
Add unit testing of the check now that it is easier to test.

### DIFF
--- a/daphne/src/roles.rs
+++ b/daphne/src/roles.rs
@@ -992,3 +992,31 @@ where
 
     Ok(())
 }
+
+/// Check for transition failures due to:
+///
+/// * the report having already been processed
+/// * the report having already been collected
+/// * the report not being within time bounds
+///
+/// Returns `Some(TransitionFailure)` if there is a problem,
+/// or `None` if no transition failure occurred.
+pub fn early_metadata_check(
+    metadata: &ReportMetadata,
+    processed: bool,
+    collected: bool,
+    min_time: u64,
+    max_time: u64,
+) -> Option<TransitionFailure> {
+    if processed {
+        Some(TransitionFailure::ReportReplayed)
+    } else if collected {
+        Some(TransitionFailure::BatchCollected)
+    } else if metadata.time < min_time {
+        Some(TransitionFailure::ReportDropped)
+    } else if metadata.time > max_time {
+        Some(TransitionFailure::ReportTooEarly)
+    } else {
+        None
+    }
+}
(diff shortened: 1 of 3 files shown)


daphne/src/roles.rs
```diff
@@ -992,3 +992,28 @@ where
 
     Ok(())
 }
+
+/// Check for transition failures due to:
+///
+/// * the report having already been processed
+/// * the report having already been collected
+/// * the report not being within time bounds
```
cjpatton, line 1000:
```suggestion
/// * the report not being within time bounds
///
/// Returns `None` if no transition failure occurred.
```

daphne/src/roles.rs
```diff
@@ -992,3 +992,28 @@ where
...
+pub fn early_metadata_check(
+    metadata: &ReportMetadata,
+    processed: bool,
+    collected: bool,
+    min_time: u64,
```
cjpatton, line 1005: Suggestion: To make this method a bit more ergonomic, we might consider making `early_metadata_check()` a derived method on trait `DapAggregator` so that, instead of computing `min_time` and `max_time` externally, we could could have this method sample the clock and compute the bounds itself.

Would this work? Or is the computation of the time bounds Daphne-Worker specific? We might require `DapAggregator` to specify bounds for the window if time it'll accept reports.

daphne/src/roles_test.rs
```diff
@@ -1593,3 +1594,75 @@ async fn e2e_taskprov(version: DapVersion) {
...
+    let t = Test::new(version);
+    let mut rng = thread_rng();
+    let metadata = ReportMetadata {
+        id: ReportId(rng.gen()),
+        time: t.now,
+        extensions: vec![],
+    };
+    // A current not processed, not collected report is OK!
+    assert_matches!(
+        early_metadata_check(&metadata, false, false, t.now - 100, t.now + 100),
```
cjpatton, line 1608: Readability: Variables improve understanding of the test without having to jump to the definition of `early_metadata_check`. 
```suggestion
        processed = false;
        collected = false;
        early_metadata_check(&metadata, processed, collected, t.now - 100, t.now + 100),
```

daphne/src/roles.rs
bhalleycf: We probably could do that, and I considered it, but I went the way I did as I wanted easy, clock-independent unit testing of the key policy.  The current policy on computing time bounds, though implemented in the worker, could be moved to the Daphne crate as it only uses global configuration.  I'm not sure that's a good idea though, as one could imagine future time bounds policy having some worker-specific inputs, e.g. total storage used.

cjpatton (approved):

PR #195, opened 2023-01-11
Sort dependency lists alphabetically

cjpatton:

PR #196, opened 2023-01-12
daphne: Add plumbing for Prometheus metrics

### LINKED ISSUE #27: Add Prometheus metrics
Prometheus metrics need to be added in order to monitor request latency, error rates, and backlog size. We would also like to replace `DapLeaderProcessTelemetry`, which is used for end-to-end tests.

- [x] Instrument Daphne with a Prometheus client (perhaps https://docs.rs/prometheus/latest/prometheus/)
- [ ] Have Daphne-Worker push metrics to an endpoint configured by an environment variable

cjpatton:
Based on #195.
Partially closes #27.

Defines a struct `DaphneMetrics` where any metrics recorded while processing or sending a DAP request will be enumerated. A registry is required to construct this type. All of its members are crate-private.

Adds a new method to the `DapAggregator` that provides access to the `DaphneMetrics` in use. This intended to be used by derived methods that implement the protocol logic.

It's more common to see the ["static metric" pattern](https://docs.rs/prometheus/latest/prometheus/index.html#static-metrics) for this library. The advantage of non-static metrics is that they are not re-used across unit tests, which makes it possible to write unit tests for the metrics.

### DIFF
--- a/daphne/src/metrics.rs
+++ b/daphne/src/metrics.rs
@@ -1,0 +1,29 @@
+// Copyright (c) 2022 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+//! Daphne metrics.
+
+use crate::DapError;
+use prometheus::{register_int_counter_vec_with_registry, IntCounterVec, Registry};
+
+pub struct DaphneMetrics {
+    /// Report metrics. How many reports have been rejected, aggregated, and collected. When
+    /// a report is rejected, the failure type is recorded.
+    //
+    // TODO(cjpatton) So far we only count the number of reports aggregated.
+    pub(crate) report_counter: IntCounterVec,
+}
+
+impl DaphneMetrics {
+    /// Regstier Daphne metrics with the specified registry.
+    pub fn register(registry: &Registry) -> Result<Self, DapError> {
+        let report_counter = register_int_counter_vec_with_registry!(
+            "daphne_report_counter",
+            "Total number reports rejected, aggregated, and collected.",
+            &["status"],
+            registry
+        )?;
+
+        Ok(Self { report_counter })
+    }
+}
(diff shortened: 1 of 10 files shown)


cjpatton:
Not yet ready to merge, as the end-to-end tests are unhappy with using the default registry:

```
---- e2e_fixed_size_no_current_draft03 stdout ----
leader_url = http://leader:8787/v03/
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `200`,
 right: `500`: unexpected response status: "failed to register daphne metrics: fatal error: prometheus: Duplicate metrics collector registration attempted"', daphne_worker_test/tests/test_runner.rs:689:5
```

I'm investigating. In the meantime, @bhalleycf it would be great if you could take a look and tell me if this is going in the right direction.

For the record, `prometheus` dependency adds about a third of a MB to the size of the uncompressed worker.

daphne/src/metrics.rs
```diff
@@ -0,0 +1,29 @@
...
+    //
+    // TODO(cjpatton) So far we only count the number of reports aggregated.
+    pub(crate) report_counter: IntCounterVec,
+}
+
+impl DaphneMetrics {
+    /// Regstier Daphne metrics with the specified registry.
+    pub fn register(registry: &Registry) -> Result<Self, DapError> {
+        let report_counter = register_int_counter_vec_with_registry!(
+            "report_counter",
```
cjpatton, line 21: Consider prefixing with "daphne_".

cjpatton: Done

cjpatton:
> Not yet ready to merge, as the end-to-end tests are unhappy with using the default registry:

It turns out the problem is a single worker re-registering the same metrics with default registry across requests. Creating a custom registry for each request seems to be the way to go. This also appears to be the pattern that promJS follows: https://github.com/weaveworks/promjs

bhalleycf (approved):

PR #198, opened 2023-01-17
Add basic infrastructure for future tracing work.

bhalleycf:
Tracing is configured with the optional DAP_LOG worker variable.

All existing logging has been converted to the tracing system.

### DIFF
--- a/daphne_worker/src/config.rs
+++ b/daphne_worker/src/config.rs
@@ -38,6 +38,7 @@ use std::{
     sync::{Arc, RwLock, RwLockReadGuard},
     time::Duration,
 };
+use tracing::{debug, info};
 use worker::{kv::KvStore, *};
 
 pub(crate) const KV_KEY_PREFIX_HPKE_RECEIVER_CONFIG: &str = "hpke_receiver_config";
@@ -172,7 +173,7 @@ impl DaphneWorkerConfig {
             DaphneWorkerDeployment::default()
         };
         if !matches!(deployment, DaphneWorkerDeployment::Prod) {
-            console_debug!("DAP deployment override applied: {:?}", deployment);
+            info!("DAP deployment override applied: {:?}", deployment);
         }
 
         let taskprov = if global.allow_taskprov {
@@ -207,7 +208,7 @@ impl DaphneWorkerConfig {
         let admin_token = match env.secret("DAP_ADMIN_BEARER_TOKEN") {
             Ok(raw) => Some(BearerToken::from(raw.to_string())),
             Err(err) => {
-                console_log!("DAP_ADMIN_BEARER_TOKEN not configured: {:?}", err);
+                info!("DAP_ADMIN_BEARER_TOKEN not configured: {:?}", err);
                 None
             }
         };
@@ -538,7 +539,7 @@ impl<'srv> DaphneWorker<'srv> {
                 .delete(kv_key.name.as_str())
                 .await
                 .map_err(|e| DapError::Fatal(format!("kv_store: {}", e)))?;
-            console_debug!("deleted KV item {}", kv_key.name);
+            debug!("deleted KV item {}", kv_key.name);
         }
 
         future_delete_durable.await.map_err(dap_err)?;
--- a/daphne_worker/src/lib.rs
+++ b/daphne_worker/src/lib.rs
@@ -152,6 +152,7 @@
 //! | `DAP_REPORT_SHARD_COUNT` | `u64` | no | Number of report shards per storage epoch. |
 //! | `DAP_REPORT_SHARD_KEY` | `String` | yes | Hex-encoded key used to hash a report into one of the report shards. |
 use crate::{config::DaphneWorkerState, dap::dap_response_to_worker};
+use chrono::{SecondsFormat, Utc};
 use daphne::{
     auth::BearerToken,
     constants,
@@ -161,6 +162,9 @@ use daphne::{
 };
 use prio::codec::Encode;
 use serde::{Deserialize, Serialize};
+use std::{io, str, sync::Once};
+use tracing::{debug, error, info};
+use tracing_subscriber::{fmt, layer::*, prelude::*, registry, EnvFilter};
 use worker::*;
 
 /// Parameters used by the Leader to select a set of reports for aggregation.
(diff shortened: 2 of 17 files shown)


bhalleycf:
hmm, probably DAP_LOG should be DAP_TRACING :)

daphne_worker_test/src/utils.rs
```diff
@@ -13,3 +16,67 @@ cfg_if! {
...
+    fn new() -> Self {
+        LogWriter {
+            buffer: String::new(),
+        }
+    }
+}
+
+impl io::Write for LogWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        // TODO(bhalley)  This is not very efficient, but it is simple!
```
cjpatton, line 40: Why is this inefficient?

bhalleycf: "inefficient" might have been too strong :).   It always accumulates the buffer into the pending output buffer, and then checks if the chunk being written ends a line.  A more optimized bit of code might deal with the common case of a whole line being written in one chunk by avoiding the extra copying.  Another simplification in this code is it assumes that whatever is generating things is also line buffering, e.g. that we're not going to get a chunk like "something\nsomething else" (without the final \n")

daphne_worker/src/config.rs
```diff
@@ -207,7 +208,7 @@ impl DaphneWorkerConfig {
         let admin_token = match env.secret("DAP_ADMIN_BEARER_TOKEN") {
             Ok(raw) => Some(BearerToken::from(raw.to_string())),
             Err(err) => {
-                console_log!("DAP_ADMIN_BEARER_TOKEN not configured: {:?}", err);
+                error!("DAP_ADMIN_BEARER_TOKEN not configured: {:?}", err);
```
cjpatton, line 211: Whether this is an error depends on how DaphneWorker is configured. I would suggest `info`.

daphne_worker/src/lib.rs
```diff
@@ -211,7 +214,7 @@ impl DaphneWorkerRouter {
     /// ```ignore
     /// use daphne_worker::DaphneWorkerRouter;
     /// use worker::*;
-    ///
+    ///cons
```
cjpatton, line 217: Typo?

daphne_worker/src/lib.rs
```diff
@@ -221,13 +224,17 @@ impl DaphneWorkerRouter {
     //
     // TODO Document endpoints that aren't defined in the DAP spec
     pub async fn handle_request(&self, req: Request, env: Env) -> Result<Response> {
+        // Ensure that tracing is initialized.  Some callers may choose to initialize earlier,
+        // but it's safe and cheap to call initialize_tracing() more than once, and this ensures
+        // it's definitely ready for use even if the caller hasn't done anything.
+        initialize_tracing(&env, "DAP_TRACING", "info");
```
cjpatton, line 230: * What's the name we should use for "info"? The default tracing level, maybe?
* What do you think about having `DaphneWorkerRouter` specify the default logging level?

daphne_worker/src/lib.rs
```diff
@@ -512,6 +519,70 @@ pub(crate) struct InternalTestAddTask {
     task_expiration: Time,
 }
 
+/// LogWriter helps us write to the worker console.
+///
+/// It provides and io::Write implementation that provides line
+/// buffering.  It also takes care of emitting a timestamp as the
```
cjpatton, line 525:
```suggestion
/// buffering. It also takes care of emitting a timestamp as the
```

daphne_worker/src/lib.rs
```diff
@@ -512,6 +519,70 @@ pub(crate) struct InternalTestAddTask {
     task_expiration: Time,
 }
 
+/// LogWriter helps us write to the worker console.
+///
+/// It provides and io::Write implementation that provides line
```
cjpatton, line 524: nit: My rust linter wraps  comments at 100 characters.

cjpatton (commented):
* Is tracing initiated properly for DOs? Since they run separately from the main Worker, so they might not be. (It seems to be in miniflare, but it might be worth checking this out in the live staging deployment.)
* It would probably be useful if we could have tracing in the `daphne` crate as well. Would this be feasible? We would need a different subscriber for `MockAggregator`. Perhaps if we had `DapAggregator` specify a subscriber?

daphne_worker/src/lib.rs
bhalleycf: That's not good for the DO case.  I'm thinking maybe both the variable name and the default should just be embedded in initialize_tracing as because of the DOs we are goign to have to call it it like 8 more places.

daphne_worker/src/config.rs
bhalleycf: Oh, and we need to initialize tracing for the DOs; I'll add that.  Good catch!

Tracing in the daphne crate will work and I've tested it already, though it's not in this PR.

daphne_worker/src/lib.rs
```diff
@@ -512,6 +519,70 @@ pub(crate) struct InternalTestAddTask {
...
+        let s = str::from_utf8(buf).map_err(|e| io::Error::new(io::ErrorKind::InvalidInput, e))?;
+        self.buffer.push_str(s);
+        // We assume the caller is doing line buffering and that any write containing a NL will have it at the end.
+        if self.buffer.ends_with('\n') {
+            let mut date = Date::now().to_string();
+            // The format we get ends with the long time zone name in parens, e.g. " (Pacific Standard Time)".
+            // This is noisy, so if we see it, chop it off.
+            if let Some(index) = date.find(" (") {
+                date.truncate(index);
+            }
```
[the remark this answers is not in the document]
bhalleycf: That's an aspect of the javascript environment I think.  I'm not sure I get to pick or can tell, but maybe!  I was hoping that a metal would just be set to UTC and it would use that.  But in the meantime the format it uses was making me have an excessively wide terminal and not adding value, so I hacked it away

daphne_worker_test/wrangler.toml
```diff
@@ -20,6 +20,7 @@ fallthrough = false
 # production. In particular, they will not be passed as environment variables
 # as they are here. See
 # https://developers.cloudflare.com/workers/wrangler/commands/#secret.
+DAP_ADMIN_BEARER_TOKEN = "administrator bearer token" # SECRET
```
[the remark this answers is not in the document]
bhalleycf: It was making noise in the test logs because it was an error; I can remove it.  Or just demote the message from an error to info as you suggested.

daphne_worker/src/lib.rs
```diff
@@ -512,6 +519,70 @@ pub(crate) struct InternalTestAddTask {
...
+
+static INITIALIZE_LOGGING: Once = Once::new();
+
+/// Setup logging.
+///
+/// Initialize logging using configuration from binding `binding` if present, otherwise
+/// using the configuration in `default`.
+///
+/// Panics if the log handler cannot be installed.
+pub fn initialize_tracing(env: &Env, binding: &str, default: &str) {
```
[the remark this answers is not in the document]
bhalleycf: Nothing would be logged as there would be no subscriber, but nothing would break.

daphne_worker/src/lib.rs
bhalleycf: I fixed this and adjust my ide though we may want to consider turning on autoformatting of comments by the rustfmt.

daphne_worker/src/lib.rs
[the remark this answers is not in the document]
cjpatton: Got it! Yeah it's fine to leave as-is for now. Definitely keep the comment for reference later.

PR #202, opened 2023-01-20
Test harness improvements

cjpatton:
Partially addresses #27.

* Improve test coverage of `Test` in `roles_test.rs`
* Test for presence of report count metrics on the Leader

### DIFF
--- a/daphne/src/roles_test.rs
+++ b/daphne/src/roles_test.rs
@@ -422,7 +422,7 @@ impl Test {
             )
             .await;
 
-        // Leader: Handle request form Collector.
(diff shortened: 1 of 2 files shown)


daphne/src/roles_test.rs
```diff
@@ -483,64 +422,16 @@ impl Test {
             )
             .await;
 
-        // Handle request.
+        // Leader: Handle request form Collector.
```
bhalleycf, line 425: "from" not "form"

bhalleycf (approved):

PR #207, opened 2023-01-23
Add report rejection metrics

cjpatton:
Partially addresses #27.

Plumb `DapAggregator::metrics()` to `VdafConfig` and count report rejections. Wherever the aggregator transitions a report to "failued", record the failure reason.

Relatedly:

* Refactor `VdafConfig::produce_report_with_extensions()` to allow us to construct invalid input shares in tests.

* Add a prefix to the metrics registered by `DaphneMetrics::register()`, allowing multiple instances of `DaphneMetrics` to be registered with the same registry. This simplifies the testing framework a bit.

* In-line variable names in a few `format!`  instances.

* Move `assert_metrics_include!` macro to the `testing` module so that we can reuse it in `vdaf/mod_test.rs`. Also, only compile it in tests.

* Rename some tests in `vdaf/mod_test.rs` to more closely correspond to the condition being tested.

### DIFF
--- a/daphne/src/vdaf/mod_test.rs
+++ b/daphne/src/vdaf/mod_test.rs
@@ -2,13 +2,15 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 use crate::{
-    async_test_version, async_test_versions,
+    assert_metrics_include, assert_metrics_include_auxiliary_function, async_test_version,
+    async_test_versions,
     hpke::HpkeReceiverConfig,
     messages::{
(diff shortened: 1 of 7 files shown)


daphne/src/vdaf/mod_test.rs
```diff
@@ -651,6 +744,85 @@ impl Test {
         reports
     }
 
+    // Tweak the Helper's share so that deocding succeeds but preparation fails.
```
bhalleycf, line 747: typo for "decoding" here

bhalleycf (approved):

PR #208, opened 2023-01-23
Add spans around major async calls from the router.

bhalleycf:
Add spans around major async calls (i.e. API entry points) from the URL router.  The span encompasses all work done by the API.

Also adjust the level of a few noisy messages.

### DIFF
--- a/daphne_worker/src/durable/garbage_collector.rs
+++ b/daphne_worker/src/durable/garbage_collector.rs
@@ -6,7 +6,7 @@ use crate::{
     durable::{DurableConnector, DurableOrdered, DurableReference},
     initialize_tracing, int_err,
 };
-use tracing::debug;
+use tracing::trace;
 use worker::*;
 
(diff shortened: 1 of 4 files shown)


daphne_worker/src/durable/garbage_collector.rs
```diff
@@ -81,9 +81,10 @@ impl DurableObject for GarbageCollector {
                             &(),
                         )
                         .await?;
-                    debug!(
+                    trace!(
                         "GarbageCollector: deleted {} instance {}",
```
cjpatton, line 85: I'm preeeetty sure this works but not 100%

```suggestion
                        "GarbageCollector: deleted {durable_ref.binding} instance {durable_ref.id_hex}",
```

cjpatton (approved):

daphne_worker/src/durable/garbage_collector.rs
bhalleycf: I tried a name with a dot in it earlier today and it was rejected because of the ".".  The doco says it must be an identifier, and I think that something like durable_ref.binding is actually an expression.

PR #213, opened 2023-01-25
Add span timing that works with WASM.

bhalleycf:
We also move emitting timestamps into the abstraction the library prefers instead of just adding them to the front of a log line.

Finally, some redundant info in some DO messages and total request timing is removed.

### DIFF
--- a/daphne_worker/src/durable/garbage_collector.rs
+++ b/daphne_worker/src/durable/garbage_collector.rs
@@ -6,7 +6,7 @@ use crate::{
     durable::{DurableConnector, DurableOrdered, DurableReference},
     initialize_tracing, int_err,
 };
-use tracing::trace;
+use tracing::{error, trace};
 use worker::*;
 
 pub(crate) const DURABLE_GARBAGE_COLLECTOR_PUT: &str = "/internal/do/garbage_collector/put";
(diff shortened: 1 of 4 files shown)


daphne_worker/src/durable/garbage_collector.rs
```diff
@@ -41,17 +41,16 @@ impl DurableObject for GarbageCollector {
                     | durable::BINDING_DAP_LEADER_COL_JOB_QUEUE
                     | durable::BINDING_DAP_HELPER_STATE_STORE => (),
                     s => {
-                        return Err(int_err(format!(
-                            "GarbageCollector: unrecognized binding: {}",
-                            s
-                        )))
+                        let message = format!("GarbageCollector: unrecognized binding: {}", s);
+                        error!("{}", message);
+                        return Err(int_err(message));
```
cjpatton, line 46: `int_err()` is kind of a hack meant to work around the fact that we didn't have very good logging. Now that we do, I wonder if it would be preferable to have simpler errors. One caveat is that Cloudflare has had issues with DO logs disappearing in the past.

daphne_worker/src/lib.rs
```diff
@@ -478,9 +487,11 @@ impl DaphneWorkerRouter {
             router
         };
 
-        let start = Date::now().as_millis();
+        // Note that we do not have a tracing span for the whole request because it typically
```
cjpatton, line 490: My editor highlights "NOTE" when capitalized. 
```suggestion
        // NOTE that we do not have a tracing span for the whole request because it typically
```

cjpatton (approved):

PR #214, opened 2023-01-26
Eliminate clippy warnings about directly using variables in format! strings

bhalleycf:
Clippy in rust 1.67.0 is picky about these by default now.  I added a few I found along the way in other format-like things, e.g. trace!.  Probably there are more of these to gradually improve.


cjpatton (approved):
Clippy you diabolical genius!

PR #225, opened 2023-02-04
daphne_worker: Enforce key separation for HPKE across DAP versions

### LINKED ISSUE #212: Daphne-Worker: Assets in KV should be scoped by DAP version
We store the Aggregato's HPKE receiver configs and task configs in KV. The latter is explicitly bound to a DAP version, but HPKE receiver configs are not. As a security measure, it's a good idea to enforce key separation draft versions.

This would require a deployment change to helper1, either by generating the data manually and uploading to KV or interrupting service briefly.

cjpatton:
Closes #212.

It's a good idea to ensure that secrets aren't reused across protocol versions. Luckily, most secrets used by Daphne-Worker, are keyed by task ID, which in turn is bound to a specific version. One exception are the HPKE receiver keys.

To ensure key spearation for HPKE receiver keys, add the DAP version to the KV names. That is, instead of

  hpke_receiver_config/{config_id}

the name is

  hpke_receiver_config/{version}/{config_id}

Relatedly:

* Check for unparsable version number in the list of KV keys for HPKE receiver configurations.

* change the string representation of `DapVersion::Unknown` to "unknown_version". This is to ensure that, if this string ends up in KV, it looks more obvious that something is wrong.

NOTE: This is a deployment-breaking change.

### DIFF
--- a/daphne_worker/src/config.rs
+++ b/daphne_worker/src/config.rs
@@ -279,7 +279,7 @@ pub(crate) struct DaphneWorkerState {
 
     /// Cached HPKE receiver config. This will be populated when Daphne-Worker obtains an HPKE
     /// receiver config for the first time from Cloudflare KV.
-    hpke_receiver_configs: Arc<RwLock<HashMap<u8, HpkeReceiverConfig>>>,
+    hpke_receiver_configs: Arc<RwLock<HashMap<HpkeReceiverKvKey, HpkeReceiverConfig>>>,
 
     /// Laeder bearer token per task.
     leader_bearer_tokens: Arc<RwLock<HashMap<Id, BearerToken>>>,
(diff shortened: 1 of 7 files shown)


daphne_worker/src/config.rs
```diff
@@ -818,3 +818,54 @@ pub(crate) enum DaphneWorkerDeployment {
...
 }
+
+#[derive(Clone, Eq, Hash, PartialEq)]
+pub(crate) struct HpkeReceiverKvKey {
+    pub(crate) version: DapVersion,
+    pub(crate) hpke_config_id: u8,
+}
+
+impl HpkeReceiverKvKey {
+    fn parse_from_name(name: &str) -> Option<Self> {
```
bhalleycf, line 829: This parser accepts invalid input, namely "trailing junk", e.g.

hpke_receiver_config/version/v03/config_id/123/this/is/junk

would be accepted because we don't do

if iter.next().is_some() {
    return None;
}

after parsing the id.  This doesn't actually break anything, but it's better to only accept valid.

bhalleycf (approved):

PR #226, opened 2023-02-04
daphne_worker: Push metrics to Prometheus server

### LINKED ISSUE #27: Add Prometheus metrics

cjpatton:
Closes #27.

### DIFF
--- a/daphne_worker/src/lib.rs
+++ b/daphne_worker/src/lib.rs
@@ -479,7 +479,7 @@ impl DaphneWorkerRouter {
 
         // NOTE that we do not have a tracing span for the whole request because it typically
         // reports the same times as the span covering the specific API entry point that the
-        // router creates.  If curious, you can add .instrument(info_span!("http")) just before
+        // router creates. If curious, you can add .instrument(info_span!("http")) just before
         // the await and see.
         let result = router.run(req, env).await;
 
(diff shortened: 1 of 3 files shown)


cjpatton:
This will be ready for review once I've had a chance to test in staging.

cjpatton:
Tested in staging 👍

daphne_worker/src/lib.rs
```diff
@@ -492,6 +492,13 @@ impl DaphneWorkerRouter {
             )])
             .inc();
 
+        // Push metrics to Prometheus metrics server, if configured.
```
bhalleycf, line 495: It would be cool if we could have some sort of "do this before you shut me down, or after n seconds" thing, as then we could amortize the cost of stats uploads across multiple invocations in the same isolate, much like we're caching config.  I haven't seen any API to do that though.

bhalleycf (approved):

PR #230, opened 2023-02-14
Only initialize state once per isolate instead of on every request.

bhalleycf:


cjpatton (commented):
I can't believe how short this PR is, given the win! One comment: Can you add a message to the commit explaining the expected side-effects in miniflare and, separately, Workers? In particular, under what conditions do we expect the benefit to be exploited; when the HTTP session is reused between requests?

cjpatton (approved):

Issue #235, opened 2023-02-15
Possible Prometheus double counting


bhalleycf:
The Prometheus stats are currently in the cached shared state, and currently a stats update is posted after every request.  This update does not zero the stats, so if the isolate is used for more than one request, we could count the prior requests again.  Given there is no way to ask the runtime to let us upload something before the isolate is shutdown, there's no advantage in having the stats in the shared state, so probably the easiest thing is to move them to the per-request state.

bhalleycf:
There was definitely double counting.  This was fixed by #236.

PR #236, opened 2023-02-16
Split DaphneWorkerState into a shared state and a per-request state.

bhalleycf:
This fixes the double counting in #235, and also creates a clear delineation between the shared and per-request state.  The per-request state consists of

      metrics
      HTTP client

### DIFF
--- a/daphne_worker/src/config.rs
+++ b/daphne_worker/src/config.rs
@@ -319,13 +319,12 @@ impl DaphneWorkerConfig {
     }
 }
 
-/// Daphne-Worker state, used for handling a DAP request. Includes long-lived configuration,
-/// metrics, cached responses from KV, etc.
-pub(crate) struct DaphneWorkerState {
+/// Daphne-Worker per-isolate state, which may be used by multiple requests. Includes long-lived configuration,
+/// cached responses from KV, etc.
+pub(crate) struct DaphneWorkerIsolateState {
     pub(crate) config: DaphneWorkerConfig,
 
-    /// HTTP client to use for making requests to the Helper. This is only used if Daphne-Worker is
-    /// configured as the DAP Helper.
+    /// HTTP client to use for making requests.
     pub(crate) client: reqwest_wasm::Client,
 
     /// Cached HPKE receiver config. This will be populated when Daphne-Worker obtains an HPKE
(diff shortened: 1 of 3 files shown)


daphne_worker/src/config.rs
```diff
@@ -319,15 +319,11 @@ impl DaphneWorkerConfig {
     }
 }
 
-/// Daphne-Worker state, used for handling a DAP request. Includes long-lived configuration,
-/// metrics, cached responses from KV, etc.
-pub(crate) struct DaphneWorkerState {
+/// Daphne-Worker shared state, which may be used by multiple requests. Includes long-lived configuration,
+/// cached responses from KV, etc.
+pub(crate) struct DaphneWorkerSharedState {
```
cjpatton, line 324: * Perhaps "Isolate" state instead of "Shared" state?
* Random thought here: If we do a "hotfix" of data stored in KV in a live deployment, we would also want to tear down any isolates that happen to be running so they can pick up the changes. If we do "wrangler publish", will this restart any isolates that are running?

daphne_worker/src/config.rs
```diff
@@ -340,6 +336,28 @@ pub(crate) struct DaphneWorkerState {
...
+        })
+    }
+}
+
+/// Daphne-Worker per-request state.
+pub(crate) struct DaphneWorkerRequestState<'srv> {
+    pub(crate) shared_state: &'srv DaphneWorkerSharedState,
+
+    /// HTTP client to use for making requests.
+    pub(crate) client: reqwest_wasm::Client,
```
cjpatton, line 360: Why not reuse this? It seems sensible to reuse any TLS/HTTP connections that we can.

cjpatton (commented):
This looks reasonable. Instead of "splitting" isolate and request state, I think I would have just  removed the metrics from the `DaphneWorkerState` and explicitly re-register them on each request. But I suppose there may be other "per-request" state blobs we'll need down the line?

daphne_worker/src/config.rs
bhalleycf: Yeah, it is safe to re-use after some research, so I restored it.

daphne_worker/src/config.rs
bhalleycf: I switched to "isolate".  I don't know if wrangler publisher will tear down old stuff, and even if it did I'm not sure it would be properly synchronized.

cjpatton (approved):

PR #237, opened 2023-02-23
Use serde-wasm-bindgen to convert JsValues.

bhalleycf:
Since workers-rs moves to serde-wasm-bindgen in 0.0.13, we probably should go the same way.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -559,6 +559,7 @@ dependencies = [
  "reqwest-wasm",
  "ring",
  "serde",
+ "serde-wasm-bindgen",
  "serde_json",
  "thiserror",
  "tracing",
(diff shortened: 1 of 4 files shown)


bhalleycf:
converting into a draft as I see more instances

cjpatton (approved):
👍 This is not yet exhaustive, right? We should file an issue to track the rest.

bhalleycf:
Re exhaustive, I think it depends!   I fixed all of the cases where there was a comment saying we were using the deprecated into_serde(), and there are no other instances of it.  We do make one wasm_bindgen::JsValue::from_str() call that maybe could be updated to use serde_wasm_bindgen.

PR #238, opened 2023-02-24
Ensure adding and draining the pending queue are race-free [#73].

### LINKED ISSUE #73: Janus interop: Leader hangs on collect requests (flaky)
@BranLwyd is working on Daphne <-> Janus interop tests. One test exercises a flaky bug in Daphne's Leader that causes collect requests to hang indefinitely:

To reproduce, checkout https://github.com/divviup/janus/tree/bran/revise-integration-test and run 
```
cargo test -p monolithic_integration_test daphne_janus -- --nocapture
```
Note that you may need to update `monolothic_integration_test/artificats/daphne_compiled`. This can be done by running `worker-build` in our `daphne_worker_test` directory and copying the contents of `build/worker`.

The bug is likely to happen after about 50 trials or so. You should see something like this:
```
thread 'daphne_janus' panicked at 'assertion failed: Instant::now() < collect_job_poll_timeout', monolithic_integration_test/tests/common/mod.rs:220:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Thu Aug 04 2022 12:26:53 GMT-0700 (Pacific Daylight Time) - [/internal/process], located at: (30.2713, -97.7426), within: Texas
DAP_ENV: Hostname override applied
agg job queue: []
```

Notes:
* We noticed that in cases where the bug occurs, not all of the reports get aggregated. We consistently see 45, but expect 46 in the batch.
* Something to investigatge is whether calls to DaphneWorkerConfig::durable_obejct are deadlocking. In fact a mutex here may not be appropriate.
* The bug may be an artifact of miniflare and might not manifest in the live deployment.

bhalleycf:
Closes #73.

fyi/ @branlwyd

### DIFF
--- a/daphne_worker/src/durable/reports_pending.rs
+++ b/daphne_worker/src/durable/reports_pending.rs
@@ -115,56 +115,18 @@ impl DurableObject for ReportsPending {
                     .size()
                     == 0;
 
-                // BUG(issue#73) There is a race condition between DURABLE_REPORTS_PENDING_GET_PENDING
-                // and DURABLE_REPORTS_PENDING_PUT_PENDING that could prevent a report from getting
-                // processed.
-                //
-                //   1. DURABLE_REPORTS_PENDING_PUT_PENDING writes a report to storage and checks if
-                //      "agg_job" is set. If not, it sets "agg_job" and enqueues the job.
-                //
-                //   2. DURABLE_REPORTS_PENDING_GET_PENDING drains a number of reports from storage,
-                //      then checks if storage is empty. If so, it deletes "agg_job" if it exists
-                //      and dequeues the job.
-                //
-                // Consider the following sequence of storage transactions (suppose "agg_job" is
-                // set):
-                //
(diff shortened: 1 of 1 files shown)


daphne_worker/src/durable/reports_pending.rs
```diff
@@ -115,56 +115,18 @@ impl DurableObject for ReportsPending {
...
-                //   processed. This way we can avoid storing "agg_job" here.
-                //
-                // In the meantime, to avoid flakyness in interop tests, supress the bug in
-                // development environments by never clearing the agg job queue if a workaround flag
-                // is set. Note that it is necessary to restart miniflare between test runs in order
-                // to reset the DO state.
-                if empty
-                    && self
-                        .env
-                        .var("DAP_ISSUE73_DISABLE_AGG_JOB_QUEUE_GARBAGE_COLLECTION")?
```
cjpatton, line 161: Can we also delete this variable from `daphne_worker_test/wrangler.toml`?

cjpatton (approved):

PR #243, opened 2023-03-10
daphne_worker: taskprov: Add support for TLS client authentication

cjpatton:
Dapphne-Worker authorizes DAP sender requests using a bearer token. Replace `BearerToken` with a more general authorization method, called `DaphneWorkerAuth`, that also permits TLS client authentication.

Limitations:

* For now, only the Helper supports TLS client auth; the Leader still expects a bearer token to be configured for the task.

* For now, TLS client auth is only enabled if the taskprov extension is configured. Enabling this feature for other tasks will require a bit plumbing.

### DIFF
--- a/daphne_worker/src/config.rs
+++ b/daphne_worker/src/config.rs
@@ -6,6 +6,7 @@
 //! Daphne-Worker configuration.
 
 use crate::{
+    auth::{DaphneWorkerAuth, DaphneWorkerAuthMethod},
     dap_err,
     durable::{
         durable_name_report_store, durable_name_task,
@@ -39,7 +40,7 @@ use std::{
     sync::{Arc, RwLock, RwLockReadGuard},
     time::Duration,
 };
-use tracing::{error, trace};
+use tracing::{debug, error, trace};
 use worker::{kv::KvStore, *};
 
 pub(crate) const KV_KEY_PREFIX_HPKE_RECEIVER_CONFIG: &str = "hpke_receiver_config";
@@ -56,11 +57,11 @@ pub(crate) struct TaskprovConfig {
     /// VDAF verify key init secret, used to generate the VDAF verification key for a taskprov task.
     pub(crate) vdaf_verify_key_init: [u8; 32],
 
-    /// Leader bearer token for all taskprov tasks
-    pub(crate) leader_bearer_token: BearerToken,
+    /// Leader, Helper: Method for authorizing Leader requests.
+    pub(crate) leader_auth: DaphneWorkerAuthMethod,
 
-    /// Collector bearer token for all taskprov tasks
-    pub(crate) collector_bearer_token: BearerToken,
+    /// Leader: Method for authorizing Collector requests.
+    pub(crate) collector_auth: Option<DaphneWorkerAuthMethod>,
 }
 
 /// Parameters required for pushing Prometheus metrics.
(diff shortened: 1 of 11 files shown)


daphne_worker/src/dap.rs
```diff
@@ -219,48 +220,124 @@ impl<'srv> BearerTokenProvider<'srv> for DaphneWorker<'srv> {
...
     'srv: 'req,
 {
     type WrappedDapTaskConfig = GuardedDapTaskConfig<'req>;
 
     async fn authorized(
         &self,
-        req: &DapRequest<BearerToken>,
+        req: &DapRequest<DaphneWorkerAuth>,
     ) -> std::result::Result<bool, DapError> {
-        self.bearer_token_authorized(req).await
```
cjpatton, line 263: I have not had a chance to test this, but will do so before we merge. We could also consider writing E2E tests for this (I believe wrangler dev can be configured to expect TLS client certs), but that would take some doing.

daphne_worker/src/config.rs
```diff
@@ -838,8 +856,23 @@ impl<'srv> DaphneWorker<'srv> {
...
         mut req: Request,
-    ) -> Result<DapRequest<BearerToken>> {
-        let sender_auth = req.headers().get("DAP-Auth-Token")?.map(BearerToken::from);
+    ) -> Result<DapRequest<DaphneWorkerAuth>> {
+        // Determine the authorization method used by the sender.
+        let bearer_token = req.headers().get("DAP-Auth-Token")?.map(BearerToken::from);
+        let tls_client_auth = req.cf().tls_client_auth();
+        let sender_auth = match (bearer_token, tls_client_auth) {
+            (Some(bearer_token), None) => Some(DaphneWorkerAuth::BearerToken(bearer_token)),
+            (None, Some(tls_client_auth)) => Some(DaphneWorkerAuth::CfTlsClientAuth {
```
cjpatton, line 865: We pull out the issuer and subject so that we can check they have the expected values. Arguably it's sufficient to just check that a certificate is present (i.e., `tls_client_auth.is_some()`) and let the zone configuration bear the burden of checking these names. My goal here is to add a bit of resilience in case the zone config gets clobbered somehow.

Note that the edge metal is expected to handle the rest of the details of certificate validation, including checking the chain and signature, enforcing lifetime, etc.

cjpatton: Thinking about this more: Perhaps it's best to just check that a certificate is present. One problem that might come up is changing, say, the subject name when rotating the certificate: We will have to synchronize that rotation with deploying the new change in the environment variable to the Worker. Would it be better to just let the zone config handle everything?

cjpatton: If we keep this feature, perhaps it would be best to make this a list of issuer/subject pairs that we trust.

cjpatton: As discussed offline: We have kept the expected issuer, but there is now a list of valid "subjects".

bhalleycf (approved):

cjpatton:
This change seems to trigger an odd bug I don't know how to account for yet: 
```
2023-03-11T02:58:52.501Z  INFO daphne_worker_test: [/v03/hpke_config], located at: (37.7703, -122.4407), within: California
[mf:err] Unhandled Promise Rejection: TypeError: Cannot read properties of undefined (reading 'length')
    at w (/Users/christopherpatton/github.com/cloudflare/daphne/daphne_worker_test/build/worker/shim.mjs:1:7418)
    at Re (/Users/christopherpatton/github.com/cloudflare/daphne/daphne_worker_test/build/worker/shim.mjs:2:11295)
    at null.<anonymous> (wasm://wasm/00b68e06:1:2001585)
    at null.<anonymous> (wasm://wasm/00b68e06:1:453079)
    at null.<anonymous> (wasm://wasm/00b68e06:1:850382)
    at null.<anonymous> (wasm://wasm/00b68e06:1:70070)
    at null.<anonymous> (wasm://wasm/00b68e06:1:1801600)
    at null.<anonymous> (wasm://wasm/00b68e06:1:1540166)
    at null.<anonymous> (wasm://wasm/00b68e06:1:2111898)
    at _t (/Users/christopherpatton/github.com/cloudflare/daphne/daphne_worker_test/build/worker/shim.mjs:2:237) 
```

Will need to investigate before proceeding.

bhalleycf:
This is a workers-rs bug that is fixed on the main branch of workers-rs but has not yet been released.  I haven't gone looking for an associated issue yet, but the broken code was fixed in commit 5280f8939c9ea4dc5c60b49f8e3321ad6d794557.

The problem is that there is code binding between Rust and Javascript with a typo in the name:

pub fn cert_subject_dn_rfc225(&self) -> String {
      self.inner.cert_subject_dn_rfc225()
}

This typo is continued in the inner wasm-bindgen'd thing:

#[wasm_bindgen(structural, method, getter, js_name=certSubjectDNRFC225, js_class = "tlsClientAuth")]
pub fn cert_subject_dn_rfc225(this: &TlsClientAuth) -> String;

This is a getter for an object field, but the correct field name is "certSubjectDNRFC2253" as you can see from the printed object below.  The getter fails to get and presumably returns undefined explaining the error we see.

Here's the object:

Some(TlsClientAuth { inner: TlsClientAuth { obj: Object { obj: JsValue(Object({"certIssuerDNLegacy":"","certIssuerSKI":"","certSubjectDNRFC2253":"","certSubjectDNLegacy":"","certFingerprintSHA256":"","certNotBefore":"","certSKI":"","certSerial":"","certIssuerDN":"","certVerified":"NONE","certNotAfter":"","certSubjectDN":"","certPresented":"0","certRevoked":"0","certIssuerSerial":"","certIssuerDNRFC2253":"","certFingerprintSHA1":""})) } } })

cjpatton:
Thanks, @bhalleycf for following up :)

cjpatton:
Squashed.

PR #245, opened 2023-03-15
daphne_worker: Clean up authorization logic

cjpatton:
Currently either TLS client auth or a bearer token can be used to
authorize requests for taskprov tasks, but not both. This is problematic
from a deployment perspective, as we're likely to need to have both
methods in use while we transition to the new authorization method (TLS
client auth). This change modifies the authorization logic in
Daphne-Worker to validation all authorization methods presented and
authorize the request if all validate.

In addition, currently only one issuer/subject pair is trusted for the
certificate. In practice we're likely to issue certificates from
multiple issuers. This change replaces the single issuer/subject pair
with a sequence of trusted pairs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,49 +83,58 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.2.6"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "342258dd14006105c2b75ab1bd7543a03bdf0cfc94383303ac212a04939dff6f"
+checksum = "0ca84f3628370c59db74ee214b3263d58f9aadd9b4fe7e711fd87dc452b7f163"
 dependencies = [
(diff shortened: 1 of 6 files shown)


daphne_worker/Cargo.toml
```diff
@@ -40,7 +40,9 @@ tracing-subscriber = {version = "0.3.16", features = ["env-filter"]}
 url = { version = "2.3.1", features = ["serde"] }
 serde_json = "1.0.94"
 serde-wasm-bindgen = "0.5.0"
-worker = "0.0.14"
+#worker = "0.0.14"
+# XXX https://github.com/cloudflare/workers-rs/pull/292
```
cjpatton, line 44: This has been merged and will be picked up in the next release.

cjpatton:
Additional requirement: Allow both auth methods to be used at the same time; if both are present, then check both.

bhalleycf (approved):

PR #250, opened 2023-03-21
Revert "Upgrade worker and serde-wasm-bindgen"

cjpatton:
Based on #253 (merge that first).

Downgrade worker 0.0.14 -> 0.0.12 and serde-wasm-bindgen -> 0.5.0 -> 0.4.5. In particular:

1. Revert commit 92903d8bdfdd121df588cffa1c9d44a95c2dde97

3. daphne_worker: Address an API-breaking change in worker

We observed bugs in workers-rs 0.0.14 that we could not reproduce locally (wrangler dev), but only in Workers. We have not yet determined the root cause, but it seemed to be associated with known bugs in 0.0.14.

The next version, 0.0.15, is out, but we will need to test it thoroughly before we move to it. See #251.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1920,9 +1920,9 @@ dependencies = [
 
 [[package]]
 name = "serde-wasm-bindgen"
-version = "0.5.0"
+version = "0.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


bhalleycf (changes requested):
This PR accidentally contains the changes in PR #253 too (dockerfile stuff)

cjpatton:
Yup, it was previously based on top of it, but I've now rebased on main.

bhalleycf (approved):

PR #260, opened 2023-04-04
wrangler --experimental-local support

bhalleycf:
This allows the wrangler "--experimental-local" option to work.  This local mode has increased fidelity to real workers, and will be what you get from "--local" in the future.

For now we ignore errors from get_alarm() in "dev" deployments as alarms are not implemented yet.

This patch also assigns unique ids to the KV namespaces for the leader and helper, ensuring that they don't mix.


daphne_worker_test/wrangler.toml
```diff
@@ -59,20 +59,20 @@ DAP_TRACING = "debug"
...
+    { name = "DAP_REPORTS_PENDING", class_name = "ReportsPending" },
+    { name = "DAP_REPORTS_PROCESSED", class_name = "ReportsProcessed" },
 ]
 
 
 [[env.leader.kv_namespaces]]
 binding = "DAP_CONFIG"
-id = "<ignored>"         # TODO(cjpatton) Figure out how to pick this.
-preview_id = "<ignored>" # TODO(cjpatton) Figure out how to pick this.
+id = "24c4dc92d5cf4680e508fe18eb8f0281"         # TODO(cjpatton) Figure out how to pick this.
```
cjpatton, line 74: Here and below:
```suggestion
id = "24c4dc92d5cf4680e508fe18eb8f0281"
```

daphne_worker/src/durable/mod.rs
```diff
@@ -140,8 +140,25 @@ macro_rules! ensure_garbage_collected {
...
+                }
+                Err(e) => {
+                    if !matches!(
+                        $object.config.deployment,
+                        crate::config::DaphneWorkerDeployment::Dev
+                    ) {
+                        // We only return an error if not in the "dev" deployment as
+                        // the experimental-local dev environment doesn't have
+                        // working get_alarm() and set_alarm() yet, so we want to
+                        // ignore errors in that case.
```
cjpatton, line 158: Should we at least log the error?
```rust
warn!("set_alaram() failed: {e}");
```

cjpatton (commented):
I think you mentioned in this in the meeting, but for the record: How did you pick the id and preview_id for the KV namespaces? Did you pick these at random or are these values "special" somehow? We should document how these were generated in the wrangler.toml.

daphne_worker_test/wrangler.toml
```diff
@@ -122,28 +136,35 @@ DAP_TRACING = "debug"
...
 ]
 
 
 [[env.helper.kv_namespaces]]
-  binding = "DAP_CONFIG"
-  id = "<ignored>"         # TODO(cjpatton) Figure out how to pick this.
-  preview_id = "<ignored>" # TODO(cjpatton) Figure out how to pick this.
+binding = "DAP_CONFIG"
+# See the comments in the leader section for the meaning of these
+# fields.  Note that a different name shoudl be used when
```
cjpatton, line 149:
```suggestion
# fields.  Note that a different name should be used when
```

cjpatton (approved):

Issue #275, opened 2023-04-12
Handle missing media type in draft04 collection job polling.


bhalleycf:
The #270 fix did not cover this after all, as I found out in some re-testing.  (I had left one of my fixes related to this issue in the code the first time, so I didn't notice that 270 wasn't sufficient.)

When doing collection job polling, draft04 does not have a media type, and we thus need to extract the collection job id directly rather than by using our resource field.

cjpatton:
Did you notice this because one of our tests, or perhaps one of @divergentdave's? As of #270 we are strictly enforcing the media type for many requests, it's possible we missed this one.

bhalleycf:
Yes, as I was assembling the PRs I was testing the updated code on divergentdave's tests to make sure I didn't break anything, and it broke with just 270.  I looked at what I had before and noticed that it had the one change in this PR in addition to 270.

bhalleycf:
Fixed by #275.

PR #277, opened 2023-04-12
Handle collection job polling without a media type [#275]

bhalleycf:


daphne_worker/src/lib.rs
```diff
@@ -368,13 +368,23 @@ impl DaphneWorkerRouter {
...
                             };
-                            let collect_job_id = match req.collection_job_id() {
-                                Ok(id) => id,
-                                Err(e) => return abort(e),
-                            };
+                            // We cannot check a resource here as the resource is set via
+                            // media type, and there is no media type when polling.
+                            //
+                            // We can unwrap() here as the parameter really must exist.
+                            let collect_job_id_text = ctx.param("collect_job_id").unwrap();
```
cjpatton, line 375:
```suggestion
                            let collect_job_id_base64url = ctx.param("collect_job_id").unwrap();
```

cjpatton (approved):
Oh this is kind of a bummer, as the protocol violates assumptions we made in worker_request_to_dap(). I think we should file an issue against the spec and try to get the media type added to this POST. (Similarly for POST to aggregation job.)

The fix seems reasonable, just one editorial suggestion.

Issue #272, opened 2023-04-12
Bounds check needed for DO key fetching


bhalleycf:
If a task configured by the task API has a max-reports value bigger than 128, we can try to fetch more than the 128-key fetch limit for the DO, causing an internal error.  We should put a bounds check on the value passed to the DO.

bhalleycf:
Also, to be clear the workers API does not say there is any possible error response from asking for too much.

PR #286, opened 2023-04-13
Impose a limit on the number of keys in DO list requests [#272].

bhalleycf:
We want to impose a limit on the number of keys in DO list requests to ensure user-specified configuration cannot cause us to run out of memory, as the workers API does not ensure a list request will not send you over memory quota.

### DIFF
--- a/daphne_worker/src/durable/mod.rs
+++ b/daphne_worker/src/durable/mod.rs
@@ -5,6 +5,7 @@ use crate::{int_err, now};
 use daphne::{messages::TaskId, DapBatchBucket, DapVersion};
 use rand::prelude::*;
 use serde::{Deserialize, Serialize};
+use std::cmp::min;
 use worker::*;
 
 pub(crate) const DURABLE_DELETE_ALL: &str = "/internal/do/delete_all";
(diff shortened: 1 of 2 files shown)


daphne_worker/src/durable/mod.rs
```diff
@@ -20,6 +21,20 @@ pub(crate) const BINDING_DAP_GARBAGE_COLLECTOR: &str = "DAP_GARBAGE_COLLECTOR";
...
+// The DO API does not say that there is any limit on the number of keys it is willing to return
+// other than overall DO/worker memory, which it warns you not to exceed. Imposing some sort of
+// limit is a good idea, lest we get DoS'd by a task configuration with a large value.
+//
+// Currently the value is set to 128, as the miniflare environment will fail with more than this.
+// This appears to be a miniflare bug and not part of the API.
+//
+// We have not been able to replicate failures with wrangler2 in local or experimental-local mode.
+//
+// TODO(bhalley) does this need to be configurable?
```
cjpatton, line 35: A constant is fine IMO: To raise the limit we'd have to re-deploy the code anyway (with a new wrangler.toml); either way we need to review a PR and get it approved.
```suggestion
```

cjpatton (approved):

PR #297, opened 2023-04-17
Add error reporting trait

xtuc:
This change adds the ability for a Daphne embedder to report internal errors (example to Sentry).

By default, we provide basic error reporting to stderr.

### DIFF
--- a/daphne_worker/src/config.rs
+++ b/daphne_worker/src/config.rs
@@ -14,6 +14,7 @@ use crate::{
         DurableConnector, BINDING_DAP_GARBAGE_COLLECTOR, BINDING_DAP_LEADER_BATCH_QUEUE,
         DURABLE_DELETE_ALL,
     },
+    error_reporting::ErrorReporter,
     int_err,
     metrics::DaphneWorkerMetrics,
     InternalTestAddTask, InternalTestEndpointForTask, InternalTestRole,
@@ -411,12 +412,16 @@ pub(crate) struct DaphneWorkerRequestState<'srv> {
     /// Hostname parsed from the HTTP request URL. Set to "unspecified-daphne-worker-hsot" if the
     /// hostname is not part of the URL.
     pub(crate) host: String,
+
+    /// Error reporting for Daphne internal errors.
+    pub(crate) error_reporter: &'srv dyn ErrorReporter,
 }
 
 impl<'srv> DaphneWorkerRequestState<'srv> {
     pub(crate) fn new(
         isolate_state: &'srv DaphneWorkerIsolateState,
         req: &Request,
+        error_reporter: &'srv dyn ErrorReporter,
     ) -> Result<Self> {
         let prometheus_registry = Registry::new();
         let metrics = DaphneWorkerMetrics::register(&prometheus_registry, None)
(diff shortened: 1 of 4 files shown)


daphne_worker/src/lib.rs
```diff
@@ -182,14 +183,26 @@ pub struct DaphneWorkerReportSelector {
 }
 
 /// HTTP request handler for Daphne-Worker.
-#[derive(Default)]
-pub struct DaphneWorkerRouter {
+pub struct DaphneWorkerRouter<'a> {
```
cjpatton, line 186: For consistency with lifetime used where this is called.
```suggestion
pub struct DaphneWorkerRouter<'srv> {
```

daphne_worker/src/error_reporting.rs
```diff
@@ -0,0 +1,21 @@
...
+pub trait ErrorReporter {
+    fn report_abort(&self, error: &DapAbort);
+}
+
+/// Default implementation of the error reporting trait. Displays error on stderr.
+pub(crate) struct StderrErrorReporter {}
+
+impl ErrorReporter for StderrErrorReporter {
+    fn report_abort(&self, error: &DapAbort) {
+        eprintln!("ABORT: {}", error);
```
cjpatton, line 19: Hrm, is this the default behavior we want? I don't think `npx wrangler dev` will actually print to stderr. Perhaps we should use `tracing::error!()` instead?

daphne_worker/src/lib.rs
```diff
@@ -198,7 +211,7 @@ pub const DEFAULT_RESPONSE_HTML: &str = "<body>Daphne-Worker</body>";
 
 static ISOLATE_STATE: OnceCell<DaphneWorkerIsolateState> = OnceCell::new();
 
-impl DaphneWorkerRouter {
+impl<'a> DaphneWorkerRouter<'a> {
```
cjpatton, line 214: Compiler seems fine with an anonymous lifetime here. If you think a named lifetime would be more useful/readible, I'd suggest `'srv` for consistency with how we've named lifetimes elsewhere (`'srv` for the process; `'req` for the request).
```suggestion
impl DaphneWorkerRouter<'_> {
```

daphne_worker/src/config.rs
```diff
@@ -486,6 +492,7 @@ impl<'srv> DaphneWorkerRequestState<'srv> {
 
     pub(crate) fn dap_abort_to_worker_response(&self, e: DapAbort) -> Result<Response> {
         let status = if matches!(e, DapAbort::Internal(..)) {
+            self.error_reporter.report_abort(&e);
```
cjpatton, line 495: Note we already log the error via `error!()` below, so this may end up being redundant, at least for the default error reporter. Maybe the default behavior should be a no-op?

cjpatton (commented):
Looks great overall! Just a few minor comments.

daphne_worker/src/lib.rs
cjpatton: Whoops, in fact we need an actual lifetime here.
```suggestion
impl<'srv> DaphneWorkerRouter<'srv> {
```

PR #304, opened 2023-04-19
Adds ability to log structured JSON

will118:
* Configured with env var `DAP_TRACING_FORMAT=json`
* Existing behavior is unchanged (i.e. free form text with ANSI colors)
* Moves logging of prometheus metrics to TRACE.


xtuc (approved):

daphne_worker/src/tracing_utils.rs
```diff
@@ -162,13 +162,35 @@ pub fn initialize_tracing(env: &Env) {
             Ok(var) => var.to_string(),
             Err(_) => "info".to_string(),
         };
-        registry()
-            .with(
-                fmt::layer()
+
+        let (ansi, json) = match env.var("DAP_TRACING_FORMAT") {
```
cjpatton, line 166: Instead of adding another environment variable, I would prefer using `DAP_DEPLOYMENT` to control this behavior:
`dev` -> console print
`prod` -> JSON

PR #308, opened 2023-04-24
Adds tracing subscriber layer for JSON -> Workers

will118:
With this change, logging under `DAP_DEPLOYMENT=prod` will flatten fields from spans into the log line, allowing us to add fields like `correlation_id` which can be helpful when debugging.

This commit also moves `tracing_utils.rs` to a new directory.


bhalleycf (approved):

daphne_worker/src/tracing_utils/workers_json_layer.rs
cjpatton, line 1: This file needs a copyright notice.

cjpatton (changes requested):
Looks good to me! One blocking comment: The new file needs a copyright notice.

cjpatton (approved):

xtuc (approved):

PR #310, opened 2023-04-28
make failures to report metrics non fatal

xtuc:
Sometimes our metrics endpoint is returning an error and propagating the error to the eyeball.

Since metrics aren't cleared, we can retry the next time. This change makes the metrics failure non fatal and makes it invisible to the eyeball.


will118:
This makes sense to me.

We should also consider pushing the prom metrics in a `wait_until`.

PR #313, opened 2023-05-02
daphne_worker: Retry requests to HelperStateStore

cjpatton:
Add a method, `with_retry()` on `DurableConnector` that, when invoked, implements a simple linear-backoff retry mechanism. Further, retry requests to HelperStateStore.

This change is intended to address transient errors in the Workers platform that we noticed during load testing. In particular, we noticed a non-negligible number of failures of requests to HelperStateStore, of which there is a fresh instance per aggregation job.

Retries are safe for HelperStateStore because requests to this object are idempotent. In general, whether retry is safe depends on the nature of how the request changes the object's state.

### DIFF
--- a/daphne_worker/src/durable/mod.rs
+++ b/daphne_worker/src/durable/mod.rs
@@ -5,7 +5,8 @@ use crate::{int_err, now};
 use daphne::{messages::TaskId, DapBatchBucket, DapVersion};
 use rand::prelude::*;
 use serde::{Deserialize, Serialize};
-use std::cmp::min;
+use std::{cmp::min, time::Duration};
+use tracing::warn;
 use worker::*;
 
 pub(crate) const DURABLE_DELETE_ALL: &str = "/internal/do/delete_all";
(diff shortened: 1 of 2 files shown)


daphne_worker/src/durable/mod.rs
```diff
@@ -94,42 +108,61 @@ impl<'a> DurableConnector<'a> {
...
+            )?,
+            _ => {
+                return Err(Error::RustError(format!(
+                    "durable_request: Unrecognized method: {method:?}",
+                )));
+            }
+        };
 
-    let mut resp = durable_stub.fetch_with_request(req).await?;
-    resp.json().await
```
cjpatton, line 132: Reviewer note: Just try the request a few times, waiting the same amount of time between each retry. This is pretty redumentary -- Ideally we'd want to do some sort of exponential scheme, like https://docs.rs/backoff/latest/backoff/index.html. However the backoff crate in particular doesn't seem to play well with miniflare (haven't tried in Workers). I think the issue is that it wants to sample the clock but the runtime doesn't have a callback configured.

daphne_worker/src/durable/mod.rs
```diff
@@ -35,14 +36,27 @@ const ERR_NO_VALUE: &str = "No such value in storage.";
 // TODO(bhalley) does this need to be configurable?
 const MAX_KEYS: usize = 128;
 
+const DEFAULT_RETRY_INTERVAL: Duration = Duration::from_millis(500);
+const DEFAULT_MAX_RETRIES: usize = 3;
```
cjpatton, line 40: Reviwer note: Eventually we'll want these parameters to be overwritable by the wrangler.toml. However for now I think we should err on the side of simplicity while we figure out what we actually want in a retry mechanism long-term.

daphne_worker/src/durable/mod.rs
```diff
@@ -35,14 +36,27 @@ const ERR_NO_VALUE: &str = "No such value in storage.";
 // TODO(bhalley) does this need to be configurable?
 const MAX_KEYS: usize = 128;
 
+const DEFAULT_RETRY_INTERVAL: Duration = Duration::from_millis(500);
```
xtuc, line 39: I would decrease it to 100ms. eyeball workers don't have much CPU / time to spare.

daphne_worker/src/durable/mod.rs
bhalleycf: The standard "no frills" exponential backoff would be to set "let mut retry_interval = DEFAULT_RETRY_INTERVAL" right after the "let mut attempt = 1" and then do retry_interval *= 2 after the attempt += 1.

xtuc (approved):
Left a minor comment but lgtm

bhalleycf (approved):

daphne_worker/src/durable/mod.rs
cjpatton: Yeah that's what I was planning initially, but then I figured it would be good to add jitter as well. We may need to re-implement this at some point, but not now!

daphne_worker/src/durable/mod.rs
cjpatton: Done.

PR #314, opened 2023-05-03
Replaces aggregation_job_gauge with aggregation_job_counter

will118:
Allows for easier aggregation of the metrics pushed by the worker.

### DIFF
--- a/daphne/src/metrics.rs
+++ b/daphne/src/metrics.rs
@@ -4,10 +4,7 @@
 //! Daphne metrics.
 
 use crate::DapError;
-use prometheus::{
-    register_int_counter_vec_with_registry, register_int_gauge_vec_with_registry, IntCounterVec,
-    IntGaugeVec, Registry,
-};
+use prometheus::{register_int_counter_vec_with_registry, IntCounterVec, Registry};
 
 pub struct DaphneMetrics {
     /// Inbound request metrics: Successful requests served, broken down by type.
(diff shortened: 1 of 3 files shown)


daphne/src/metrics.rs
```diff
@@ -120,3 +115,11 @@ pub enum DaphneRequestType {
     /// DAP collect request.
     Collect,
 }
+
+#[derive(Clone, Copy, Debug)]
+pub enum DaphneAggregationJobStatus {
+    /// AggregationJobInitReq
+    Started,
+    /// AggregationJobContinueReq
```
cjpatton, line 123: I would avoid leaking protocol details here, since they are subject to change over time and somewhat orthogonal to these variants (at least at a conceptual level).
```suggestion
    Started,
```

daphne/src/metrics.rs
```diff
@@ -94,18 +91,16 @@ impl ContextualizedDaphneMetrics<'_> {
             .inc_by(val);
     }
 
-    pub fn agg_job_inc(&self) {
-        self.metrics
-            .aggregation_job_gauge
-            .with_label_values(&[self.host])
-            .inc();
-    }
+    pub fn agg_job_inc(&self, status: DaphneAggregationJobStatus) {
```
cjpatton, line 94: This works, but to avoid defining a new type (`DaphneAggregationJobStatus`) we could split this into two methods, one for each status: `agg_job_started_inc()` and `agg_job_finished_inc()`. I guess the benefit of defining an enum is that we may eventually define more intermediate statuses?

PR #318, opened 2023-05-05
taskprov: Miscellaneous improvements

cjpatton:
Based on #317 (merge that first).

Just some minor things I noticed after some recent refactoring:

*    tasprov: Dedup some shared logic

    Both `MockAggregator` and `DaphneWorker` need to parse need to do the
    following:

    1. Resolve advertisement
    2. Decide to opt-in
    3. Store the task

    Step (1.) is backend-agnostic; steps (2.) and (3.) can be handled as
    callbacks. Refactor the code so that they are.

*    daphne: Add field to `DapTaskConfig` indicating if taskprov was used

    For backwards compatibility, JSON deserialization defaults to `false`.
    This means that any tasks configured via taskprov will be treated as
    non-taskprov. However this is currently a noop, since no protocol logic
    uses this flag.

### DIFF
--- a/daphne/src/lib.rs
+++ b/daphne/src/lib.rs
@@ -320,6 +320,10 @@ pub struct DapTaskConfig {
 
     /// The Collector's HPKE configuration for this task.
     pub collector_hpke_config: HpkeConfig,
+
+    /// If true, then the taskprov extension was used to configure this task.
+    #[serde(default)]
+    pub taskprov: bool,
 }
 
 impl DapTaskConfig {
--- a/daphne_worker/src/dap.rs
+++ b/daphne_worker/src/dap.rs
@@ -45,19 +45,19 @@ use crate::{
 };
 use async_trait::async_trait;
 use daphne::{
-    aborts::DapAbort,
     auth::{BearerToken, BearerTokenProvider},
     constants::DapMediaType,
     hpke::HpkeDecrypter,
     messages::{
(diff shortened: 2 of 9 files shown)


daphne/src/lib.rs
```diff
@@ -320,6 +320,10 @@ pub struct DapTaskConfig {
 
     /// The Collector's HPKE configuration for this task.
     pub collector_hpke_config: HpkeConfig,
+
+    /// If true, then the taskprov extension was used to configure this task.
+    #[serde(default)]
+    pub taskprov: bool,
```
cjpatton, line 326: @bhalleycf suggested doing this at some point IIRC.

daphne_worker/src/dap.rs
```diff
@@ -363,57 +377,18 @@ where
...
-            return Ok(());
-        }
-
-        let global_config = self.get_global_config();
-        if !global_config.allow_taskprov {
-            // TODO(bhalleycf) if DAP gets a generic denied error, we should use it here.
-            return Err(DapError::Abort(DapAbort::InvalidTask {
-                detail: "Taskprov extension is disabled.".to_string(),
-                task_id: task_id.clone(),
-            }));
```
cjpatton, line 389: I don't think aborting here is actually desirable. If taskprov is disabled then we just want to ensure we don't try to configure a task.

bhalleycf (approved):

PR #327, opened 2023-05-23
Add task id to log line wherever available

mendess:


cjpatton (commented):
Looks good to me. My one suggestion would be to deduplicate the code. My guess is we can wrap `info_span!()` with a function (or maybe a macro) that wraps the logic for extracting the task ID from the request.

PR #328, opened 2023-05-24
Add and fix e2e tests

mendess:
Fixed the e2e tests such that they pass and added them to CI, I'm not entirely sure the fixes make sense though, since I know the codebase that well yet, so I need a review of the code changes I made.

### DIFF
--- a/daphne_worker_test/tests/e2e.rs
+++ b/daphne_worker_test/tests/e2e.rs
@@ -26,7 +26,6 @@ use serde::Deserialize;
 use serde_json::json;
 use std::cmp::{max, min};
 use test_runner::{TestRunner, MIN_BATCH_SIZE, TIME_PRECISION};
-use url::Url;
 
 // Redefine async_test_version locally because we want a
 // cfg_attr as well.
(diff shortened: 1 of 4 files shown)


daphne_worker_test/tests/test_runner.rs
```diff
@@ -89,8 +89,8 @@ impl TestRunner {
             DapVersion::Draft04 => "v04",
             _ => panic!("unimplemented DapVersion"),
         };
-        let mut leader_url = Url::parse(&format!("http://leader:8787/{}/", version_path)).unwrap();
-        let mut helper_url = Url::parse(&format!("http://helper:8788/{}/", version_path)).unwrap();
+        let mut leader_url = Url::parse(&format!("http://leader:8080/{}/", version_path)).unwrap();
+        let mut helper_url = Url::parse(&format!("http://helper:8080/{}/", version_path)).unwrap();
```
cjpatton, line 93: In order to conveniently run tests locally, it's helpful to have the Leader and Helper on separate ports so that they can use the same address (localhost).

daphne_worker_test/tests/e2e.rs
```diff
@@ -1529,7 +1529,7 @@ async fn e2e_helper_admin_add_task(version: DapVersion) {
         "vdaf_verify_key": "y4e6alnJMQ0MZTvdJRJx5Q"
     });
 
-    let url = Url::parse("http://127.0.0.1:8788/task").unwrap();
+    let url = Url::parse("http://helper:8080/task").unwrap();
```
cjpatton, line 1532: For local testing, it would be necessary to add `helper` as an alias for `localhost`. Would you mind noting this in daphne_worker_test/README.md?

cjpatton (approved):
I verified that the e2e tests still pass on my machine.

cjpatton:
Hmm, that's odd. I just got an email about a failed build on main, but the tests pass before we merged. I re-ran it and will follow up.

Issue #267, opened 2023-04-07
Add support for Prio3SumVec


cjpatton:
The prio crate implements a VDAF that we'll need to support in the near future: https://docs.rs/prio/latest/prio/vdaf/prio3/type.Prio3SumVec.html. To start, add a variant to the [Prio3Config enum](https://github.com/cloudflare/daphne/blob/main/daphne/src/lib.rs#L829). Building and running test should give you an indication of what needs to be changed next. Eventually test vectors for Prio3SumVec will need to be added to https://github.com/cloudflare/daphne/blob/main/daphne/src/vdaf/prio3_test.rs.

cjpatton:
cc/ @dorryspears

Issue #271, opened 2023-04-12
Add support for Prio3Histogram to interop test API


bhalleycf:
Although the larger code base supports histogram VDAFs, the configuration parser for adding tasks does not.

PR #331, opened 2023-05-30
Add support for Prio3SumVec

tholop:
This PR should solve https://github.com/cloudflare/daphne/issues/267 by adding support for Prio3SumVec. I added a test vector to check that the shares are prepared properly.

@cjpatton I left a couple of comments with `TODO(tholop)` with my questions, I'm still exploring the codebase so I might be misunderstanding some things

### DIFF
--- a/daphne/src/lib.rs
+++ b/daphne/src/lib.rs
@@ -483,6 +483,7 @@ impl AsRef<DapTaskConfig> for DapTaskConfig {
 pub enum DapMeasurement {
     U64(u64),
     U32Vec(Vec<u32>),
+    U128Vec(Vec<u128>),
 }
 
 /// The aggregate result computed by the Collector.
@@ -730,6 +731,10 @@ pub enum Prio3Config {
     /// The sum of 64-bit, unsigned integers. Each measurement is an integer in range `[0,
     /// 2^bits)`.
     Sum { bits: usize },
+
+    /// The element-wise sum of vectors. Each vector has `len` elements.
+    /// Each element is a 64-bit unsigned integer in range `[0,2^bits)`.
+    SumVec { bits: usize, len: usize },
 }
 
 /// DAP sender role.
--- a/daphne/src/vdaf/prio3.rs
+++ b/daphne/src/vdaf/prio3.rs
@@ -60,6 +60,10 @@ pub(crate) fn prio3_shard(
             let vdaf = Prio3::new_sum(2, *bits)?;
             Ok(shard!(vdaf, &(measurement as u128), nonce))
         }
+        (Prio3Config::SumVec { bits, len }, DapMeasurement::U128Vec(measurement)) => {
+            let vdaf = Prio3::new_sum_vec(2, *bits, *len)?;
+            Ok(shard!(vdaf, &measurement, nonce))
+        }
         _ => panic!("prio3_shard: unexpected VDAF config"),
     }
 }
@@ -147,6 +151,21 @@ pub(crate) fn prio3_prepare_init(
                 VdafMessage::Prio3ShareField128(share),
             ))
         }
+        Prio3Config::SumVec { bits, len } => {
+            let vdaf = Prio3::new_sum_vec(2, *bits, *len)?;
+            let (state, share) = prep_init!(
+                vdaf,
+                verify_key,
+                agg_id,
+                nonce,
+                public_share_data,
+                input_share_data
+            );
+            Ok((
+                VdafState::Prio3Field128(state),
+                VdafMessage::Prio3ShareField128(share),
+            ))
+        }
     }
 }
 
@@ -213,6 +232,16 @@ pub(crate) fn prio3_leader_prepare_finish(
             let agg_share = VdafAggregateShare::Field128(vdaf.aggregate(&(), [out_share])?);
             (agg_share, outbound)
         }
+        (
+            Prio3Config::SumVec { bits, len },
+            VdafState::Prio3Field128(state),
+            VdafMessage::Prio3ShareField128(share),
+        ) => {
+            let vdaf = Prio3::new_sum_vec(2, *bits, *len)?;
+            let (out_share, outbound) = leader_prep_fin!(vdaf, state, share, helper_share_data);
+            let agg_share = VdafAggregateShare::Field128(vdaf.aggregate(&(), [out_share])?);
+            (agg_share, outbound)
+        }
         _ => panic!("prio3_leader_prepare_finish: {ERR_FIELD_TYPE}"),
     };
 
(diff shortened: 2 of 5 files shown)


daphne/src/vdaf/prio3.rs
```diff
@@ -147,6 +153,34 @@ pub(crate) fn prio3_prepare_init(
                 VdafMessage::Prio3ShareField128(share),
             ))
         }
+        Prio3Config::SumVec { bits, len } => {
+            let vdaf: Prio3<
```
cjpatton, line 157: Is this type hint needed? I suspect (and hope!) not :)

daphne/src/vdaf/prio3.rs
```diff
@@ -147,6 +153,34 @@ pub(crate) fn prio3_prepare_init(
...
+                        prio::field::Field128,
+                        prio::flp::gadgets::BlindPolyEval<prio::field::Field128>,
+                    >,
+                >,
+                prio::vdaf::prg::PrgSha3,
+                16,
+            > = Prio3::new_sum_vec(2, *bits, *len)?;
+            // TODO(tholop): we don't need vec of Prio3Field128, right?
+            // Since the Prio3PrepareState<F, SEED_SIZE> will contain a Vec<F>
+            // https://github.com/divviup/libprio-rs/blob/3615c6d92dea9c5d4ab8faa9700970f685729fe1/src/vdaf/prio3.rs#L908
```
cjpatton, line 170: I believe we already have a variant of `VdafState` that we'll need: https://github.com/cloudflare/daphne/blob/main/daphne/src/vdaf/mod.rs#L80

Note that libprio's `Prio3PrepareState` is not specific to a `Type`.

daphne/src/vdaf/prio3_test.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) 2022 Cloudflare, Inc. All rights reserved.
 // SPDX-License-Identifier: BSD-3-Clause
 
+use std::vec;
```
cjpatton, line 4: I'm pretty sure you don't need to `use` this explicitly.

daphne/src/vdaf/prio3_test.rs
```diff
@@ -46,6 +48,16 @@ fn prepare_histogram() {
     .unwrap();
 }
 
+#[test]
+fn prepare_sum_vec() {
+    test_prepare(
+        &Prio3Config::SumVec { bits: 23, len: 3 },
+        DapMeasurement::U64Vec(vec![1, 0, 42]),
```
cjpatton, line 55: I'd add `2^23 - 1` as an additional use case here.

daphne_worker/src/config.rs
```diff
@@ -837,6 +837,7 @@ impl<'srv> DaphneWorker<'srv> {
                 let bits = bits.parse().map_err(int_err)?;
                 VdafConfig::Prio3(Prio3Config::Sum { bits })
             }
+            // TODO(tholop): should we add SumVec too? Why is Histogram not here?
```
cjpatton, line 840: Yes, please add it! Histogram should be here, too, but we'll add it in a PR for https://github.com/cloudflare/daphne/issues/271.

tholop: I added SumVec there, but it required me to extend `InternalTestVdaf`. I don't know how to run the interop tests, so not sure if it breaks anything

daphne/src/vdaf/prio3_test.rs
tholop: I added that as an extra test, and also added U128 as a separate test

daphne/src/vdaf/prio3.rs
```diff
@@ -60,6 +60,12 @@ pub(crate) fn prio3_shard(
             let vdaf = Prio3::new_sum(2, *bits)?;
             Ok(shard!(vdaf, &(measurement as u128), nonce))
         }
+        (Prio3Config::SumVec { bits, len }, DapMeasurement::U64Vec(measurement)) => {
+            let vdaf = Prio3::new_sum_vec(2, *bits, *len)?;
+            // TODO(tholop): shard expects a u128. Is it ok to allocate a new vector to cast each element?
```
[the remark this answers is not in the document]
tholop: I added a new `DapMeasurement` variant, for which we don't need this allocation. But do we still want to keep the U64 type? I used it originally because that's what `Count` and `Sum` are using (and they cast to u128).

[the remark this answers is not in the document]
cjpatton: Hmm, I don't think `Count` casts to `u128`: https://github.com/cloudflare/daphne/pull/331/files#diff-6660c5f7b70da1bf7dd0a22e00b5ee3191e693b9cef73a0fb5aeb31ae514f28cR53

Good point about `Sum`, however! (Also true for `Histogram`.) I don't remember the original motivation for this (or if this was well thought out), but I think it makes sense to use the measurement type that's native for the `Type`, i.e.:
* `Count` should be `u64`
* `Histogram`, `Sum` should be `u128`
* `SumVec` should be `Vec<u128` (in particular we should not define a `U64Vec` variant)

However to keep this change minimal, I'd suggest we create another PR to fix `Count` and `Histogram`. (Alternatively, we could file an issue and address it later.)

daphne/src/vdaf/prio3_test.rs
```diff
@@ -46,6 +46,30 @@ fn prepare_histogram() {
...
+        DapAggregateResult::U128Vec(vec![1, 0, 42]),
+    )
+    .unwrap();
+
+    test_prepare(
+        &Prio3Config::SumVec { bits: 23, len: 1 },
+        DapMeasurement::U64Vec(vec![1 << 23 - 1]),
```
cjpatton, line 60: I believe this evaluates to `1<<22`: https://doc.rust-lang.org/reference/expressions.html#expression-precedence
```suggestion
        DapMeasurement::U64Vec(vec![(1 << 23) - 1]),
```

daphne_worker/src/lib.rs
```diff
@@ -698,6 +698,7 @@ pub(crate) struct InternalTestVdaf {
     typ: String,
     #[serde(skip_serializing_if = "Option::is_none")]
     bits: Option<String>,
```
cjpatton, line 700: Otherwise this field is mandatory.

```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    bits: Option<String>,
```

cjpatton (commented):
A couple more comments -- overall this looks great. I think the test coverage is fine, as it at least matches

tholop:
Thanks @cjpatton, I pushed fixes for your comments and I'll open a new PR soon for the `Histogram` and `Sum` (not `Count` indeed) types.

daphne/src/lib.rs
```diff
@@ -483,6 +483,8 @@ impl AsRef<DapTaskConfig> for DapTaskConfig {
 pub enum DapMeasurement {
     U64(u64),
     U32Vec(Vec<u32>),
+    U64Vec(Vec<u64>),
```
cjpatton, line 486: this variant is no longer needed, right?

daphne_worker/src/config.rs
cjpatton: We don't have interop tests running at the moment. If you'd like to add tests, we could add unit tests for decoding a JSON blob with the config we're expecting.

cjpatton (commented):
LGTM, just a couple of minor things. Before merging, can you squash the commits into one?

daphne_worker/src/config.rs
tholop: I added a new issue for that (https://github.com/cloudflare/daphne/issues/333), it might overlap with https://github.com/cloudflare/daphne/issues/271.

cjpatton (approved):

PR #334, opened 2023-06-02
Adds new AuditLog trait to daphne

will118:
Enables implementations backed by services like: https://developers.cloudflare.com/analytics/analytics-engine/

* For now, there is just a basic handler for successful aggregation job requests.
* Adds a `Display` to `VdafConfig` for a reasonably succinct string repr, useful for event logging.
* Adds some assertions on the handler being invoked

### DIFF
--- a/daphne/src/audit_log.rs
+++ b/daphne/src/audit_log.rs
@@ -1,0 +1,35 @@
+// Copyright (c) 2023 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+use crate::{messages::TaskId, DapTaskConfig};
+
+pub enum AggregationJobAuditAction {
+    Init,
+    Continue,
+}
+
+pub trait AuditLog {
+    fn on_aggregation_job(
+        &self,
+        host: &str,
+        task_id: &TaskId,
+        task_config: &DapTaskConfig,
+        report_count: u64,
+        action: AggregationJobAuditAction,
+    );
+}
+
+/// Default implementation of the trait, which is a no-op.
+pub struct NoopAuditLog;
+
+impl AuditLog for NoopAuditLog {
+    fn on_aggregation_job(
+        &self,
+        _host: &str,
+        _task_id: &TaskId,
+        _task_config: &DapTaskConfig,
+        _report_count: u64,
+        _action: AggregationJobAuditAction,
+    ) {
+    }
+}
--- a/daphne/src/lib.rs
+++ b/daphne/src/lib.rs
@@ -62,7 +62,7 @@ use std::{
     borrow::Cow,
     cmp::{max, min},
     collections::{HashMap, HashSet},
-    fmt::Debug,
+    fmt::{Debug, Display},
 };
 use url::Url;
 
@@ -702,7 +702,7 @@ pub enum DapHelperTransition<M: Debug> {
     Finish(Vec<DapOutputShare>, M),
 }
 
-/// Specificaiton of a concrete VDAF.
+/// Specification of a concrete VDAF.
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 #[serde(rename_all = "snake_case")]
 pub enum VdafConfig {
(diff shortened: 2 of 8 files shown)


mendess:
Isn't it possible to leverage tracing somehow to achieve this? Adding this effectively means we have to visibility mechanisms

daphne_worker/Cargo.toml
```diff
@@ -40,7 +40,7 @@ tracing-subscriber = {version = "0.3.17", features = ["env-filter", "json"]}
 url = { version = "2.3.1", features = ["serde"] }
 serde_json = "1.0.96"
 serde-wasm-bindgen = "0.5.0"
-worker = "0.0.16"
+worker = "0.0.17"
```
cjpatton, line 43: Is bumping the `worker` crate necessary in order to use new features? If not, then I'd prefer to bump the version in a separate PR. The reason is that in the past we've seen bugs in `worker` releases that only manifested in the real Workers runtime (not when run locally via `wrangler dev`). Ideally we would test updates to `worker` in a live deployment before merging.

daphne/src/audit_log.rs
cjpatton, line 1: New files must have copyright notices.

daphne/src/audit_log.rs
```diff
@@ -0,0 +1,42 @@
+use crate::{messages::TaskId, DapTaskConfig};
+
+pub trait AuditLog {
+    fn on_aggregation_job_init_req(
```
cjpatton, line 4: Why do we need a separate callback for agg init and and continue? How about a single callback for aggregate requests?

daphne/src/audit_log.rs
```diff
@@ -0,0 +1,42 @@
+use crate::{messages::TaskId, DapTaskConfig};
+
```
cjpatton, line 2: Can you add a docucomment briefly describing what this does? IIUC the implementation of the trait is in the internal repo for now, but it would be good to at least to say what this does and provide a link to internal documentation.

daphne/src/lib.rs
```diff
@@ -716,6 +716,15 @@ impl std::str::FromStr for VdafConfig {
     }
 }
 
+impl Display for VdafConfig {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            VdafConfig::Prio3(conf) => write!(f, "Prio3({conf})"),
```
cjpatton, line 722: nit: For consistency with match on `VdafConfig` elsewhere 
```suggestion
            VdafConfig::Prio3(prio3_config) => write!(f, "Prio3({conf})"),
```

daphne/src/roles.rs
```diff
@@ -1028,6 +1039,13 @@ where
                     }
                 };
 
+                self.audit_log().on_aggregation_job_continue_req(
+                    req.host(),
+                    task_id,
+                    task_config,
+                    out_shares_count,
+                );
```
cjpatton, line 1047: Without knowing what we're doing with this information, I don't quite see the value of having two callbacks, one for agg init and one for agg cont. It seems like the same information is being reported.

daphne_worker/Cargo.toml
will118: we can do it in a separate PR :+1:

daphne/src/audit_log.rs
will118: we don't, I can update it to take a boolean if that is preferred.

daphne/src/audit_log.rs
will118: yep, I can do so after we merge this into the fork

daphne/src/audit_log.rs
mendess: I'd rather keep boolean parameters to a minimum, they make code hard to read. You can instead define a simple two variant enum, same performance, much better readability
```rs
enum AuditAction {
    Init,
    Continue,
}
```

mendess (approved):
Approving to unblock this, we can iterate on using tracing for this at a latter date

PR #335, opened 2023-06-02
Simplify trait lifetimes with GATs

mendess:
The main traits of the protocol were overly complex in terms of lifetimes. Rust
recently added GATs which are perfect for expressing what we need here.

### DIFF
--- a/daphne/src/auth.rs
+++ b/daphne/src/auth.rs
@@ -51,21 +51,23 @@ impl AsRef<BearerToken> for BearerToken {
 
 /// A source of bearer tokens used for authorizing DAP requests.
 #[async_trait(?Send)]
-pub trait BearerTokenProvider<'a> {
+pub trait BearerTokenProvider {
     /// A reference to a bearer token owned by the provider.
-    type WrappedBearerToken: AsRef<BearerToken>;
+    type WrappedBearerToken<'a>: AsRef<BearerToken>
+    where
+        Self: 'a;
 
     /// Fetch the Leader's bearer token for the given task, if the task is recognized.
-    async fn get_leader_bearer_token_for(
-        &'a self,
-        task_id: &'a TaskId,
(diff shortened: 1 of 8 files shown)


daphne/src/auth.rs
```diff
@@ -51,21 +51,23 @@ impl AsRef<BearerToken> for BearerToken {
...
 
     /// Fetch the Collector's bearer token for the given task, if the task is recognized.
-    async fn get_collector_bearer_token_for(
-        &'a self,
-        task_id: &'a TaskId,
-    ) -> Result<Option<Self::WrappedBearerToken>, DapError>;
+    async fn get_collector_bearer_token_for<'s>(
+        &'s self,
+        task_id: &'s TaskId,
+    ) -> Result<Option<Self::WrappedBearerToken<'s>>, DapError>;
```
cjpatton, line 70: What naming convention are you using for the lifetimes? (`'i` here and `'s` above.) Our current conventions (perhaps could be improved):
* If the thing must live as long as the request, then `'req`
* Else if the thing must live as long as the server, then `'srv`
* Else `'a`, `'b`, and so on.

cjpatton (commented):
This is awesome! One minor comment about your lifetime naming conventions, as it clashes with what we currently do. (I don't have a strong opinion about what we do, I'd just like to be consistent if possible.)

daphne/src/auth.rs
mendess: That makes sense. My naming pattern for this PR was `'s` for `self` and `'i` for `TaskId` (`'i` of `Id`).

In this particular method, it's a bit awkward because it must live as long as `TaskId` and as long as `self`, so I just picked self. Which is inconsistent with the method directly above it :sweat_smile:.

Also a note on picking `'a` for the associated type lifetime, I like to pick a name that doesn't clash with any surrounding name to avoid any possible confusion of intention

PR #336, opened 2023-06-02
Reduce version specification duplication by using workspace inheritance

mendess:
Having the same version written two or three times is a binary size hazard and
just a tech debt pitfall in general. Using workspace inheritance we can specify
versions only once and reuse them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,3 +12,27 @@ members = [
 [profile.release]
 # Tell `rustc` to optimize for small code size.
 opt-level = "s"
+
+[workspace.dependencies]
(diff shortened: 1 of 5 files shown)


cjpatton (approved):
Awesome - 🦆 . Quick question: How does having multiple crates share dependencies result in a codesize blow up? I figured if they are using the same version then they'd be able to use the same code. Is that not true?

mendess:
@cjpatton if the version matches 100% then cargo will not compile the crate twice, but if for some reason it deviates then the dependency is compiled twice. Currently this was not the case for any dependency but this way it we are sure it's never the case

PR #338, opened 2023-06-05
Removed unused dependencies

mendess:
Remove unused dependencies and setup a lint to catch when this happens.

This lint is not perfect, in slightly more complex build configurations it
signals a lot of false positives (which is the case for the `daphne_worker_test`
crate).

https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unused-crate-dependencies

> This lint is "allow" by default because it can provide false positives
> depending on how the build system is configured. For example, when using
> Cargo, a "package" consists of multiple crates (such as a library and a
> binary), but the dependencies are defined for the package as a whole. If there
> is a dependency that is only used in the binary, but not the library, then the
> lint will be incorrectly issued in the library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -492,14 +492,11 @@ name = "dapf"
 version = "0.3.0"
 dependencies = [
  "anyhow",
- "assert_matches",
- "base64 0.21.2",
  "clap",
  "daphne",
  "prio",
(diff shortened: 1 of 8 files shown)


cjpatton (approved):
❤️

Issue #315, opened 2023-05-03
New cloning warning with --experimental-local


bhalleycf:
I now see a lot of this message (below) from the experimental-local server when running tests.  It happens on both leader and helper.  I'm not sure if this reflects some new capability in the experimental local server or some local change, but I suspect the latter as AFAIK my wrangler config and miniflare tre haven't changed.

`Your worker called response.clone(), but did not read the body of both clones. This is wasteful, as it forces the system to buffer the entire response body in memory, rather than streaming it through. This may cause your worker to be unexpectedly terminated for going over the memory limit. If you only meant to copy the response headers and metadata (e.g. in order to be able to modify them), use `new Response(response.body, response)` instead.`

bhalleycf:
This message comes from workerd, so it's something cloning at Javascript level, presumably something in workers-rs.  I don't think that rules out us initiating the clone, though I haven't been able to find one.  I do see some cloning in workers-rs.

cjpatton:
@bhalleycf we should see if these warnings still show up with more recent tooling (wrangler 3 or workers-rs 0.0.17).

bhalleycf:
Checking the workerd source, I see  that the message has been updated to mention requests too, as apparently while it said "response" it meant "request or response".  I will now check for Request cloning.  _grumble_.

bhalleycf:
Aha!  The warning is triggered by durable_request(), as it creates the request and then clones it every time it tries an attempt in the attempt loop.  This means that in all cases the original req is never read if it has a body, and that triggers the warning over and over as there is no limiting in workerd for it.  I tested this theory by changing the code to always make a new request rather than cloning, and that made everything quiet.  I can submit a PR...

PR #342, opened 2023-06-05
Do not clone request objects when making DO requests [#315].

bhalleycf:
Our prior technique made a request and then cloned it on each attempt, never using the original.  This triggered a warning from workerd on every DO request, as cloning prevents streaming of the body.  (This wasn't important in our case as the body is a small piece of JSON we just made, but workerd does not know that.)  We now just make a new request object for each attempt.


cjpatton (approved):
Simple enough, thanks Bob!

Issue #326, opened 2023-05-22
Add problem details to unrecognizedMessage abort


cjpatton:
Recent changes to the spec are coalescing various error types into a generic "invalid message" error. When we abort, we should provide more context so that the reason the message was deemed invalid is clear.

* https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/465
* https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/464

cjpatton:
First step: Add a details string to `UnrecgonziedMessage` variant and add details where appropriate. Also, consider plumbing in the task ID as well.

cjpatton:
Assigning to @tholop.

PR #343, opened 2023-06-05
Add problem details to UnrecognizedMessage abort

tholop:
As described in https://github.com/cloudflare/daphne/issues/326, the `UnrecognizedMessage` abort type was too generic. This PR adds a `detail` field that contains more information, depending on where the error was raised. I looked for conditions raising the `UnrecognizedMessage` abort, and translated them into hopefully helpful strings. I also added a `task_id` field that optionally contains the task ID, when available.

### DIFF
--- a/daphne/src/aborts.rs
+++ b/daphne/src/aborts.rs
@@ -93,7 +93,10 @@ pub enum DapAbort {
 
     /// Unrecognized message. Sent in response to a malformed or unexpected message.
     #[error("unrecognizedMessage")]
-    UnrecognizedMessage,
+    UnrecognizedMessage {
+        detail: String,
+        task_id: Option<TaskId>,
+    },
 
     /// Unrecognized DAP task. Sent in response to a request indicating an unrecognized task ID.
     #[error("unrecognizedTask")]
@@ -134,9 +137,8 @@ impl DapAbort {
                 Some("The request indicates an aggregation job that does not exist.".into()),
                 Some(agg_job_id_base64url),
             ),
-            Self::ReportTooLate | Self::UnrecognizedMessage | Self::UnrecognizedTask => {
-                (None, None, None)
-            }
+            Self::UnrecognizedMessage { detail, task_id } => (task_id, Some(detail), None),
+            Self::ReportTooLate | Self::UnrecognizedTask => (None, None, None),
             Self::Internal(e) => (None, Some(e.to_string()), None),
         };
 
@@ -252,8 +254,8 @@ impl DapAbort {
             Self::UnrecognizedAggregationJob { .. } => {
                 ("Unrecognized aggregation job", Some(self.to_string()))
             }
-            Self::UnrecognizedMessage => {
-                ("Failed to parse the request body", Some(self.to_string()))
+            Self::UnrecognizedMessage { .. } => {
+                ("Malformed or invalid message", Some(self.to_string()))
             }
             Self::UnrecognizedTask => (
                 "Task indicated by request is not recognized",
@@ -281,8 +283,11 @@ impl From<DapError> for DapAbort {
 }
 
 impl From<CodecError> for DapAbort {
-    fn from(_e: CodecError) -> Self {
-        Self::UnrecognizedMessage
+    fn from(e: CodecError) -> Self {
+        Self::UnrecognizedMessage {
+            detail: format!("codec error: {e}"),
+            task_id: None,
+        }
     }
 }
 
--- a/daphne/src/roles.rs
+++ b/daphne/src/roles.rs
@@ -334,7 +334,13 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
 
         if report.encrypted_input_shares.len() != 2 {
             // TODO spec: Decide if this behavior should be specified.
-            return Err(DapAbort::UnrecognizedMessage);
+            return Err(DapAbort::UnrecognizedMessage {
+                detail: format!(
+                    "expected exactly two encrypted input shares; got {}",
+                    report.encrypted_input_shares.len()
+                ),
+                task_id: Some(task_id.clone()),
+            });
         }
 
         // Check that the indicated HpkeConfig is present.
@@ -822,7 +828,11 @@ pub trait DapHelper<S>: DapAggregator<S> {
                             .map(|report_share| &report_share.report_metadata);
                     } else if using_taskprov != 0 {
                         // It's not all taskprov or no taskprov, so it's an error.
-                        return Err(DapAbort::UnrecognizedMessage);
+                        return Err(DapAbort::UnrecognizedMessage {
+                            detail: "some reports include the taskprov extensions and some do not"
+                                .to_string(),
+                            task_id: Some(task_id.clone()),
+                        });
                     }
                 }
                 resolve_taskprov(self, task_id, req, first_metadata).await?;
@@ -1166,10 +1176,13 @@ fn check_part_batch(
         ));
     }
 
-    // Check that the aggreation parameter is suitable for the given VDAF.
+    // Check that the aggregation parameter is suitable for the given VDAF.
     if !task_config.vdaf.is_valid_agg_param(agg_param) {
         // TODO spec: Define this behavior.
-        return Err(DapAbort::UnrecognizedMessage);
+        return Err(DapAbort::UnrecognizedMessage {
+            detail: "invalid aggregation parameter".into(),
+            task_id: Some(task_id.clone()),
+        });
     }
 
     Ok(())
@@ -1186,10 +1199,13 @@ async fn check_batch<S>(
     let global_config = agg.get_global_config();
     let batch_overlapping = agg.is_batch_overlapping(task_id, batch_sel);
 
-    // Check that the aggreation parameter is suitable for the given VDAF.
+    // Check that the aggregation parameter is suitable for the given VDAF.
     if !task_config.vdaf.is_valid_agg_param(agg_param) {
         // TODO spec: Define this behavior.
-        return Err(DapAbort::UnrecognizedMessage);
+        return Err(DapAbort::UnrecognizedMessage {
+            detail: "invalid aggregation parameter".into(),
+            task_id: Some(task_id.clone()),
+        });
     }
 
     // Check that the batch boundaries are valid.
--- a/daphne/src/roles_test.rs
+++ b/daphne/src/roles_test.rs
@@ -690,7 +690,7 @@ async fn handle_agg_job_req_zero_round(version: DapVersion) {
         .await;
     assert_matches!(
         t.helper.handle_agg_job_req(&req).await,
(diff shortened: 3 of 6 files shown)


daphne/src/aborts.rs
```diff
@@ -134,9 +137,8 @@ impl DapAbort {
                 Some("The request indicates an aggregation job that does not exist.".into()),
                 Some(agg_job_id_base64url),
             ),
-            Self::ReportTooLate | Self::UnrecognizedMessage | Self::UnrecognizedTask => {
-                (None, None, None)
-            }
+            Self::UnrecognizedMessage { detail, task_id } => (task_id, Some(detail), None),
```
cjpatton, line 140: We could move this to the arm above to make the code less redundant:
```
            | Self::QueryMismatch { detail, task_id }
            | Self::UnauthorizedRequest { detail, task_id }
            |  Self::UnrecognizedMessage { detail, task_id }  => (Some(task_id), Some(detail), None),
```

daphne/src/aborts.rs
```diff
@@ -252,8 +254,8 @@ impl DapAbort {
             Self::UnrecognizedAggregationJob { .. } => {
                 ("Unrecognized aggregation job", Some(self.to_string()))
             }
-            Self::UnrecognizedMessage => {
-                ("Failed to parse the request body", Some(self.to_string()))
+            Self::UnrecognizedMessage { .. } => {
+                ("Malformed or unexpected message", Some(self.to_string()))
```
cjpatton, line 258: This language better matches what we'll see in the next version of the spec: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/main/draft-ietf-ppm-dap.md?plain=1#L502
```suggestion
                ("Malformed or invalid message", Some(self.to_string()))
```

daphne/src/aborts.rs
```diff
@@ -282,7 +284,10 @@ impl From<DapError> for DapAbort {
 
 impl From<CodecError> for DapAbort {
     fn from(_e: CodecError) -> Self {
```
cjpatton, line 286: Since we're  not ignoring this variable anymore :)
```suggestion
    fn from(e: CodecError) -> Self {
```

daphne/src/roles.rs
```diff
@@ -330,7 +330,10 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
 
         if report.encrypted_input_shares.len() != 2 {
             // TODO spec: Decide if this behavior should be specified.
-            return Err(DapAbort::UnrecognizedMessage);
+            return Err(DapAbort::UnrecognizedMessage {
+                detail: "expected exactly two encrypted input shares".into(),
```
cjpatton, line 334: Consider including the number of aggregate shares we got.
```suggestion
                detail: format!("expected exactly two encrypted input shares; got {}", report.encrypted_input_shares.len()),
```

cjpatton: A rule of thumb about error handling: Errors should not leak anything that's not observable on the wire. In this case the list of encrypted input share are visible to the party who gets this error, so it's not sensitive.

daphne/src/roles.rs
```diff
@@ -822,7 +825,10 @@ pub trait DapHelper<S>: DapAggregator<S> {
                             .map(|report_share| &report_share.report_metadata);
                     } else if using_taskprov != 0 {
                         // It's not all taskprov or no taskprov, so it's an error.
-                        return Err(DapAbort::UnrecognizedMessage);
+                        return Err(DapAbort::UnrecognizedMessage {
+                            detail: "some reports use taskprov and some do not".to_string(),
```
cjpatton, line 829:
```suggestion
                            detail: "some reports include the taskprov extensions and some do not".to_string(),
```

daphne/src/roles_test.rs
```diff
@@ -669,7 +669,7 @@ async fn http_post_aggregate_zero_round(version: DapVersion) {
         .await;
     assert_matches!(
         t.helper.http_post_aggregate(&req).await,
-        Err(DapAbort::UnrecognizedMessage)
+        Err(DapAbort::UnrecognizedMessage { .. })
```
cjpatton, line 672: More idiomatic (you will need to implement `Default` for `UnrecognizedMessage`). Here and below.
```suggestion
        Err(DapAbort::UnrecognizedMessage::default())
```

daphne/src/vdaf/mod.rs
```diff
@@ -531,7 +531,13 @@ impl VdafConfig {
         let mut transitions = Vec::with_capacity(num_reports);
         for report_share in agg_job_init_req.report_shares.iter() {
             if processed.contains(&report_share.report_metadata.id) {
-                return Err(DapAbort::UnrecognizedMessage);
+                return Err(DapAbort::UnrecognizedMessage {
+                    detail: format!(
+                        "report ID {} has already been processed",
```
cjpatton, line 536: This is a little bit ambiguous: Does "processed" mean the report was replayed or already appeared in the current request? I believe the latter is intended.

daphne/src/vdaf/mod.rs
```diff
@@ -594,7 +600,10 @@ impl VdafConfig {
         metrics: &ContextualizedDaphneMetrics<'_>,
     ) -> Result<DapLeaderTransition<AggregationJobContinueReq>, DapAbort> {
         if agg_job_resp.transitions.len() != state.seq.len() {
-            return Err(DapAbort::UnrecognizedMessage);
+            return Err(DapAbort::UnrecognizedMessage {
+                detail: "length mismatch".into(),
```
cjpatton, line 604: Length of what? I'm not sure this will be clear without a bit of context.

daphne/src/vdaf/mod.rs
```diff
@@ -606,7 +615,13 @@ impl VdafConfig {
         {
             // TODO spec: Consider removing the report ID from the AggregationJobResp.
             if helper.report_id != leader_report_id {
-                return Err(DapAbort::UnrecognizedMessage);
+                return Err(DapAbort::UnrecognizedMessage {
+                    detail: format!(
+                        "helper ({}) and leader report ID ({}) mismatch",
```
cjpatton, line 620: I think we also need more context here: which reports are we comparing here? what error did the peer make that caused this to happen?

daphne/src/vdaf/mod.rs
```diff
@@ -737,8 +760,23 @@ impl VdafConfig {
             // would be nice if we didn't have to keep track of the set of processed reports. One
             // way to avoid this would be to require the leader to send the reports in a well-known
             // order, say, in ascending order by ID.
-            if !recognized.contains(&leader.report_id) || processed.contains(&leader.report_id) {
-                return Err(DapAbort::UnrecognizedMessage);
```
cjpatton, line 741: Similar comment here about context.

cjpatton (commented):
Nice, this looks pretty good! Most of my comments are just about adding some context to the problem details. Imagine you're an engineer trying to debug the service and you're looking at the problem details in logs.

daphne/src/aborts.rs
tholop: I'm not sure I can do that here since the `task_id` field in `UnrecognizedMessage` is actually of type `Option<TaskID>`, while it is of type `TaskID` for the other structs.

daphne/src/roles_test.rs
tholop: I just looked a bit into that, is this really what we want? Right now I'm checking that the request returns any sort of `UnrecognizedMessage`, but with `default()` it would check that the request returns exactly the default `UnrecognizedMessage` (which is not the case when `detail` contains some information about the round), no? I might be misunderstanding some notation though.

daphne/src/aborts.rs
cjpatton: Oh you're right, good catch :)

daphne/src/roles_test.rs
cjpatton: Ahh you're right, I was a bit too quick here. This is an arm of a match statement, so it doesn't make sense to compare to the default value.

daphne/src/vdaf/mod.rs
tholop: Leaving this one unresolved, as I'm not sure my new message is correct. I added: `"The Leader is preparing report ID ({}), but it received {} from the Helper's last aggregate response. The Helper must return reports in the same order as AggregationJobInitReq.report_shares"`

daphne/src/vdaf/mod.rs
```diff
@@ -531,7 +531,13 @@ impl VdafConfig {
         let mut transitions = Vec::with_capacity(num_reports);
         for report_share in agg_job_init_req.report_shares.iter() {
             if processed.contains(&report_share.report_metadata.id) {
-                return Err(DapAbort::UnrecognizedMessage);
+                return Err(DapAbort::UnrecognizedMessage {
+                    detail: format!(
+                        "report ID {} already appeared in the current Aggregate initialization request being processed by the Helper",
```
cjpatton, line 536: This works, but it can be simplified a bit more :)
```suggestion
                        "report ID {} appears twice in the same aggregation job",
```

daphne/src/vdaf/mod.rs
```diff
@@ -594,7 +600,14 @@ impl VdafConfig {
         metrics: &ContextualizedDaphneMetrics<'_>,
     ) -> Result<DapLeaderTransition<AggregationJobContinueReq>, DapAbort> {
         if agg_job_resp.transitions.len() != state.seq.len() {
-            return Err(DapAbort::UnrecognizedMessage);
+            return Err(DapAbort::UnrecognizedMessage {
+                detail: format!(
+                    "the Leader has {} reports, but it received {} reports from the Helper (while handling an aggregate response)",
```
cjpatton, line 605:
```suggestion
                    "aggregation job response has {} reports; expected {}",
```

daphne/src/vdaf/mod.rs
```diff
@@ -606,7 +619,13 @@ impl VdafConfig {
         {
             // TODO spec: Consider removing the report ID from the AggregationJobResp.
             if helper.report_id != leader_report_id {
-                return Err(DapAbort::UnrecognizedMessage);
+                return Err(DapAbort::UnrecognizedMessage {
+                    detail: format!(
+                        "The Leader is preparing report ID ({}), but it received {} from the Helper. The Helper must return reports in the same order as AggregationJobInitReq.report_shares",
```
cjpatton, line 624:
```suggestion
                        "report ID {} appears out of order in aggregation job response",

cjpatton: Similarly below.

cjpatton (approved):
LGTM. I think the details can be siimplified somewhat, but they're fine as is.

cjpatton:
Note that you'll need to rebase your branch on main in order to resolve conflicts.

cjpatton (changes requested):

cjpatton (approved):
🦆

PR #346, opened 2023-06-12
Improve hpke_config visibility

mendess:
- first commit: Make sure internal fatal error details are not leaked out to the api user.
- second commit: Require that fatal errors are created using the `fatal_error`
    macro, which will guarantee that all fatal errors are logged to tracing.


daphne/src/lib.rs
```diff
@@ -981,6 +927,7 @@ pub mod auth;
 pub mod constants;
 #[cfg(test)]
 mod constants_test;
+pub mod dap_error;
```
cjpatton, line 930: I think this should be called "error", for consistency with the "abort" module.

daphne/src/dap_error.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl std::error::Error for FatalDapError {}
+
+impl Display for FatalDapError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+#[macro_export]
```
cjpatton, line 53: Why is a macro needed? I wonder if we couldn't just add a method on `DapError` (or maybe replace `DapError::fatal()`) with something that does the same thing as this code.

The reason I would prefer not to use a macro (or at least not this one) Is that, in use, the syntax looks relatively complex. For example:

```rust
fatal_error!(
                err = format!("{e:?}"), // `HpkeError` doesn't implement Display or Error :(
                ?kem_id,
                "bad key generation for KEM",
            )
```
The `err = ` and `?kem_id` look really strange to someone who is not as expert at Rust as you (me :D)

daphne/src/dap_error.rs
cjpatton, line 1: * All files must have copyright notices.
* What's the reasoning for sticking this stuff inside its own module?

daphne/src/lib.rs
mendess: yeah I thought about that, but there are multiple errors defined in this crate (`TransitionFailure`, `DapAbort`, for example) and since each is defined in a different module it felt a bit misleading to have this be called `error`, as usually such a module houses all the errors of a crate.

daphne/src/dap_error.rs
mendess: Maybe I should have specified the design goals of the macro, I'm not 100% happy with it either but I pushed the PR because I wanted feedback.

I wanted the first argument to it to be an implementation of `std::error::Error`, sadly if I require that `&str` and `String` aren't allowed and that just breaks a lot of current code, so I added the `err =` thing to force the caller to think about it a bit and pass in the error in situations like `.map_err(|e| ...)` where you do have an error.

The `?kem_id` part is just how tracing works and so it felt like something okay to introduce, since this is effectively a thin wrapper around `tracing` anyway.

Finally the big reason I want this is:
- it forces the `fatal_error` field on the log line, which is easy to filter for in kibana
- it makes the line numbers happen exactly where the error is generating, making it really easy to trace back to the origin

daphne/src/dap_error.rs
mendess: I wanted the constructor of `DapFatalError` to be private, if I left it in the `lib.rs` it would be public for all submodules

daphne/src/dap_error.rs
cjpatton: > ... and that just breaks a lot of current code ...

Does this indicate to you that there is a desirable code design pattern that we are not following (consistently)? If so, it might be worth taking the time to fix/refactor the code such that it does.

I don't object to this macro if you think it's the right thing to do. The main consideration for me is that we do right thing rather than the fast thing.

daphne/src/dap_error.rs
```diff
@@ -0,0 +1,71 @@
...
+pub struct FatalDapError(String);
+
+impl std::error::Error for FatalDapError {}
+
+impl Display for FatalDapError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
```
cjpatton, line 55: Because the syntax is somewhat weird, can we add a comment here explaining how the macro is used, how the tracing part works, etc.? It would also be useful to document, with examples, all of the variants that the macro accepts.

daphne/src/lib.rs
cjpatton: `TransitionFailure` is a protocol message and so is not an error as such. (It's never used as the error type of `Result`, IIRC.) 

What about merging `abort` and `dap_error` into a common module, called `error`?

daphne/src/testing.rs
```diff
@@ -300,7 +301,7 @@ impl HpkeDecrypter for MockAggregator {
         task_id: Option<&TaskId>,
     ) -> Result<Self::WrappedHpkeConfig<'s>, DapError> {
         if self.hpke_receiver_config_list.is_empty() {
-            return Err(DapError::fatal("emtpy HPKE receiver config list"));
+            return Err(fatal_error!(err = "emtpy HPKE receiver config list"));
```
cjpatton, line 304: While we're here ...
```suggestion
            return Err(fatal_error!(err = "empty HPKE receiver config list"));
```

daphne/src/testing.rs
```diff
@@ -773,12 +774,12 @@ impl DapLeader<BearerToken> for MockAggregator {
...
             .leader_state_store
             .lock()
-            .map_err(|e| DapError::Fatal(e.to_string()))?;
+            .map_err(|e| fatal_error!(err = e))?;
         let leader_state_store = leader_state_store_mutex_guard.deref_mut();
 
         let leader_state = leader_state_store
             .get(task_id)
-            .ok_or_else(|| DapError::fatal("collect job not found for task_id"))?;
+            .ok_or_else(|| fatal_error!(err = "collect job not found for task_id", %task_id))?;
```
cjpatton, line 782: I am not familiar with the `%` syntax here. Please document it along with the macro definition.

daphne/src/vdaf/mod.rs
```diff
@@ -385,7 +388,8 @@ impl VdafConfig {
                 extensions: vec![],
                 payload: encoded_input_share,
             },
-            _ => PlaintextInputShare::get_decoded(&encoded_input_share)?,
+            _ => PlaintextInputShare::get_decoded(&encoded_input_share)
+                .map_err(|e| fatal_error!(err = e))?,
```
cjpatton, line 392: `DapError::Fatal` is intended for situations where the host encountered an unrecoverable error of its own making. In this particular situation, the error is triggered by the peer, so the appropriate action is to abort. I.e., I believe the current behavior is to map the CodeError into a `DapError::Abort`.

cjpatton (changes requested):
Nice, this is a huge improvement! One minor bug (hence "Requested changes"), all others are suggestions. The main thing is to document how to use fatal_error!() since it looks kind of magical.

daphne/src/vdaf/mod.rs
mendess: I'll review, but if I made this change it's because the compiler forced me to, which came about from me deleting the from conversions for DapError. Which means this was already wrong to begin with. I'll change it to an abort

cjpatton: On `main` we have this:
```rust
impl From<CodecError> for DapError {
    fn from(e: CodecError) -> Self {
        Self::Fatal(format!("codec: {e}"))
    }
}
```
I believe you just need to add that back?

cjpatton: Ah, which is incorrect: It should be `Self::Abort`.

cjpatton (approved):
🦆 - feel free to merge after resolving conflicts.

PR #349, opened 2023-06-15
Parallelize DaphneWorker::internal_delete_all

mendess:


cjpatton (approved):
Good idea, this ought to speed up e2e tests.

Issue #350, opened 2023-06-20
DAP-07


cjpatton:
The plan will be to push changes to a development branch ([draft05](https://github.com/cloudflare/daphne/tree/draft05)) until we're ready to drop support for DAP-04.

- [ ] Bump version
- [x] Ping-pong flow
- [ ] PUT idempotency https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/471/files
- [ ] Implement the latest taskprov version

cjpatton:
Can we rename ﻿﻿`Transition` to `PrepareResp`?

mendess:
`PrepareState`? Since it's basically a state machine?

PR #351, opened 2023-06-20
Break compatibility with draft 04

cjpatton:
Partially addresses #350.


mendess (commented):
This is a kv breaking change at least, right? What else could this possibly break?

cjpatton:
> This is a kv breaking change at least, right? What else could this possibly break?

This is a breaking change for draft-04 of the spec only; draft-02 is not impacted.

bhalleycf (approved):

cjpatton:
Fixed clippy

PR #353, opened 2023-06-20
Improve kv access patterns by not listing all keys

mendess:
## Commits
- move HpkeConfig into hpke module
    I was finding it very confusing that some parts of hpke were not in the hpke module, so I reorganized this.
- Use fmt::Display instead of ToString
    It is more ideomatic and slightly more efficient to use Display over ToString
- Change hpke config schema in kv
    See bellow for longer explanation.
- Configure the hpke receiver config from the tests
    See commit message for more details.

## New Kv Schema
Instead of having one key per config like so
- hpke_receiver_config/version/v02/config_id/24 => `{ public_key, private_key }`
- hpke_receiver_config/version/v02/config_id/25 => `{ public_key, private_key }`
- hpke_receiver_config/version/v02/config_id/26 => `{ public_key, private_key }`
- hpke_receiver_config/version/v04/config_id/45 => `{ public_key, private_key }`

which requires listing the keys from kv to find the available configs, this commit changes the schema to store the configs in a single key per version in the following schema:
- hpke_receiver_config/v02 =>
    ```
    {
        24: { public, private },
        25; { public, private },
        26; { public, private },
    }
    ```
- hpke_receiver_config/v04 =>
    ```
    {
        45: { public, private },
    }
    ```

This avoids the `kv::list` operation, and makes the implementation of key rotation in the future feasible.

## Removed automatic generation
I also decided to remove the automatic generation of configurations since it's a code path that was never used in production and key configurations should be managed from outside the worker. On top of that it relied on the `set_if_not_exists` method whose implementation has a TOCTOU bug


daphne_worker_test/tests/test_runner.rs
```diff
@@ -236,6 +236,36 @@ impl TestRunner {
             res.status, res.error
         );
 
+        let mk_config = || {
```
cjpatton, line 239: nit: "generate_config" or "gen_config" (prefer to match existing verbiage)

daphne_worker_test/tests/test_runner.rs
```diff
@@ -236,6 +236,36 @@ impl TestRunner {
...
+            )
+            .await;
+
+        assert_eq!(
+            res.status, "success",
+            "response status: {}, error {:?}",
+            res.status, res.error
+        );
+
+        let res: InternalTestAddTaskResult = t
```
cjpatton, line 256: We probably should replace `InternalTestAddTaskResult` with a more general name. How about `InternalTestCommandResult`?

daphne_worker/src/config.rs
```diff
@@ -645,14 +647,29 @@ impl<'srv> DaphneWorker<'srv> {
...
-            Cow::Owned(hpke_receiver_kv_key),
-        )
-        .await
+        version: DapVersion,
+        config_id: u8,
+    ) -> Result<Option<HpkeReceiverConfig>> {
+        Ok(self
+            .get_hpke_receiver_configs(version)
+            .await?
+            .and_then(|mut configs| configs.remove(&config_id)))
```
cjpatton, line 659: Why remove it? We're likely going to use the same `HpkeReceiverConfig` multiple times in the same request. If I understand the code correctly, we would end up hitting KV multiple times as a result.

daphne/src/hpke.rs
```diff
@@ -60,6 +63,104 @@ fn check_suite<T: HpkeCrypto>(
     }
 }
 
+/// Codepoint for KEM schemes compatible with HPKE.
```
cjpatton, line 66: What's the motivation for moving the `HpkeConfig`? The reason it's in `messages` now is that this module contains all of the protocol messages.

Also, did the implementation change? It's kind of hard to tell as a result of the code movement (though I didn't look closely yet).

daphne_worker/src/config.rs
```diff
@@ -360,6 +360,8 @@ impl DaphneWorkerConfig {
     }
 }
 
+pub(crate) type HpkeConfigs = HashMap<u8, HpkeReceiverConfig>;
```
cjpatton, line 363: nit: `HpkeConfigs` is slightly difficult to visually distinguish from `HpkeConfig`. (I had some type confusion while reviewing.) A more distinguished name would be useful.

cjpatton (commented):
I'm on board with the new schema. Just a few minor comments, mostly to clarify.

daphne_worker/src/config.rs
mendess: the method returns an Owned value, so any mutations on it doesn't affect anyone else. I use `remove` so that I can move the config out of the hashmap

daphne/src/hpke.rs
mendess: there was no implementation change, the one thing I added was a `From<u16> for HpkeKemId|HpkeFdfId|HpkeAreaId`.

The reason I moved it out of there is because we have a dedicated hpke module, and hpke is used outside of the domain of messages (stored in kv, used for encryption and decryption, etc). So it felt odd of to have that be in messages. What makes sense to have in messages is the `Encode` and `Decode` implementations for these types, since this is a concern of serialization and not of cryptography. What I left in the messages module was the `HpkeConfigList` (which the diff poorly demonstrates).

daphne_worker/src/config.rs
mendess: would `HpkeKeySet` be a good name? :thinking: I don't like the name I picked either :sweat_smile:

cjpatton: How about `HpkeRecieverConfigSet`? It's a little verbose but it is consistent.

daphne_worker/src/config.rs
```diff
@@ -645,14 +650,29 @@ impl<'srv> DaphneWorker<'srv> {
...
+    /// Get a reference to the HPKE receiver configs, ensuring that the config indicated by
+    /// `hpke_config_id` is cached (if it exists).
+    pub(crate) async fn get_hpke_receiver_configs(
+        &self,
+        version: DapVersion,
+    ) -> Result<Option<HpkeRecieverConfigSet>> {
+        Ok(self
+            .kv_get_cached(
+                &self.isolate_state().hpke_receiver_configs,
+                KV_KEY_PREFIX_HPKE_RECEIVER_CONFIG,
```
cjpatton, line 671: I think it makes sense to rename this variable now that this string is no longer a prefix of the KV key. (It's now the KV key itself.) While at it, what do you think about renaming the key to `"hpke_receiver_config_set"`? It's currently `"hpke_receiver_config"`.
```suggestion
                KV_KEY_HPKE_RECEIVER_CONFIG_SET,
```

daphne/src/hpke.rs
cjpatton: Ack. I'll defer to your preference here, but I think it would be best to remove this change and discuss it in another PR. As a general rule (at least in my opinion), code movement should be avoided unless clearly related to the PR.

In any case I carefully reviewed the changes and they look correct to me.

cjpatton (commented):
One last thing before merging: Can we avoid cloning the entire HPKE receiver config set? (See inline comments.)

daphne_worker/src/config.rs
cjpatton: Bump

daphne_worker/src/dap.rs
```diff
@@ -173,9 +111,11 @@ impl<'srv> HpkeDecrypter for DaphneWorker<'srv> {
     ) -> std::result::Result<bool, DapError> {
         let version = self.try_get_task_config(task_id).await?.as_ref().version;
         Ok(self
-            .get_hpke_receiver_config(HpkeReceiverKvKey {
-                version,
-                hpke_config_id: config_id,
+            .get_hpke_receiver_config(version, |config_set| {
+                config_set.get(&config_id).map(|_| {
```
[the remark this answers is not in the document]
mendess: this would result in a an `Option<bool>` and the `is_some` check below would be wrong. I'd have to change it to `.unwrap_or(false)` which is one of my personal pet peeves because of how unreadable it is :sweat_smile:

PR #358, opened 2023-06-22
Reduce metrics error cardinality

mendess:
Using display implementations for program behavior sensitive code is a bad
idea. Different parts of the code need different string representations.

### DIFF
--- a/daphne_worker/src/config.rs
+++ b/daphne_worker/src/config.rs
@@ -522,7 +522,7 @@ impl<'srv> DaphneWorkerRequestState<'srv> {
         };
         self.metrics
             .dap_abort_counter
-            .with_label_values(&[&self.host, &e.to_string()])
+            .with_label_values(&[&self.host])
             .inc();
         let problem_details = e.into_problem_details();
         error!(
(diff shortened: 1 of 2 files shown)


daphne/src/error/aborts.rs
```diff
@@ -106,6 +106,27 @@ pub enum DapAbort {
 }
 
 impl DapAbort {
+    pub fn as_kind_str(&self) -> &'static str {
+        match self {
+            DapAbort::BadRequest(_) => "bad request",
+            DapAbort::BatchInvalid { .. } => "batchInvalid",
+            DapAbort::BatchMismatch { .. } => "batchMismatch",
+            DapAbort::BatchOverlap { .. } => "batchOverlap",
+            DapAbort::Internal(_) => "internal error",
```
will118, line 115: is it possible to camelCase these 2 (`BadRequest` and `Internal`) for consistency?

mendess: I didn't do it so that it retained the old behavior as much as possible, but it did bother me as well :sweat_smile:

oliy (changes requested):

daphne_worker/src/config.rs
```diff
@@ -522,7 +522,7 @@ impl<'srv> DaphneWorkerRequestState<'srv> {
         };
         self.metrics
             .dap_abort_counter
-            .with_label_values(&[&self.host, &e.to_string()])
```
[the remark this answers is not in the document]
mendess: I'm confused. This removes the error details from the metrics system and the details were already in a log message

lbaquerofierro- (approved):

will118 (approved):

PR #360, opened 2023-07-03
Add exponential backoff to retries

mendess:


daphne_worker/src/durable/mod.rs
```diff
@@ -207,7 +207,14 @@ impl<'srv> DurableConnector<'srv> {
                 Err(err) => {
                     if attempt < attempts {
                         warn!("DO {durable_binding}: post {durable_path}: attempt #{attempt} failed: {err}");
-                        Delay::from(DEFAULT_RETRY_INTERVAL).await;
+                        Delay::from(Duration::from_millis(
+                            DEFAULT_RETRY_INTERVAL
+                                .as_millis()
+                                .pow(attempt)
+                                .try_into()
+                                .unwrap_or_default(),
```
cjpatton, line 215: We probably want `unwrap_or(u32::MAX)` so the interval doesn't go down to `0`.

cjpatton (approved):

daphne_worker/src/durable/mod.rs
mendess: hmm, but u32::MAX is not a safe value either, that would take an eternity to resolve :thinking: either way it will never happen (I did the math) so it should just be a straight `.expect("delay time in retries can should never overflow"`

cjpatton: Agreed -- edge cases we don't expect should fail loudly.

daphne_worker/src/durable/mod.rs
```diff
@@ -207,7 +207,14 @@ impl<'srv> DurableConnector<'srv> {
                 Err(err) => {
                     if attempt < attempts {
                         warn!("DO {durable_binding}: post {durable_path}: attempt #{attempt} failed: {err}");
-                        Delay::from(DEFAULT_RETRY_INTERVAL).await;
+                        Delay::from(Duration::from_millis(
+                            DEFAULT_RETRY_INTERVAL
+                                .as_millis()
+                                .pow(attempt)
+                                .try_into()
+                                .expect("delay time in retries can should never overflow"),
```
cjpatton, line 215:
```suggestion
                                .expect("delay time in retries should never overflow"),
```

cjpatton (changes requested):

cjpatton (approved):

PR #362, opened 2023-07-07
Prepare for offloading VDAF preparation

cjpatton:
We want to be able to offload VDAF preparation to durable objects. This PR refactors the code to make this easier and resolves some prerequisites. Details in the commit messages, but at a high level:
* Split `consume_report_share()` into two steps and wrap the state after each step with a state object.
* Implement codec traits on VdafPrepState and VdafPrepMessage, which will be required for serializing the input and output to DO.
* Fix a bug that triggers abort if a reports plaintext input share is malformed. (We should just reject instead.)

### DIFF
--- a/daphne/src/vdaf/mod.rs
+++ b/daphne/src/vdaf/mod.rs
@@ -39,7 +39,7 @@ use prio::{
 };
 use rand::prelude::*;
 use serde::{Deserialize, Serialize};
-use std::{collections::HashSet, convert::TryInto};
+use std::{borrow::Cow, collections::HashSet, convert::TryInto};
 
 const CTX_INPUT_SHARE_DRAFT02: &[u8] = b"dap-02 input share";
 const CTX_INPUT_SHARE_DRAFT04: &[u8] = b"dap-04 input share";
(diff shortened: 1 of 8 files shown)


daphne/src/vdaf/mod.rs
```diff
@@ -75,12 +77,142 @@ impl AsRef<[u8]> for VdafVerifyKey {
...
-pub struct VdafPrepInput<'req> {
-    pub(crate) public_share: Cow<'req, Vec<u8>>,
-    pub(crate) input_share: Vec<u8>,
-    pub(crate) metadata: Cow<'req, ReportMetadata>,
+pub enum EarlyReportStateConsumed<'req> {
+    Ready {
+        metadata: Cow<'req, ReportMetadata>,
+        public_share: Cow<'req, [u8]>,
+        input_share: Vec<u8>,
+    },
```
mendess, line 88: The use of `Cow` here seems unnecessary, unless I'm missing something

mendess (approved):
LGTM

daphne/src/vdaf/mod.rs
cjpatton: So my thinking is that we're going to want an owned variant of this struct when we plumb it in DO. I'll remove this if it turns out not to be necessary.

PR #370, opened 2023-07-13
Upgrade Libprio to 0.13.0 / move to VDAF-06

tholop:
I upgraded Libprio and implemented the new histogram format introduced by VDAF-06.

 I have one question: histogram measurements are of type `usize` in Libprio (https://github.com/divviup/libprio-rs/blob/main/src/flp/types.rs#L353), but so far we've been using the `DapMeasurement::U64` type for histograms. Right now I am casting to an `usize`, which could cause problems on 32-bit systems. Should we add a U32 measurement type, or is it fine to keep things as they are?

Two other notes, about interoperability:
- the TaskProv spec doesn't support VDAF-06, so I had to comment out some test vectors until the spec catches up. See https://github.com/wangshan/draft-wang-ppm-dap-taskprov/issues/33
- similarly for https://datatracker.ietf.org/doc/html/draft-dcook-ppm-dap-interop-test-design-04#name-internal-test-add_task. I still added an endpoint as it might be useful for debugging, as a temporary fix for https://github.com/cloudflare/daphne/issues/271


daphne/src/messages/taskprov.rs
```diff
@@ -54,7 +54,7 @@ impl KeyType for VdafType {
 pub enum VdafTypeVar {
     Prio3Aes128Count,
     Prio3Aes128Sum { bit_length: u8 },
-    Prio3Aes128Histogram { buckets: Vec<u64> },
+    Prio3Aes128Histogram { len_length: u8 },
```
cjpatton, line 57: * `len_length` is an odd name. Maybe just `len`?
* `u8` is probably too small. I'd go with `u32`.
* We need to add a note here that this doesn't comply with the spec because the spec is out-of-date. Also, let's link to the tracking issue here.

daphne/src/vdaf/prio3.rs
```diff
@@ -52,9 +52,9 @@ pub(crate) fn prio3_shard(
             let vdaf = Prio3::new_count(2)?;
             Ok(shard!(vdaf, &measurement, nonce))
         }
-        (Prio3Config::Histogram { buckets }, DapMeasurement::U64(measurement)) => {
-            let vdaf = Prio3::new_histogram(2, buckets)?;
-            Ok(shard!(vdaf, &(measurement as u128), nonce))
+        (Prio3Config::Histogram { len }, DapMeasurement::U64(measurement)) => {
+            let vdaf = Prio3::new_histogram(2, *len)?;
+            Ok(shard!(vdaf, &(measurement as usize), nonce))
```
cjpatton, line 57: The safe option would be `usize::try_from()`

daphne_worker/src/config.rs
```diff
@@ -924,6 +924,12 @@ impl<'srv> DaphneWorker<'srv> {
                 let len = length.parse().map_err(int_err)?;
                 VdafConfig::Prio3(Prio3Config::SumVec { bits, len })
             }
+            // NOTE: the VDAF-06 format is not supported yet
+            // https://datatracker.ietf.org/doc/html/draft-dcook-ppm-dap-interop-test-design-04#name-internal-test-add_task
```
cjpatton, line 928: Let's do David a favor and file an issue: https://github.com/divergentdave/draft-dcook-ppm-dap-interop-test-design/issues

daphne/src/messages/taskprov.rs
tholop: For some reason I thought this was the number of bytes required to store the `len`, but indeed it's just the length itself (https://www.ietf.org/id/draft-wang-ppm-dap-taskprov-04.html#section-3-9). I moved to `u32` assuming that no one will need histograms with more than 4.3 billion buckets.

daphne/src/vdaf/prio3.rs
tholop: Ok, my new code will panic if `try_from` fails while converting a `u64` measurement on a 32-bit system.

cjpatton (approved):

PR #371, opened 2023-07-14
Reorganize e2e tests

mendess:
This organization let's us remove the `allow(dead_code)` directives and also
lets us remove the `e2e_` prefix from the tests since the test name will get
that from the module name.


cjpatton (approved):
🙏 Much nicer

PR #372, opened 2023-07-19
Reorganize daphne

mendess:
I've split the `roles.rs` file into three
- aggregator.rs where `DapAggregator` and `DapAuthorizedSender` live
- helper.rs where `DapHelper` live
- leader.rs where `DapLeader` and `DapReportInitializer` live

This makes it more obvious where to find the specific functionality of each of the roles and it also let's us reduce the scope certain utilities are visible to, for example, it's now obvious that `fn check_part_batch` is only used for `DapHelper` and doesn't need to pollute the space around the other traits.

The second change is putting the test modules inline bellow everything else, this avoids having files related to tests in the file tree which makes a file tree view and file file fuzzy finder harder to use. They are both also the more idiomatic way of doing things.

None of these changes have a visible effect on import statements outside of the `daphne` crate since I've reexported all types that changed place in the places they use to be before


daphne/src/roles/aggregator.rs
```diff
@@ -0,0 +1,192 @@
+use std::borrow::Cow;
```
cjpatton, line 1: Add copyright to new files

cjpatton (commented):
I don't object, but this is quite a lot of code movement, so we should be really sure this is the right structure.

I'll also ask @armfazh to chime in on inlining the tests ... he originally had the idea to split them into their own modules. One advantage is that it makes the size of the files way shorter and thus easier to find stuff (at least in my opinion). Though one obvious disadvantage is that some thing have to be `pub(crate)` just because there is test code for them in another module.

Also, before merging we need to make sure to add copyright notices on each of the new files.

mendess:
I don't think in-lining the test modules makes it harder to find stuff, sure the files are larger but the test module is always at the bottom, so it doesn't really increase the file size in a a way that makes stuff harder to find, if you see the test module you know you've reached the end of the file.

cjpatton:
Hearing no objection from @armfazh, let's go ahead with the change. I trust your judgment.

cjpatton (approved):

PR #373, opened 2023-07-21
Improve error visibility

mendess:


cjpatton (approved):
Yup, that seems to solve the problem :) We get the error types for 400s but don't expose all of the variants for internal errors.

PR #374, opened 2023-07-21
Wait to mark reports as aggregated until just before committing

cjpatton:
Loads that require a lot of CPU time can trigger the Workers runtime to reset itself, causing a 500 error. This is most likely to happen when processing the AggregationJobInitReq, as this involves the expensive, VDAF prep initialization step. Some reports may still get marked as aggregated; if the peer retries the request, those reports will get rejected.

This problem is so common that we need to make this request idempotent. As a first step, wait to mark the reports as aggregated until just before we have committed to aggregating them. In particular, instead of doing this in `DapReportInitializer::initialize_reports()`, wait until `DapAggregator::put_out_shares()`.

### DIFF
--- a/daphne/src/lib.rs
+++ b/daphne/src/lib.rs
@@ -310,20 +310,19 @@ impl DapTaskConfig {
         self.quantized_time_lower_bound(time) + self.time_precision
     }
 
-    /// Compute the "batch span" of a set of output shares and, for each buckent in the span,
-    /// aggregate the output shares into an aggregate share.
+    /// Compute the "batch span" of a set of output shares.
     pub fn batch_span_for_out_shares<'a>(
         &self,
         part_batch_sel: &'a PartialBatchSelector,
         out_shares: Vec<DapOutputShare>,
-    ) -> Result<HashMap<DapBatchBucket<'a>, DapAggregateShare>, DapError> {
+    ) -> Result<HashMap<DapBatchBucket<'a>, Vec<DapOutputShare>>, DapError> {
         if !self.query.is_valid_part_batch_sel(part_batch_sel) {
             return Err(fatal_error!(
                 err = "partial batch selector not compatible with task",
             ));
         }
 
-        let mut span: HashMap<DapBatchBucket<'a>, DapAggregateShare> = HashMap::new();
+        let mut span: HashMap<DapBatchBucket<'a>, Vec<DapOutputShare>> = HashMap::new();
         for out_share in out_shares.into_iter() {
             let bucket = match part_batch_sel {
                 PartialBatchSelector::TimeInterval => DapBatchBucket::TimeInterval {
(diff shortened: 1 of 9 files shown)


daphne/src/lib.rs
```diff
@@ -334,14 +333,7 @@ impl DapTaskConfig {
...
             };
 
-            let agg_share = span.entry(bucket).or_default();
-            agg_share.merge(DapAggregateShare {
-                report_count: 1,
-                min_time: out_share.time,
-                max_time: out_share.time,
-                checksum: out_share.checksum,
-                data: Some(out_share.data),
-            })?;
```
cjpatton, line 344: We can't aggregate the output shares yet because we still may need to reject some.

daphne_worker/src/dap.rs
```diff
@@ -274,38 +279,22 @@ impl DapReportInitializer for DaphneWorker<'_> {
...
-                if let EarlyReportStateInitialized::Rejected {
-                    ref mut failure, ..
-                } = initialized_report
-                {
-                    // The request to ReportsProcessed is not idempotent. If it was retried, then
-                    // it's possible that one or more of the reports were erroneously marked as
-                    // replayed. This would happen if, for example, a fresh report was marked as
-                    // aggregated in the failed request, then marked as replayed by a retried
-                    // request. In this case we cannot determine definitively if the report was
-                    // replayed, so we drop it to be safe.
```
cjpatton, line 304: We no longer mark the reports as aggregated, so this logic for handling failed DO requests is no longer necessary. Retry should be safe.

daphne/src/roles/helper.rs
```diff
@@ -245,10 +246,22 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+                    DapHelperTransition::Finish(out_shares, mut agg_job_resp) => {
                         let out_shares_count = u64::try_from(out_shares.len()).unwrap();
-                        self.put_out_shares(task_id, &part_batch_sel, out_shares)
+                        let replayed = self
+                            .put_out_shares(task_id, task_config, &part_batch_sel, out_shares)
                             .await?;
+
+                        // If there are multiple aggregation jobs in flight that contain the
+                        // same report, then we may need to reject the report at this late stage.
+                        for transition in agg_job_resp.transitions.iter_mut() {
```
bhalleycf, line 257: Wd don't need to look at all the transitions if replayed is empty, which is probably the usual case, so might be worth a "if it's not empty" check for efficiency.

bhalleycf (approved):
This makes sense to me, and I don't see any problems.

daphne/src/roles/helper.rs
cjpatton: Good idea, done.

bhalleycf (approved):

PR #375, opened 2023-07-24
Reorganize daphne_worker

mendess:
This PR cleans up the codebase a bit more by:
- Inlining test modules in daphne_worker
- Removing unnecessary macros from daphne and daphne_worker
- Moving module statements to the top of files

This PR is best viewed one commit at a time!

### DIFF
--- a/daphne_worker/src/durable/aggregate_store.rs
+++ b/daphne_worker/src/durable/aggregate_store.rs
@@ -12,6 +12,8 @@ use daphne::DapAggregateShare;
 use tracing::Instrument;
 use worker::*;
 
+use super::{DapDurableObject, GarbageCollectable};
+
 pub(crate) const DURABLE_AGGREGATE_STORE_GET: &str = "/internal/do/aggregate_store/get";
 pub(crate) const DURABLE_AGGREGATE_STORE_MERGE: &str = "/internal/do/aggregate_store/merge";
 pub(crate) const DURABLE_AGGREGATE_STORE_MARK_COLLECTED: &str =
@@ -68,18 +70,13 @@ impl DurableObject for AggregateStore {
 
 impl AggregateStore {
     async fn handle(&mut self, req: Request) -> Result<Response> {
-        let id_hex = self.state.id().to_string();
-        let ControlFlow::Continue(mut req) = super::run_garbage_collection(
-            req,
-            &self.state,
-            &self.env,
-            self.config.deployment,
-            &mut self.touched,
-            id_hex,
-            BINDING_DAP_AGGREGATE_STORE,
-        )
-        .await? else {
-            return Response::from_json(&());
+        let mut req = match self
+            .schedule_for_garbage_collection(req, BINDING_DAP_AGGREGATE_STORE)
+            .await?
+        {
+            ControlFlow::Continue(req) => req,
+            // This req was a GC request and as such we must return from this function.
+            ControlFlow::Break(_) => return Response::from_json(&()),
         };
 
         match (req.path().as_ref(), req.method()) {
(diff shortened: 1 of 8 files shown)


daphne/src/roles/leader.rs
```diff
@@ -22,47 +22,54 @@ use crate::{
...
-            ),
-            taskprov: None,
-        };
+struct LeaderPostOptions<'p> {
+    path: &'p str,
+    req_media_type: DapMediaType,
+    resp_media_type: DapMediaType,
+    resource: DapResource,
+    req_data: Vec<u8>,
+    is_put: bool,
```
cjpatton, line 31: If set, this PUTs instead of POSTs. Since we're changing this code, we might as well rename it to be more accurate. How about "leader_send_http_request" and "LeaderHttpRequestOptions"?

daphne_worker/src/durable/aggregate_store.rs
```diff
@@ -57,9 +59,20 @@ impl DurableObject for AggregateStore {
...
+            req,
+            &self.state,
+            &self.env,
+            self.config.deployment,
+            &mut self.touched,
+            id_hex,
+            BINDING_DAP_AGGREGATE_STORE,
+        )
+        .await? else {
+            return Response::from_json(&());
```
cjpatton, line 74: Under what conditions does this happen, and why return this particular response? (Similar question below.)

daphne_worker/src/durable/mod.rs
```diff
@@ -225,70 +234,86 @@ impl<'srv> DurableConnector<'srv> {
...
+/// Run garbage collection requests.
+///
+/// If a garbage collection request is handled no further processing needs to be done, as such,
+/// this function will consume the request, otherwise it will return the passed in request for
+/// further handling.
+async fn run_garbage_collection(
+    req: Request,
+    state: &State,
+    env: &Env,
+    deployment: crate::config::DaphneWorkerDeployment,
```
cjpatton, line 246: * Suggestion: `staet`, `env`, `deployment`, and `touched` are attributes common to all DOs that are "garbage collected". WDYTA about defining a trait for these and have `run_garbage_collection` be a derived method on this trait? I think this would make the code a lot cleaner.
* This doesn't actually "run" garbage collection: It puts the DO instance on the garbage collection queue, which is dequeued before each e2e tests.

cjpatton (commented):
Looks good overall, just a few things to clean up in my opinion. (All open for discussion, of course.)

daphne_worker/src/durable/mod.rs
cjpatton: To be clear: The second suggestion is to rename the method.

daphne_worker/src/durable/aggregate_store.rs
mendess: This is exactly what the code was doing before, but it was hidden behind the macro. I've rewritten this control flow a bit to make it more readable and added a comment.

daphne/src/roles/leader.rs
```diff
@@ -22,28 +22,33 @@ use crate::{
...
 ) -> Result<DapResponse, DapError> {
-    let LeaderPostOptions {
+    let LeaderHttpRequestOptions {
         path,
         req_media_type,
         resp_media_type,
         resource,
         req_data,
-        is_put,
+        method: is_put,
```
cjpatton, line 51: nit: s/is_put/method

daphne_worker/src/durable/aggregate_store.rs
```diff
@@ -61,17 +63,13 @@ impl DurableObject for AggregateStore {
...
-            BINDING_DAP_AGGREGATE_STORE,
-        )
-        .await? else {
-            return Response::from_json(&());
+        let mut req = match self
+            .schedule_for_garbage_collection(req, id_hex, BINDING_DAP_AGGREGATE_STORE)
+            .await?
+        {
+            ControlFlow::Continue(req) => req,
+            // this req was a GC request and as such we must return from this funcition.
```
cjpatton, line 71:
```suggestion
            // This req was a GC request and as such we must return from this function.
```

daphne_worker/src/durable/aggregate_store.rs
```diff
@@ -138,3 +136,24 @@ impl DurableObject for AggregateStore {
         }
     }
 }
+
+impl DapDurableObject for AggregateStore {
+    fn state(&self) -> &State {
+        &self.state
+    }
+
+    fn deployment(&self) -> crate::config::DaphneWorkerDeployment {
```
cjpatton, line 145: nit: It would be slightly more general to return a reference to `self..config`, i.e., change the signature here to `config(&self) -> &Confg` (I forget the name of the actual type at the meoment)

daphne_worker/src/durable/aggregate_store.rs
cjpatton: (Here and below)

cjpatton (approved):
Nice :) Just a few nits.

PR #376, opened 2023-07-28
daphne_worker: Cache the "collected" bit in AggregateStore

cjpatton:


daphne_worker/src/durable/aggregate_store.rs
```diff
@@ -104,7 +107,13 @@ impl DurableObject for AggregateStore {
             // Idempotent
             // Output: `bool`
             (DURABLE_AGGREGATE_STORE_CHECK_COLLECTED, Method::Get) => {
-                let collected: bool = state_get_or_default(&self.state, "collected").await?;
+                let collected = if let Some(collected) = self.collected {
+                    collected
+                } else {
+                    let collected = state_get_or_default(&self.state, "collected").await?;
+                    self.collected = Some(collected);
```
mendess, line 114: this branch doesn't write to storage, so if you set `self.collected` to `Some(false)` (the default) the other if-branch above will always return `false` even if `DURABLE_AGGREGATE_STORE_MARK_COLLECTED` is called in the mean time.

mendess (changes requested):

daphne_worker/src/durable/aggregate_store.rs
mendess: ah lol, I totally missed that. Than this is good to go!

mendess (approved):

PR #377, opened 2023-08-03
Propagate tracing fields to DOs

mendess:
The first commit is not related to this PR and I apologize for that, but it's
small and the code is reused later in the PR, so I included it to reduce churn.

This PR can be reviewed in 3 parts, one for each commit.

- Shorten file location to reduce logpush truncation chance. Instead of having `daphne_worker/src/durable/mod.rs:333` we instead have `wo/sr/du/mod.rs:333`, it's enough to know where the logline is, but takes up less space in the logline.

- Split the tracing JsonLayer into 2 layers (see commit message description for why this is needed)

- Pass tracing fields as headers

### DIFF
--- a/daphne_worker/src/durable/aggregate_store.rs
+++ b/daphne_worker/src/durable/aggregate_store.rs
@@ -3,10 +3,11 @@
 
 use crate::{
     config::DaphneWorkerConfig,
-    durable::{state_get_or_default, BINDING_DAP_AGGREGATE_STORE},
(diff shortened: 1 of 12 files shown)


cjpatton (approved):
I've reviewed the code movement parts and they appear to me to be unchanged. I'm uncertain that the new code is correct because I'm not familiar enough with the tracing code. It would be good to get extra eyes on this.

CI is complaining about `cargo fmt`.

PR #382, opened 2023-08-07
Use bincode for DO request serialization

mendess:
Bincode reduces the size of payloads by more than 50%

### DIFF
--- a/daphne_worker/src/durable/mod.rs
+++ b/daphne_worker/src/durable/mod.rs
@@ -16,10 +16,10 @@ use crate::{
 };
 use daphne::{messages::TaskId, DapBatchBucket, DapVersion};
 use rand::prelude::*;
-use serde::{Deserialize, Serialize};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::{cmp::min, time::Duration};
 use tracing::{info_span, warn};
-use worker::*;
+use worker::{js_sys::Uint8Array, *};
 
 pub(crate) const DURABLE_DELETE_ALL: &str = "/internal/do/delete_all";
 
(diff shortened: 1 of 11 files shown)


daphne/src/vdaf/mod.rs
```diff
@@ -65,7 +65,7 @@ pub(crate) enum VdafError {
 }
 
 /// A VDAF verification key.
-#[derive(Clone, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
```
cjpatton, line 68: Is `Debug` required for something? We have generally tried to avoid deriving it for cryptographic keys to make it harder for a programmer to accidentally leak its value in a log or something. If we need to implement it, I'd suggest providing one that avoids displaying the value.

daphne/src/vdaf/mod.rs
```diff
@@ -89,7 +89,7 @@ pub trait EarlyReportState {
 
 /// Report state during aggregation initialization after consuming the report share. This involves
 /// decryption as well a few validation steps.
-#[derive(Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
cjpatton, line 92: Similar comment here: In general the prep state is sensitive.

daphne_worker/src/durable/mod.rs
```diff
@@ -652,6 +656,12 @@ fn span_to_headers() -> Headers {
     })
 }
 
+async fn req_parse<T: DeserializeOwned>(req: &mut Request) -> Result<T> {
+    let bytes = req.bytes().await?;
+    bincode::deserialize(&bytes)
+        .map_err(|e| Error::RustError(format!("failed to deserialize bincode {e:?}")))
```
cjpatton, line 662:
```suggestion
        .map_err(|e| Error::RustError(format!("failed to deserialize bincode: {e:?}")))
```

cjpatton (changes requested):
LGTM! One change request, which is to implement `Derive` manually on types that have sensitive contents.

PR #384, opened 2023-08-09
Improve storage access patterns in ReportsProcessed DO

mendess:


daphne_worker/src/durable/reports_processed.rs
```diff
@@ -45,26 +43,56 @@ pub(crate) const DURABLE_REPORTS_PROCESSED_MARK_AGGREGATED: &str =
 /// where `<report_id>` is the hex-encoded report ID.
 #[durable_object]
 pub struct ReportsProcessed {
-    #[allow(dead_code)]
     state: State,
     env: Env,
     config: DaphneWorkerConfig,
     touched: bool,
     alarmed: bool,
 }
```
cjpatton, line 52: Please document the meaning of this constant. How did you pick 131,072? Also, is this the best name?

daphne_worker/src/durable/reports_processed.rs
```diff
@@ -185,15 +198,16 @@ impl ReportsProcessed {
...
-                let replayed_reports = try_join_all(
-                    report_ids
-                        .into_iter()
-                        .map(|report_id| self.to_checked(report_id)),
-                )
-                .await?
-                .into_iter()
-                .flatten()
-                .collect::<Vec<ReportId>>();
+                let mut seen = self.load_seen_reports().await?;
```
cjpatton, line 201: High level question: Isn't this going to get expensive if the number of reports processed gets large? Do we have a sense of how many is too many? Once we reach that point, what should we do at the protocol level?

cjpatton (changes requested):
Since we're changing the data we're storing, we need to document the new format in the commit message.

daphne_worker/src/durable/reports_processed.rs
```diff
@@ -45,26 +43,56 @@ pub(crate) const DURABLE_REPORTS_PROCESSED_MARK_AGGREGATED: &str =
...
-            .await?
-            .unwrap_or(false);
-        if processed {
-            Ok(Some(report_id))
-        } else {
-            Ok(None)
+    async fn load_seen_reports(&self) -> Result<HashSet<ReportId>> {
+        futures::stream::iter(0..)
+            .then(|i| async move {
+                state_get::<Vec<u8>>(&self.state, &format!("processed_reports/{i}"))
```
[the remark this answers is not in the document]
mendess: I already changed the prefix key, it used to be just `processed`

Cargo.toml
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 [workspace]
+resolver = "2"
```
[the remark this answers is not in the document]
mendess: This is so cargo can properly resolve the feature sets and not pull in tokio when building for wasm (since tokio doesn't support wasm)

daphne_worker/src/durable/reports_processed.rs
mendess: that's the max amount a value can have inside a Durable Object, I got it from the error message. I'll document

daphne_worker/src/durable/reports_processed.rs
mendess: I tested for more than 8k reports and it behaved just fine. But perhaps we need to measure what a real workload looks like

cjpatton: Did you test this in `wrangler dev` or on Workers?

To avoid creating a DoS vector here, we're going to want to enforce a cap on the number of reports we're willing to store in a single instance. If we overflow this, then we'll need to reject reports for which we aren't tracking replay (with [PrepareError "report_dropped"](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-05.html#section-4.5.1.2-5)). Note that the "mark aggregated" call back will have to handle the rejection logic.

Cargo.toml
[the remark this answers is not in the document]
cjpatton: Would you mind adding a comment to clarify?

PR #388, opened 2023-08-28
Maintenance

cjpatton:
* `cargo upgrade && cargo update`
* `cargo fmt --all`
* Update docker to rust 1.70


mendess (approved):
Finally cargo fmt for let else statements is here :tada:

PR #390, opened 2023-09-07
daphne_worker: Stabilize the PRF for internal use

### LINKED ISSUE #266: Stabilize the PRF we use for internal purposes
We use a PRF for a couple of internal functions:
1. Map a report to a shard of the ReportsPending and ReportsProcessed DOs
2. draft02: Generate collection job IDs

We are currently using `PrgAes128` for this purpose, which is deprecated as of VDAF-05. As of #259 we will use `PrgSha3`, which is the new default as of VDAF-05. This is a breaking change, since the generated values will change. In fact, `PrgSha3`is likely to change in backwards incompatible ways in future drafts: see https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/209.

In order to reduce the number of crypto dependencies, we were hoping to reuse something that we already need for VDAF. But since this is still a moving target, we ought to consider using something like HMAC-SHA256.

cjpatton:
Closes #266.

We currently use cSHAKE128 for deriving ReportsPending and ReportsProcessed shards and for computing the collection job ID for DAP-02. This comes from `prio::vdaf::prg::PrgSha3`. This API is not stable: the algorithm it uses depends on the VDAF version used, and we have not settled on which XOF to use for VDAF. (In VDAF-07 we have replaced cSHAKE128 with SHAKE128.) To avoid breaking changes for our internal use cases, use `ring::hmac::HMAC_SHA256` instead.

The recommended key size for this PRF is 32 bytes. Accordingly, rotate the keys in daphne_worker_test/wrangler.toml.


daphne_worker/src/config.rs
```diff
@@ -348,10 +345,10 @@ impl DaphneWorkerConfig {
...
         report_time: Time,
     ) -> String {
-        let mut shard_seed = [0; 8];
-        PrgSha3::seed_stream(&self.report_shard_key, b"report shard", report_id.as_ref())
-            .fill(&mut shard_seed);
-        let shard = u64::from_be_bytes(shard_seed) % self.report_shard_count;
+        let key = ring::hmac::Key::new(ring::hmac::HMAC_SHA256, &self.report_shard_key);
+        let tag = ring::hmac::sign(&key, report_id.as_ref());
+        let shard =
+            u64::from_be_bytes(tag.as_ref()[..8].try_into().unwrap()) % self.report_shard_count;
```
mendess, line 351: use `std::mem::size_of::<u64>()` here instead of `8` so it's really obvious why the number is picked.

mendess (approved):

daphne_worker/src/config.rs
cjpatton: Done!

PR #392, opened 2023-09-20
Archive daphne-worker{,-test}

mendess:


cjpatton (commented):
Let's wait to fix the revision we point to until we get through the DAP-07 changes. Let's just update the README for now.

PR #394, opened 2023-09-20
Reorganize daphne_worker

mendess:
Split the role implementations in such a way as to mirror the `daphne` crate and
move the router out of the root file. In the router module also split the leader
and helper implementations as well as get the test routes of the way.


daphne_worker/src/router/test_routes.rs
cjpatton, line 1: I'd suggest "internal_routes.rs"

daphne_worker/src/router/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+            let daph = ctx.data.handler(&ctx.env);
+            let req = daph.worker_request_to_dap(req, &ctx).await?;
+
+            let span = info_span_from_dap_request!("hpke_config", req);
+
+            match daph.handle_hpke_config_req(&req).instrument(span).await {
+                Ok(req) => dap_response_to_worker(req),
+                Err(e) => daph.state.dap_abort_to_worker_response(e),
+            }
+        });
```
cjpatton, line 91: Perhaps this route should go in the `aggregator` module?

cjpatton (approved):
I like it! A few minor suggestions in-line, plus a few asks here:

1. Please clean up daphne_worker_test/wrangler.toml (remove the admin token)
2. Add Copyright info to new files
3. In the future, it would be easier to review these big code movement PRs if it is made up of a bunch of commits, where in each commit, one component (say an `impl`) is moved at a time.
4. How about calling the new `dap` folder `roles`? This would line up better with the `daphne` crate.

daphne_worker/src/router/test_routes.rs
mendess: the name "internal_route" is weird because all routes are public in the sense that they are exposed by the service equally. I think test routes is more evocative of when they are enabled.

PR #395, opened 2023-09-26
Refactor and cleaup daphne and daphne_worker crates

mendess:
This is a collection of mostly unrelated commits that clean up the codebase a
bit, they should be reviewed individually instead of as a single unit.

### DIFF
--- a/daphne/src/roles/helper.rs
+++ b/daphne/src/roles/helper.rs
@@ -16,10 +16,10 @@ use crate::{
     fatal_error,
     messages::{
         constant_time_eq, AggregateShare, AggregateShareReq, AggregationJobContinueReq,
-        AggregationJobInitReq, PartialBatchSelector, ReportMetadata, TaskId, TransitionFailure,
-        TransitionVar,
+        AggregationJobInitReq, Draft02AggregationJobId, PartialBatchSelector, ReportMetadata,
+        TaskId, TransitionFailure, TransitionVar,
     },
-    metrics::DaphneRequestType,
+    metrics::{ContextualizedDaphneMetrics, DaphneRequestType},
     DapError, DapHelperState, DapHelperTransition, DapRequest, DapResource, DapResponse,
     DapTaskConfig, DapVersion, MetaAggregationJobId,
 };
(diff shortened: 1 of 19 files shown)


cjpatton (approved):
Beautiful work. Only one blocker: Let's drop ed0947898d2a6559180aa38f5555470d6ea92d84, as I strongly suspect we're going to want `MockAggregator` in the public API soon in order to support interop testing. This is especially important after we sunset `daphne_worker` (in the public repo).

daphne/src/roles/helper.rs
```diff
@@ -44,254 +44,272 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+        }
+
+        let agg_job_cont_req =
+            AggregationJobContinueReq::get_decoded_with_param(&req.version, &req.payload)
+                .map_err(|e| DapAbort::from_codec_error(e, task_id.clone()))?;
+
+        // draft02 compatibility: In draft02, the aggregation job ID is parsed from the
+        // HTTP request payload; in the latest, the aggregation job ID is parsed from the
+        // request path.
+        let agg_job_id = match (
```
oliy, line 220: Seems like some of this complex logic is duplicated and should be consolidated between init & cont?

daphne/src/roles/leader.rs
```diff
@@ -162,7 +162,16 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
             .map_err(|e| DapAbort::from_codec_error(e, task_id.clone()))?;
         debug!("report id is {}", report.report_metadata.id);
 
-        resolve_taskprov(self, task_id, req, Some(&report.report_metadata)).await?;
+        if let Some(taskprov_version) = self.get_global_config().taskprov_version {
```
oliy, line 165: if you are checking this for every use of resolve_taskprov, maybe move the check into resolve_taskprov itself?

mendess: I thought about that, the reason I kept it out is:
- [in roles/helper.rs](https://github.com/cloudflare/daphne/blob/mendess/cleanup-taskprov/daphne/src/roles/helper.rs#L61-L88) you'd have to check twice
- you can't forget to check because you need the value inside the option to call `resolve_taskprov`

PR #396, opened 2023-09-26
Reduce peak memory usage during the handling of a AggregateContitnueReq

mendess:
# Memory usage investigation

To test for this I ran an aggregation job of 80 reports of Prio2 { dimension: 100_000 }

## Legend
The method under investigation was `DapHelper::handle_agg_job_cont_req`, I added
some logs after each operation and queried the allocator for how much memory was
currently allocated. The following legend maps a numerical value to the point in
the code where the allocation size was recorded.

- 0: before calling the method
- 1: before any operations
- 2: after resolving taskprov
- 3: after getting the task config
- 4: after checking for unauthorized reason (and not returning)
- 5: after checking for a version mismatch (and not returning)
- 6: after deserializing the `AggregationJobContinueReq`
- 7: after getting `agg_job_id`
- 8: after getting helper state and cloning the `part_batch_sel`
- 9: after calling `task_config.vdaf.handle_agg_job_cont_req`
- 10: after processing the transitions and shares from `handle_agg_job_cont_req`
- 11: after returning from the method

**Note:** This investigation was done using native rust code, not wasm, a latter
wasm experiment was run as well with different parameters (due to worker
limitations) just to ensure that allocations in wasm were just as predictable as
in native, see bellow for more information on this.

## Before changes

Before making any changes, I ran this experiment, it's important to note that
the native implementation uses the `MockAggregator` which simulates storage in
memory, so we have to account for this when doing the math on the amount
allocated.

```
allocated  0: 93215
allocated  1: 94175
allocated  2: 94175
allocated  3: 94257 # helper state allocates 82 bytes
allocated  4: 94257
allocated  5: 94257
allocated  6: 100401 # the deserializing of AggregationJobContinueReq allocs 6144 bytes
allocated  7: 100401
allocated  8: 100401
allocated  9: 128104241 # the method allocates 128_003_840 bytes
                        # decomposed this is:
                        # out_shares  7040 + 128000000 = 128_007_040
                        # transitions 3840 + 0         =       3_840
                                                       = 128_010_880

allocated 10: 1699957
allocate  11: 1690007
```

After running the method the size taken up my the aggregate store can be
calculated

```
report_store processed len 80 cap 112 size in bytes 1792
agg_store              len  1 cap   3 size in bytes 1600408

  total allocated - in memory store affected by this method
= 1690007         - (1600408 + 1792)
= 87807
```

Which is less than when we started, which probably means the parameters to the
method itself have been dealloc and I probably missed something in the in memory
store like the metrics.

**Peak allocation:** 128_104_241

## After changes

After making changes I introduced the `deepsize` crate, which let me query the
size of objects at runtime after some instrumenting the code.

```
allocated  0       : 93177
deep size of helper: 11893
allocated  1       : 94329
allocated  2       : 94329
allocated  3       : 94411 # helper state allocates 82 bytes
allocated  4       : 94411
allocated  5       : 94411
allocated  6       : 100555 # the deserializing of AggregationJobContinueReq allocs 6144 bytes
allocated  7       : 100555
allocated  8       : 100555
allocated  9       : 104939 # the method allocates 4384 bytes
                            # 2192 bytes for the processed hashset
                            # 2192 bytes for the recognized hashset
allocated 10       : 1700111
allocated 11       : 1689969
deep size of helper: 1607053
```

Now using the `deepsize` crate we can see that our estimate in the previous case
of the size of the in memory store almost matches up with its size now.

Since this method aggregates each share individually instead of all of once, we
have to measure peak allocation inside the loop. Each iteration produces 3
variables:

- `transition` which takes up 48 bytes on the stack (since it's always the `Finished` variant)
- `out_share` which takes up `1_600_088` bytes

**Peak allocation:** 3_311_535 bytes

## Inside wasm

Unfortunately inside wasm, with this current implementation of the worker, we
can't test with the same parameters because the leader can't handle 80 reports
of 100_000k dimension. As such we had to run the test with much smaller
parameters. Because of this I can't really provide comparable values to the ones
gathered in the previous experiments, what I did confirm is that:

- WASM being a 32bit system means that structs are smaller on average (Vec are 12 bytes instead of 24)
- WASM allocations are as predictable as native allocations, there doesn't seem to be any weird overhead going on


daphne/src/vdaf/mod.rs
```diff
@@ -325,6 +326,21 @@ pub enum VdafPrepState {
     Prio3Field128(Prio3PrepareState<Field128, 16>),
 }
 
+#[cfg(any(test, feature = "test-utils"))]
+impl deepsize::DeepSizeOf for VdafPrepState {
+    fn deep_size_of_children(&self, _context: &mut deepsize::Context) -> usize {
+        // This method is, as documented, an estimation of the size of the children. Since it can't
+        // be known for this type due to it's encapsulation, I will count the size of it as 0.
+        //
+        // TODO: Make a better estimation.
```
cjpatton, line 335: We would need to know (1) which VDAF is in use and (2) whether we're the Leader or the Helper. How does `deepsize::Context` work? Could we use it to plumb what we need here?

For what it's worth, for Prio2 and prio3 the size is constant, as we're using the `Share::Helper` variant in all cases.

daphne/src/vdaf/mod.rs
```diff
@@ -1009,89 +1036,87 @@ impl VdafConfig {
...
+                    let res = match self {
+                        Self::Prio3(prio3_config) => {
+                            prio3_helper_prepare_finish(prio3_config, helper_step, leader_message)
+                        }
+                        Self::Prio2 { dimension } => {
+                            prio2_helper_prepare_finish(*dimension, helper_step, leader_message)
+                        }
+                    };
+
+                    let (var, share) = match res {
```
cjpatton, line 1088:
```suggestion
                    let (var, out_share) = match res {
```

daphne/src/roles/helper.rs
```diff
@@ -241,39 +243,38 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+        let mut transitions = Vec::with_capacity(num_reports);
+        let mut out_shares_count = 0;
+
+        for transition_and_share in transitions_and_shares {
+            let (mut transition, share) = transition_and_share?;
+            if let Some(share) = share {
+                out_shares_count += 1;
                 let replayed = self
-                    .put_out_shares(task_id, task_config, &part_batch_sel, out_shares)
+                    .put_out_shares(task_id, task_config, &part_batch_sel, vec![share])
```
cjpatton, line 264: Hmm, this doesn't look like it'll be very efficient for `daphne_worker`. What about `put_out_shares()` an iterator that we filter here (`replayed.contains(&transiition.report_id`)

daphne/src/vdaf/mod.rs
```diff
@@ -964,14 +991,17 @@ impl VdafConfig {
...
+        task_id: &'s TaskId,
+        agg_job_id: &'s MetaAggregationJobId,
         state: DapHelperState,
-        agg_cont_req: &AggregationJobContinueReq,
-        metrics: &ContextualizedDaphneMetrics<'_>,
-    ) -> Result<DapHelperTransition<AggregationJobResp>, DapAbort> {
+        agg_cont_req: &'s AggregationJobContinueReq,
+        metrics: &'s ContextualizedDaphneMetrics<'s>,
+    ) -> Result<
+        impl Iterator<Item = Result<(Transition, Option<DapOutputShare>), DapAbort>> + 's,
```
cjpatton, line 1002: * Are we handling the `DapAbort` properly? If the result of `handle_agg_job_cont_req()` is `Err`, then what we actually want to abort the request.
* An alternative to iterating over `Option<DapOutputShare>` is to merge all of the `DapOutputShare`s returned by this function into a single `DapAggregateShare` and return that isntead.

cjpatton (commented):
Thanks for getting this started! High level questions:

1. Which parts of this change do we intend to keep? Do we want to merge the PR with all of the instrumentation here? I think it would be fine as long as it doesn't mess with APIs (it doesn't seem to, although I noticed that `cargo test` fails in directory `daphne/src/`) and doesn't create unnecessary overhead.
2. Passing one output share to `put_out_shares()` at a time is likely going to be really inefficient for `daphne_worker`. I thin there are better alternatives (see inline comments).
3. We need to double check that we're still handling DAP aborts properly.

daphne/src/hpke.rs
```diff
@@ -155,6 +158,16 @@ pub struct HpkeConfig {
     pub public_key: HpkePublicKey,
 }
 
+#[cfg(any(test, feature = "test-utils"))]
```
oliy, line 161: Not super familiar with deepsize crate, but I assume its derive macro should automatically provide a proper implementation of this?  why does this need manual derivation, where we might forget to update this with changes in the future?

mendess: We can only derive when all the members of a struct implement the trait, `HpkePublicKey` is a foreign type so we can't implement this foreign trait on it, hence we need to implement by hand.

daphne/src/vdaf/mod.rs
mendess: The size being constant for the helper means this function is correct for the helper case, as this estimation is supposed to only account for heap allocations/stuff behind pointers.

`deepsize::Context` is an opaque type, it only has functionality for the `deepsize` crate, we can't use it for anything

cjpatton (commented):
Alrighty, this is looking to be on the right track. Once more, great catch on the retry thing!

I've read up to `daphne/src/roles/helepr.rs`. There's a comment there I'd like you to address before I get any deeper.

daphne/src/roles/helper.rs
```diff
@@ -193,50 +193,57 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
-            &agg_job_cont_req,
-            &metrics,
-        )?;
-
-        let (agg_job_resp, out_shares_count) = match transition {
-            DapHelperTransition::Continue(..) => {
-                return Err(fatal_error!(err = "unexpected transition (continued)").into());
+            })?;
+        let (out_shares_count, transitions) = loop {
+            if agg_job_cont_req.transitions.iter().all(is_replayed) {
```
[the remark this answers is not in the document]
mendess: I find the implementation of is_replayed is too verbose to be inline in the closure, this way the if statement reads more like english "if agg_job_cont_req transitions are all replayed". Hence why I did it this way

daphne/src/roles/helper.rs
```diff
@@ -193,50 +193,57 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+                )?;
+
+            let out_shares_count = u64::try_from(dap_aggregate_share_span.report_count()).unwrap();
+            let replayed = self
+                .put_out_shares(task_id, task_config, dap_aggregate_share_span)
+                .await?;
+
+            if let Some(replayed) = replayed {
+                // remove replayed reports and try again
+                for t in &mut agg_job_cont_req.transitions {
```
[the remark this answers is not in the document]
mendess: if you look at what I did in `handle_agg_job_cont_req` you'll see that I modified it accordingly :wink: and you'll also see that I didn't like doing it. Adding this information to state doesn't make much sense imo, this is transient state of this retry algorithm, `state` outlives it. I tried many approaches and all of them were just as bad as this if not worse.

Effectively I'm using `agg_job_cont_req.transitions` as a work vector for the algorithm

daphne/src/roles/helper.rs
```diff
@@ -193,50 +193,57 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
-            &metrics,
-        )?;
-
-        let (agg_job_resp, out_shares_count) = match transition {
-            DapHelperTransition::Continue(..) => {
-                return Err(fatal_error!(err = "unexpected transition (continued)").into());
+            })?;
+        let (out_shares_count, transitions) = loop {
+            if agg_job_cont_req.transitions.iter().all(is_replayed) {
+                break (0, agg_job_cont_req.transitions);
```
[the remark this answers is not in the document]
mendess: See my [other comment](https://github.com/cloudflare/daphne/pull/396#discussion_r1348571269) about how this `transitions` vector is being used as a "work vector" and not just as the immutable input from the leader.

daphne/src/vdaf/mod.rs
```diff
@@ -964,42 +991,45 @@ impl VdafConfig {
...
+        task_config: &DapTaskConfig,
+        state: &DapHelperState,
+        is_replay: impl Fn(&ReportId) -> bool,
+        agg_job_id: &MetaAggregationJobId<'_>,
+        agg_job_cont_req: &AggregationJobContinueReq,
         metrics: &ContextualizedDaphneMetrics<'_>,
-    ) -> Result<DapHelperTransition<AggregationJobResp>, DapAbort> {
-        if let Some(round) = agg_cont_req.round {
-            if round == 0 {
+    ) -> Result<(DapAggregateShareSpan, Vec<Transition>), DapAbort> {
```
cjpatton, line 1004: For consistency with the other methods, let's return an `AggregationJobResp` instead of a `Vec<Transition>`

daphne/src/vdaf/mod.rs
```diff
@@ -1814,39 +1852,27 @@ mod test {
             .handle_agg_job_resp(leader_state, agg_job_resp)
             .unwrap_uncommitted();
 
-        let (helper_out_shares, agg_job_resp) = t
-            .handle_agg_job_cont_req(helper_state, &agg_job_cont_req)
-            .unwrap_finish();
-        assert_eq!(helper_out_shares.len(), 5);
-        assert_eq!(agg_job_resp.transitions.len(), 5);
+        let (helper_share_span, transitions) =
```
cjpatton, line 1855: Here and below.
```suggestion
        let (helper_agg_share_span, transitions) =
```

daphne_worker/src/durable/reports_processed.rs
```diff
@@ -53,18 +54,66 @@ pub struct ReportsProcessed {
...
-        let key = format!("processed/{}", report_id.to_hex());
-        let processed: bool = state_set_if_not_exists(&self.state, &key, &true)
-            .await?
-            .unwrap_or(false);
-        if processed {
-            Ok(Some(report_id))
-        } else {
-            Ok(None)
+#[derive(Debug, Clone)]
+struct ReportIdKey<'s>(&'s ReportId, String);
```
cjpatton, line 58:
```suggestion
struct ReportIdKey<'s>(&'s ReportId, String);

```

daphne_worker/src/durable/reports_processed.rs
```diff
@@ -53,18 +54,66 @@ pub struct ReportsProcessed {
...
+fn report_id_to_do_key(id: &ReportId) -> ReportIdKey<'_> {
```
cjpatton, line 59: `impl<'r> From<&'r ReportId> for ReportIdKey<'r>`?

daphne_worker/src/durable/reports_processed.rs
```diff
@@ -53,18 +54,66 @@ pub struct ReportsProcessed {
...
+}
+
+impl<'r> Default for CheckedReplays<'r> {
+    fn default() -> Self {
+        Self::NotReplayed(vec![])
+    }
+}
+
+impl<'r> CheckedReplays<'r> {
+    pub fn add_replay(mut self, id: &'r ReportId) -> Self {
```
cjpatton, line 76: Here and below.
```suggestion
    pub(crate) fn add_replay(mut self, id: &'r ReportId) -> Self {
```

daphne_worker/src/durable/reports_processed.rs
```diff
@@ -53,18 +54,66 @@ pub struct ReportsProcessed {
...
+        match &mut self {
+            Self::Replayed(r) => {
+                r.push(id);
+                self
+            }
+            Self::NotReplayed(_) => Self::Replayed(vec![id]),
+        }
+    }
+
+    pub fn add_not_replayed(mut self, id: ReportIdKey<'r>) -> Self {
```
cjpatton, line 86:
```suggestion
    pub fn add_fresh(mut self, id: ReportIdKey<'r>) -> Self {
```

daphne_worker/src/durable/reports_processed.rs
```diff
@@ -53,18 +54,66 @@ pub struct ReportsProcessed {
...
+    async fn check_replays<'s>(&self, report_ids: &'s [ReportId]) -> Result<CheckedReplays<'s>> {
+        futures::stream::iter(report_ids.iter().map(report_id_to_do_key))
+            .then(|id| {
+                let state = &self.state;
+                async move {
+                    state_get::<bool>(state, &id.1)
+                        .await
+                        .map(|presence| match presence {
+                            // if it's present then it's a replay
+                            Some(_) => Err(id.0),
```
cjpatton, line 105: Suggestion: I'd be a bit stricter here. Maybe handle `Some(false)` the same way as `None`?
```suggestion
                            Some(true) => Err(id.0),
```

daphne_worker/src/durable/reports_processed.rs
```diff
@@ -178,24 +227,29 @@ impl ReportsProcessed {
...
-                        .into_iter()
-                        .map(|report_id| self.to_checked(report_id)),
-                )
-                .await?
-                .into_iter()
-                .flatten()
-                .collect::<Vec<ReportId>>();
+                match self.check_replays(&report_ids).await? {
+                    CheckedReplays::Replayed(replayed) => Response::from_json(&replayed),
+                    CheckedReplays::NotReplayed(not_replayed) => {
```
cjpatton, line 241: "Replay" is used a lot here. I think using names that are more semantically meaningful here would be helpfulo.
```suggestion
                    CheckReplayResult::SomeReplayed(report_ids) => Response::from_json(&report_ids),
                    CheckReplayResult::AllFresh(report_ids) => {
```

daphne_worker/src/roles/aggregator.rs
```diff
@@ -411,40 +412,38 @@ impl<'srv> DapAggregator<DaphneWorkerAuth> for DaphneWorker<'srv> {
...
 
-    async fn put_out_shares(
+    async fn try_put_agg_share_span(
         &self,
         task_id: &TaskId,
         task_config: &DapTaskConfig,
-        part_batch_sel: &PartialBatchSelector,
-        out_shares: Vec<DapOutputShare>,
-    ) -> std::result::Result<HashSet<ReportId>, DapError> {
+        aggregate_shares: DapAggregateShareSpan,
```
cjpatton, line 419: Consistency
```suggestion
        agg_share_span: DapAggregateShareSpan,
```

daphne_worker/src/roles/aggregator.rs
```diff
@@ -465,31 +464,28 @@ impl<'srv> DapAggregator<DaphneWorkerAuth> for DaphneWorker<'srv> {
...
-                let agg_share = DapAggregateShare::try_from_out_shares(
-                    out_shares
-                        .into_iter()
-                        .filter(|out_share| !replayed.contains(&out_share.report_id)),
-                )?;
-
-                std::result::Result::<_, DapError>::Ok(
+        // Only aggregate the output shares if none are replayed
+        if replayed.is_empty() {
+            // TODO: don't produce a vec
```
cjpatton, line 469: When you add a TODO like this, it's useful to add your GitHub handle so that others can follow up with you about what you meant.

daphne_worker/src/roles/aggregator.rs
```diff
@@ -465,31 +464,28 @@ impl<'srv> DapAggregator<DaphneWorkerAuth> for DaphneWorker<'srv> {
         .flatten()
         .collect::<HashSet<ReportId>>();
```
cjpatton, line 466: Suppose there are two DO calls made: the first detects no replays and thus marks all reports as aggregated, but another detects replays and so returns the subset that were replayed. What happens?

If i understand the code correctly, `replayed` will be non-empty, so we'll return `Ok(Some(replayed))`. When we retry, the reports that were marked aggregated on the first attempt will be considred replays on the second try. Do I have that right?

If so, I think we need to separate `MARK_AGGREGATED` into two phases. In the first phase we check for replays; in the second we either commit all the state changes or none of them.

cjpatton (commented):
Mostly nit-picky commnets left: everything looks correct modulo what might be a bug in `DaphneWorker`'s implementation of `try_put_agg_share_span()`.

daphne/src/roles/helper.rs
```diff
@@ -199,44 +199,50 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+        //
+        // - try to aggregate the output shares into an `DapAggregateShareSpan`
+        // - pass it to `try_put_agg_share_span`
+        //   - if replays are found, remove them from the `agg_job_cont_req` and drop all work that
+        //     was done and try again (now without replays)
+        //   - else break with the finished (of failed) transitions
+        //
+        // The reason we do this is because we don't expect replays to happen but we have to guard
+        // against them, as such, even though retrying is possibly very expensive, it probably
+        // won't happen often enough that it matters.
```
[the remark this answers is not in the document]
mendess: If the leader is misbehaving isn't the availability of the system in general already compromised?

daphne/src/roles/leader.rs
```diff
@@ -424,18 +424,21 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
...
+            uncommited,
+            agg_job_resp,
+            &metrics,
+        )?;
+        let out_shares_count = agg_share_span.report_count() as u64;
 
         // At this point we're committed to aggregating the reports: if we do detect a report was
         // replayed at this stage, then we may end up with a batch mismatch. However, this should
         // only happen if there are multiple aggregation jobs in-flight that include the same
         // report.
```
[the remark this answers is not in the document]
mendess: I kept the behavior of the old leader, but we can change it to a panic

daphne_worker/src/durable/reports_processed.rs
mendess: Can actually just not be pub

daphne/src/roles/helper.rs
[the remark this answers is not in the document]
cjpatton: Imagine Daphne has several (or 100s of customers): if one of them misbehaves, then we need to do what we can to make sure it doesn't impact other customers.

daphne/src/roles/leader.rs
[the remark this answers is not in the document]
cjpatton: Got it. It seems reasonable to panic here.

daphne_worker/src/durable/reports_processed.rs
cjpatton: SG.

daphne_worker/src/roles/aggregator.rs
mendess: No, mark aggregated is an all or nothing call now. Hence that complex type in the `ReportsProcessed` durable object. Either no reports are fresh or nothing is considered aggregated. This is a change from the old behavior of this call

daphne_worker/src/durable/reports_processed.rs
cjpatton: Nice idea.

daphne/src/roles/helper.rs
```diff
@@ -199,44 +198,47 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
-                        if replayed.contains(&transition.report_id) {
-                            let failure = TransitionFailure::ReportReplayed;
-                            transition.var = TransitionVar::Failed(failure);
-                            metrics.report_inc_by(&format!("rejected_{failure}"), 1);
-                        }
-                    }
-                }
-
-                (agg_job_resp, out_shares_count)
+                let Some(replayed) = replayed else {
```
cjpatton, line 237: readability: I'd probably use `if let Some(...) { ... } else { ... }` to make the branching a littler more clear.

daphne_worker/src/roles/aggregator.rs
cjpatton: Per discussion, let's resolve this for now by just leaving a TODO noting the bug. (Since it exists before this PR.)

daphne_worker/src/durable/reports_processed.rs
```diff
@@ -130,13 +182,13 @@ impl ReportsProcessed {
                         .iter()
                         .filter(|consumed_report| consumed_report.is_ready())
                         .map(|consumed_report| async {
-                            if let Some(replayed) = state_get::<bool>(
+                            if let Some(exits) = state_get::<bool>(
```
cjpatton, line 185:
```suggestion
                            if let Some(exists) = state_get::<bool>(
```

daphne_worker/src/durable/reports_processed.rs
```diff
@@ -53,18 +54,69 @@ pub struct ReportsProcessed {
...
+    fn from(id: &'id ReportId) -> Self {
+        ReportIdKey(id, format!("processed/{}", id.to_hex()))
+    }
+}
+
+#[derive(Debug)]
+enum CheckedReplays<'s> {
+    SomeReplayed(Vec<&'s ReportId>),
+    AllFresh(Vec<ReportIdKey<'s>>),
+}
```
cjpatton, line 70: I believe this works as an alternative `Default` implementation
```suggestion
#[derive(Debug, Default)]
enum CheckedReplays<'s> {
    SomeReplayed(Vec<&'s ReportId>),
    #[default]
    AllFresh(Vec<ReportIdKey<'s>>),
}
```

daphne/src/roles/aggregator.rs
```diff
@@ -106,17 +106,22 @@ pub trait DapAggregator<S>: HpkeDecrypter + DapReportInitializer + Sized {
...
+    /// - `Ok(Some(set))` if at least one report was a replay. This also means no aggregate shares where merged.
+    /// - `Err(err)` if an error occurred.
+    async fn try_put_agg_share_span(
         &self,
         task_id: &TaskId,
         task_config: &DapTaskConfig,
-        part_batch_sel: &PartialBatchSelector,
-        out_shares: Vec<DapOutputShare>,
-    ) -> Result<HashSet<ReportId>, DapError>;
+        out_shares: DapAggregateShareSpan,
```
cjpatton, line 123:
```suggestion
        agg_share_span: DapAggregateShareSpan,
```

daphne_worker/src/roles/aggregator.rs
```diff
@@ -410,40 +413,38 @@ impl<'srv> DapAggregator<DaphneWorkerAuth> for DaphneWorker<'srv> {
...
                 agg_store_request_data
-                    .entry(agg_store_name)
-                    .or_default()
-                    .push(out_share);
-            }
+                    .insert(agg_store_name, aggregate_share)
+                    .is_none(),
+                "agg_store_names should be unique"
+            );
         }
```
cjpatton, line 449:
```suggestion
// TODO(mendess) <Note the bug we found here (Either all DO requests must return "no replays" or no DO requests commit "mark aggregated". We need to make sure these events are mutually exclusive.)
```

daphne_worker/src/roles/aggregator.rs
```diff
@@ -410,40 +413,38 @@ impl<'srv> DapAggregator<DaphneWorkerAuth> for DaphneWorker<'srv> {
...
+            debug_assert!(
                 agg_store_request_data
-                    .entry(agg_store_name)
-                    .or_default()
-                    .push(out_share);
-            }
+                    .insert(agg_store_name, aggregate_share)
+                    .is_none(),
+                "agg_store_names should be unique"
+            );
```
cjpatton, line 447: We don't expect the agg_store_names to be unique. I think what we want to do is merge `agg_share` into the request data.

daphne_worker/src/durable/reports_processed.rs
mendess: sadly it doesn't work, it only works for Unit variants

cjpatton: :(

cjpatton (approved):

PR #397, opened 2023-10-04
Change aggregation job size metric

mendess:


cjpatton (commented):
Looks good to me, but the commit message needs context before merging. (Why are we changing this?)

cjpatton (approved):

PR #407, opened 2023-10-12
Shard aggregate store in order to not reach durable object value size limits

mendess:


cjpatton (commented):
Now that this is in the public repo, this can do with more code comments. For starters, I don't see any documentation about what limits we're working around.

daphne_worker/src/durable/aggregate_store.rs
```diff
@@ -47,6 +53,137 @@ pub struct AggregateStore {
...
+            let kind = meta.kind.expect("if there is data there should be a type");
+
+            fn from_slice<T: FieldElement>(chunks: &[u8]) -> Result<AggregateShare<T>> {
+                let share = T::byte_slice_into_vec(chunks).map_err(|e| {
+                    worker::Error::Internal(
+                        serde_wasm_bindgen::to_value(&e.to_string())
+                            .expect("string never fails to convert to JsValue"),
+                    )
+                })?;
+                // TODO: this an abuse of this API, this type should not be constructed this way.
```
cjpatton, line 172: Please file an issue as well. We should make sure libprio-rs supports our use case here.
```suggestion
                // TODO(mendess): this an abuse of this API, this type should not be constructed this way.
```

cjpatton (commented):
It looks like the docucomment for the `AggregateStore` needs to be updated. In particular, the schema has changed, correct?

cjpatton (approved):
🦆  - go ahead and merge once CI is happy.

Issue #408, opened 2023-10-12
Report metrics are over-counted if `try_put_agg_share_span()` is retried


cjpatton:
Suppose `handle_agg_job_cont_req()` marks a report rejected due to VDAF prep failure and increments the corresponding prometheus metric. If, on the subsequent call to `try_put_agg_share_span()`, another report is marked as replayed, then the first report will be marked rejected once more and the prometheus metric will be incremented again.

cjpatton:
One idea to address this: Dump the retry logic. Split up `try_put_agg_share_span()` so we can do the following:
* replays := subset of reports in the job that exist in ReportsProcessed
* agg_share_span, agg_job_resp := handle_agg_job_cont_req(state, agg_job_cont_req, replays)
* commit agg_share_span to AggregateStore
* commit replays to ReportsProcessed

cjpatton:
This doesn't work because it leads to a race condition between DO transactions across aggregation jobs. @mendess  suggestion is to just move the metrics out of the protocol logic (vdaf/mod.rs) to roles logic.

PR #409, opened 2023-10-13
Implement the aggregation flow as in draft07

cjpatton:
Stacked on #412.
Partially addresses #350.

In draft07, the Leader sends its prep share in the
AggregationJobInitReq; the Helper computes its prep share, computes the
prep message, and commits. For 1-round VDAFs, the flow is complete after
one request.

The main goal of this change is to implement this new control flow.
Other changes:

* Clean up and unify data structures for aggregation job state.

* Rename the `aggregation_job_continue_repeats_due_to_replays` metric
  (we need to do an analogous thing during AggregationJobInitReq
  processing).

* Regression: Remove the `handle_agg_job_req_init_expired_task` tests.
  Ideally we'd like to keep these, but it's difficult to do so without
  major refactoring.

Work-in-progress: The code is not yet wire-compatible with draft07.

### DIFF
--- a/daphne/src/messages/mod.rs
+++ b/daphne/src/messages/mod.rs
@@ -412,6 +412,45 @@ impl TryFrom<Query> for BatchSelector {
     }
 }
 
+/// The PrepareInit message consisting of the report share and the Leader's initial prep share.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PrepareInit {
+    pub report_share: ReportShare,
+    pub draft07_payload: Option<Vec<u8>>,
+}
+
+impl ParameterizedEncode<DapVersion> for PrepareInit {
+    fn encode_with_param(&self, version: &DapVersion, bytes: &mut Vec<u8>) {
+        self.report_share.encode_with_param(version, bytes);
+        match (version, &self.draft07_payload) {
+            (DapVersion::Draft02, None) => (),
+            (DapVersion::Draft07, Some(payload)) => {
+                encode_u32_bytes(bytes, payload);
+            }
+            _ => unreachable!("unhandled version {version:?}"),
+        }
+    }
+}
+
+impl ParameterizedDecode<DapVersion> for PrepareInit {
+    fn decode_with_param(
+        version: &DapVersion,
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        let report_share = ReportShare::decode_with_param(version, bytes)?;
+        let draft07_payload = match version {
+            DapVersion::Draft02 => None,
+            DapVersion::Draft07 => Some(decode_u32_bytes(bytes)?),
+            _ => unreachable!("unhandled version {version:?}"),
+        };
+
+        Ok(Self {
+            report_share,
+            draft07_payload,
+        })
+    }
+}
+
 /// Aggregate initialization request.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct AggregationJobInitReq {
@@ -419,7 +458,7 @@ pub struct AggregationJobInitReq {
     pub draft02_agg_job_id: Option<Draft02AggregationJobId>, // Set in draft02
     pub agg_param: Vec<u8>,
     pub part_batch_sel: PartialBatchSelector,
-    pub report_shares: Vec<ReportShare>,
+    pub prep_inits: Vec<PrepareInit>,
 }
 
 impl ParameterizedEncode<DapVersion> for AggregationJobInitReq {
@@ -440,7 +479,7 @@ impl ParameterizedEncode<DapVersion> for AggregationJobInitReq {
             DapVersion::Unknown => unreachable!("unhandled version {version:?}"),
         };
         self.part_batch_sel.encode(bytes);
-        encode_u32_items(bytes, version, &self.report_shares);
+        encode_u32_items(bytes, version, &self.prep_inits);
     }
 }
 
@@ -464,7 +503,7 @@ impl ParameterizedDecode<DapVersion> for AggregationJobInitReq {
             draft02_agg_job_id,
             agg_param,
             part_batch_sel: PartialBatchSelector::decode(bytes)?,
-            report_shares: decode_u32_items(version, bytes)?,
+            prep_inits: decode_u32_items(version, bytes)?,
         })
     }
 }
@@ -1293,32 +1332,38 @@ mod test {
                 part_batch_sel: PartialBatchSelector::FixedSizeByBatchId {
                     batch_id: BatchId([0; 32]),
                 },
-                report_shares: vec![
-                    ReportShare {
-                        report_metadata: ReportMetadata {
-                            id: ReportId([99; 16]),
-                            time: 1637361337,
-                            extensions: Vec::default(),
+                prep_inits: vec![
+                    PrepareInit {
+                        report_share: ReportShare {
+                            report_metadata: ReportMetadata {
+                                id: ReportId([99; 16]),
+                                time: 1637361337,
+                                extensions: Vec::default(),
+                            },
+                            public_share: b"public share".to_vec(),
+                            encrypted_input_share: HpkeCiphertext {
+                                config_id: 23,
+                                enc: b"encapsulated key".to_vec(),
+                                payload: b"ciphertext".to_vec(),
+                            },
                         },
-                        public_share: b"public share".to_vec(),
-                        encrypted_input_share: HpkeCiphertext {
-                            config_id: 23,
-                            enc: b"encapsulated key".to_vec(),
-                            payload: b"ciphertext".to_vec(),
-                        },
+                        draft07_payload: None,
                     },
-                    ReportShare {
-                        report_metadata: ReportMetadata {
-                            id: ReportId([17; 16]),
-                            time: 163736423,
-                            extensions: Vec::default(),
+                    PrepareInit {
+                        report_share: ReportShare {
+                            report_metadata: ReportMetadata {
+                                id: ReportId([17; 16]),
+                                time: 163736423,
+                                extensions: Vec::default(),
+                            },
+                            public_share: b"public share".to_vec(),
+                            encrypted_input_share: HpkeCiphertext {
+                                config_id: 0,
+                                enc: vec![],
+                                payload: b"ciphertext".to_vec(),
+                            },
                         },
-                        public_share: b"public share".to_vec(),
-                        encrypted_input_share: HpkeCiphertext {
-                            config_id: 0,
-                            enc: vec![],
-                            payload: b"ciphertext".to_vec(),
-                        },
+                        draft07_payload: None,
                     },
                 ],
             },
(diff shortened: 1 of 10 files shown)


daphne/src/roles/mod.rs
```diff
@@ -879,56 +864,6 @@ mod test {
         assert_eq!(t.helper.audit_log.invocations(), 1);
     }
 
-    async_test_versions! { handle_agg_job_req_bad_round }
-
-    // Test that the Helper rejects reports with a bad round id
```
cjpatton, line 884: Reviewer note: This test does not apply to draft02 and no longer applies to the latest version.

daphne/src/roles/helper.rs
```diff
@@ -140,22 +147,71 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
                 agg_job_resp
             }
-            DapHelperTransition::Finish(..) => {
-                return Err(fatal_error!(err = "unexpected transition (finished)").into());
+            DapHelperAggregationJobTransition::Finished(
+                mut maybe_agg_share_span,
+                mut agg_job_resp,
+            ) => {
+                // XXX This loop is messier than I'd like, but should be cleaned up after
+                // https://github.com/cloudflare/daphne/issues/408.
```
cjpatton, line 158: Reviewer note: Before cleaning this up, I'd like to see where we land on this #408.

daphne/src/roles/helper.rs
```diff
@@ -458,42 +514,33 @@ mod tests {
...
     use crate::MetaAggregationJobId;
     use crate::{roles::test::TestData, DapVersion};
 
     #[tokio::test]
-    async fn replay_reports_when_continuing_aggregation() {
+    async fn replay_reports_when_continuing_aggregation_draft02() {
         let mut data = TestData::new(DapVersion::Draft02);
         let task_id = data.insert_task(
             DapVersion::Draft02,
-            crate::VdafConfig::Prio2 { dimension: 10 },
```
cjpatton, line 473: Reviewer note: The VDAF declared here needs to be compatible with the measurement generated by `test.gen_test_report()`. This is Prio3Count, not Prio2.

daphne/src/messages/mod.rs
```diff
@@ -412,14 +412,53 @@ impl TryFrom<Query> for BatchSelector {
     }
 }
 
+/// The PrepareInit message consisting of the report share and the Leader's initial prep share.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PrepareInit {
+    pub report_share: ReportShare,
+    pub draft05_payload: Option<Vec<u8>>,
```
mendess, line 419: draft_05?

daphne/src/roles/mod.rs
```diff
@@ -804,51 +849,20 @@ mod test {
...
+    async_test_versions! { handle_agg_job_init_req_unauthorized_request }
 
     // Test that the Helper rejects reports with a bad round number.
-    async fn handle_agg_job_req_bad_round(version: DapVersion) {
-        let t = Test::new(version);
-        if version == DapVersion::Draft02 {
+    #[tokio::test]
+    async fn handle_agg_job_req_bad_round_draft02() {
+        let t = Test::new(DapVersion::Draft02);
+        if DapVersion::Draft02 == DapVersion::Draft02 {
```
mendess, line 858: this check seems rather silly :thinking:

daphne/src/roles/helper.rs
```diff
@@ -201,16 +247,32 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+                    task_config,
+                    &state,
+                    report_status,
+                    &agg_job_id,
+                    &agg_job_cont_req,
+                )
+            })
+            .await?;
+
+        for transition in agg_job_resp.transitions.iter() {
```
cjpatton, line 263: Reviewer note: Moved this out of `finish_agg_job_and_aggregate()` to avoid double-counting.

daphne/src/roles/helper.rs
```diff
@@ -201,16 +247,32 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+            if let TransitionVar::Failed(failure) = &transition.var {
+                metrics.report_inc_by(&format!("rejected_{failure}"), 1)
+            }
+        }
+
```
cjpatton, line 268: Reviewer note: I'm fairly certain we accidently deleted this previously.

daphne/src/roles/leader.rs
```diff
@@ -346,9 +346,16 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
...
             .await?;
         let (state, agg_job_init_req) = match transition {
-            DapLeaderTransition::Continue(state, agg_job_init_req) => (state, agg_job_init_req),
-            DapLeaderTransition::Skip => return Ok(0),
-            DapLeaderTransition::Uncommitted(..) => {
+            DapLeaderAggregationJobTransition::Continued(state, agg_job_init_req) => {
+                (state, agg_job_init_req)
+            }
+            DapLeaderAggregationJobTransition::Finished(agg_span)
+                if agg_span.report_count() == 0 =>
```
cjpatton, line 353: I removed `Skip` because it feels a bit redundant. But the fact that we need this `if` statement here makes me wonder if we really want an empty agg span to distinguish between these branches.

daphne/src/roles/leader.rs
```diff
@@ -385,59 +392,67 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
...
+                )?
             }
-            DapLeaderTransition::Skip => return Ok(0),
-            DapLeaderTransition::Continue(..) => {
+            DapLeaderAggregationJobTransition::Finished(agg_span) => {
+                if agg_span.report_count() > 0 {
+                    agg_span
+                } else {
+                    return Ok(0);
+                }
```
cjpatton, line 439: Similar comment here.

daphne/src/roles/leader.rs
```diff
@@ -385,59 +392,67 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
...
-        )?;
-        let out_shares_count = agg_share_span.report_count() as u64;
+        let out_shares_count = agg_span.report_count() as u64;
 
         // At this point we're committed to aggregating the reports: if we do detect a report was
         // replayed at this stage, then we may end up with a batch mismatch. However, this should
         // only happen if there are multiple aggregation jobs in-flight that include the same
         // report.
 
+        // TODO(cjaptton) Treat a failure here as fatal.
```
cjpatton, line 453: Let me know if you agree and I'll file an issue so we don't forget.

daphne/src/vdaf/mod.rs
```diff
@@ -839,52 +850,137 @@ impl VdafConfig {
...
 
-                EarlyReportStateInitialized::Rejected { metadata, failure } => {
-                    metrics.report_inc_by(&format!("rejected_{failure}"), 1);
-                    Transition {
-                        report_id: metadata.into_owned().id,
-                        var: TransitionVar::Failed(failure),
+    /// Consume an initial aggregate request from the Leader. The outputs are the Helper's state
+    /// for the aggregation flow and the aggregate response to send to the Leader. This method is
+    /// run by the Helper.
+    pub(crate) fn handle_agg_job_init_req(
```
mendess, line 859: I think we should just split this function in two, one for each version of the protocol, even if that means having a bit of repeated code. Otherwise they get incredibly long and it's harder to follow when debugging because half the code doesn't matter (usually when debugging you only care about one version of the draft). It also makes it easier to rip out draft2 later and makes it easier to compare the 2 drafts.

daphne/src/vdaf/mod.rs
```diff
@@ -932,67 +1028,107 @@ impl VdafConfig {
...
-                    helper_message,
-                ),
-                Self::Prio2 { dimension } => prio2_prep_finish_from_shares(
-                    *dimension,
-                    leader_step,
-                    leader_message,
-                    helper_message,
-                ),
-            };
+            match task_config.version {
```
mendess, line 1031: Same thing about splitting into two functions here, but since there is quite a bit of common code before this point, you can split into two inner functions (inside this function) that way it doesn't get lost at the bottom of the file

mendess (changes requested):

daphne/src/roles/leader.rs
mendess: I think we should either make it fatal now (in this PR) or not at all, opening an issue pointing to a part of the code that is incomplete and not used in prod (leader) I think is counterproductive. It's not something we need to fix yet and it's likely something we don't know how to fix yet. It's just gonna "rot away" in the issues tab

daphne/src/messages/mod.rs
mendess: My question is more "why draft 5" and not "7", this is now the only mention of draft 5 in the code

daphne/src/roles/leader.rs
mendess: why do we need this if?

daphne/src/messages/mod.rs
cjpatton: 😮‍💨 Yeah that was just my dumb brain. Should be 7 -- fixed.

daphne/src/roles/leader.rs
cjpatton: If the agg span is empty, then we quite the agg job. Otherwise, if the agg span is non-empty, we send the next request for the agg job.

daphne/src/roles/leader.rs
cjpatton: Done!

daphne/src/vdaf/mod.rs
```diff
@@ -384,13 +386,25 @@ impl<'a> ParameterizedDecode<(&'a VdafConfig, bool /* is_leader */)> for VdafPre
...
     Prio3ShareField128(Prio3PrepareShare<Field128, 16>),
 }
 
+#[cfg(any(test, feature = "test-utils"))]
+impl deepsize::DeepSizeOf for VdafPrepMessage {
+    fn deep_size_of_children(&self, _context: &mut deepsize::Context) -> usize {
+        match self {
+            Self::Prio2Share(msg) => std::mem::size_of_val(msg),
+            Self::Prio3ShareField64(msg) => std::mem::size_of_val(msg),
+            Self::Prio3ShareField128(msg) => std::mem::size_of_val(msg),
```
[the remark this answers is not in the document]
cjpatton: Fixed, please take another look to make sure I've got this right (I don't fully understand the requirements).

daphne/src/vdaf/mod.rs
cjpatton: Great idea, done.

daphne/src/vdaf/mod.rs
cjpatton: Done. I ended up finding a bug above which, after fixing, result in less duplicated code. I decided to split the function by version just as I did above.

mendess (approved):

PR #410, opened 2023-10-13
Move static labels to Registry to clean up Prometheus metrics.

oliy:


cjpatton (approved):
One suggestion, but looks good overall. May be worth validating in the Testing environment before deploying to Prod or Preprod.

daphne/src/roles/mod.rs
```diff
@@ -344,7 +345,22 @@ mod test {
             let taskprov_leader_token = BearerToken::from("taskprov_leader_token");
             let taskprov_collector_token = BearerToken::from("taskprov_collector_token");
 
-            let prometheus_registry = prometheus::Registry::new();
```
mendess, line 347: rename this (and the struct field it initializes) to `helper_registry`

mendess (approved):

daphne/src/roles/mod.rs
oliy: done

PR #412, opened 2023-10-16
Fix consistency bug when storing aggregate shares

mendess:
- Shard aggregate store in order to not reach durable object value size limits
- Remove legacy deserialization code
- WIP
- make try_put_out_shares atomic
- refactor handle_agg_job_cont_req to not require the full AggregationJobContReq


daphne/src/roles/helper.rs
```diff
@@ -216,42 +217,104 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+                agg_job_cont_req.round,
+                &agg_job_id,
+                &input_transitions,
+            )?;
+
+            let put_shares_result = self
+                .try_put_agg_share_span(task_id, task_config, agg_share_span)
+                .await;
+
+            for t in agg_job_resp.transitions {
```
cjpatton, line 242: You should be able to accomplish this with `iter_mut()` and `zip()` since `aggJ_job_resp` has the same sequence of reports and in the same order. (Otherwise there is a bug.)

daphne/src/roles/helper.rs
```diff
@@ -216,42 +217,104 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+
+            let inc_restart_metric = Once::new();
+            for (_, result) in put_shares_result {
+                match result {
+                    // we got 0 replays
+                    (Ok(replays), reports) if replays.is_empty() => {
+                        // remove successfull reports in this batch as they are finished
+                        input_transitions
+                            .retain(|t| reports.iter().all(|(id, _)| *id != t.report_id));
+                        // set those reports as finished in the output_transitions
```
cjpatton, line 259: I'd like code comments to consistently use proper punctuation.
```suggestion
                        // Set those reports as finished in the output_transitions.
```

daphne/src/roles/helper.rs
```diff
@@ -216,42 +217,104 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
+                            .iter_mut()
+                            .filter(|(id, _)| replays.contains(id))
+                            .for_each(|(_, var)| {
+                                *var = Some(TransitionVar::Failed(
+                                    crate::messages::TransitionFailure::ReportReplayed,
+                                ))
+                            });
+                        inc_restart_metric.call_once(|| metrics.agg_job_cont_restarted_inc());
+                    }
+                    (Err(e), _) => return Err(e.into()),
```
cjpatton, line 280: Worth commenting that failure here might result in an a botched aggregate result (Leader and Helper may have inconsistent aggregate shares).

daphne/src/roles/helper.rs
```diff
@@ -216,42 +217,104 @@ pub trait DapHelper<S>: DapAggregator<S> {
...
-            loop {
-                if retry_count < 1 {
-                    // we need to prevent an attacker from keeping this loop running for too long,
-                    // potentialy enabling an DOS attack.
-                    return Err(DapAbort::BadRequest(
-                        "AggregationJobContinueReq contained too many replays".into(),
-                    ));
+        const RETRY_COUNT: u32 = 3;
+        let mut input_transitions = agg_job_cont_req.transitions;
+        let mut output_transitions: Vec<(ReportId, Option<TransitionVar>)> = input_transitions
```
cjpatton, line 222: Let's try to think of names for these that are more evocative of their roles. Perhaps "uncommitted transitions" and "committed transitions", in the sense of "committed to the storage change"

cjpatton (commented):
1. Looks reasonable over all. However I'm wondering if you could help me with something: In https://github.com/cloudflare/daphne/pull/411 I've tried to encapsulate this logic so that I can reuse it in draft07. (In draft07 we need to commit to the state change at an earlier point, just after AggregationJobInitReq.)
2. It would be great to exercise the retry logic in unit tests. Could you set up `MockAggregator` to haave one of the batch buckets return a replay and another not?
3. nit: It's good practice to link PRs to issues they close, e.g., by adding `Closes #408` to the top comment.

daphne/src/lib.rs
```diff
@@ -340,6 +332,40 @@ impl DapAggregateShareSpan {
...
+}
+
+impl<T> Extend<(DapBatchBucket, (T, Vec<(ReportId, Time)>))> for DapAggregateSpan<T> {
+    fn extend<I>(&mut self, iter: I)
+    where
+        I: IntoIterator<Item = (DapBatchBucket, (T, Vec<(ReportId, Time)>))>,
+    {
+        self.span.extend(iter)
+    }
+}
```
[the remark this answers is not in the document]
mendess: you can't have methods on type aliases so we'd lose out on the nice specialized methods for DapAggregateShare

daphne/src/roles/helper.rs
mendess: They have to be in the same order but it's not the same sequence. So I can't use zip but I can make this O(n) instead of O(n^2)

daphne/src/lib.rs
```diff
@@ -340,6 +332,40 @@ impl DapAggregateShareSpan {
...
+
 /// Per-task DAP parameters.
 #[derive(Clone, Deserialize, Serialize)]
 pub struct DapTaskConfig {
```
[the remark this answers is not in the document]
mendess: It's used for the return type of `try_put_agg_share_span`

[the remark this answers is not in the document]
cjpatton: Yup, I see that.

daphne/src/roles/helper.rs
cjpatton: How might the sequences differ? I would expect `agg_job_resp.transitions[i].report_id` should be equal to `output-transitions[i].report_id` for all `i`.

daphne/src/roles/helper.rs
```diff
@@ -449,6 +423,93 @@ fn resolve_agg_job_id<'id, S>(
     }
 }
```
cjpatton, line 425: What should we call this function?

cjpatton (commented):
I've tacked on a commit with a suggestion for cleaning up the retry logic. The basic idea is to add a table that tracks what we did with each report the last time we tried `try_put_agg_share_span()`. Feel free to iterate on this or replace it with something you think is better.

cjpatton (approved):
🦆  - thanks for the hard work on a really tricky change.

cjpatton (commented):
I just noticed the changes to DOs. (Sorry, I should have caught this earlier.) There might be changes here that break assumptions in `daphne`, we need to carefully document anything that's broken.

Also, `ReportsProcessed` is now obsolete, so we should either delete it or at least document somewhere that it's not used and will be deleted.

daphne/src/roles/leader.rs
```diff
@@ -436,13 +436,18 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
...
-        let replayed = self
+        let put_agg_share_result = self
             .try_put_agg_share_span(task_id, task_config, agg_share_span)
-            .await?;
+            .await;
+
+        let replayed: usize = put_agg_share_result
+            .into_iter()
+            .map(|(_, (r, _))| r.map(|replayed| replayed.len()).unwrap_or_default())
+            .sum();
```
cjpatton, line 446: If an error occurs, we need to handle it as fatal.

PR #415, opened 2023-10-26
Remove INITIALIZE and MARK_AGGREGATED from ReportsProcessed

mendess:
The initial idea was to remove ReportsProcessed but it's still needed in order
to detect replays in the initialization step


daphne/src/vdaf/mod.rs
```diff
@@ -307,6 +319,18 @@ impl<'req> EarlyReportStateInitialized<'req> {
         };
         Ok(early_report_state_initialized)
     }
+
+    /// Turn this report into a rejected report using `failure` as the reason for it's rejection.
```
cjpatton, line 323: Here again: Just want to double check that overwriting the failure reason in case it was already rejected is intended.

daphne_worker/src/roles/aggregator.rs
```diff
@@ -48,133 +46,112 @@ impl DapReportInitializer for DaphneWorker<'_> {
...
-                durable_name,
-            ));
-        }
-        let agg_store_responses: Vec<bool> = try_join_all(agg_store_requests)
+                Ok(initialized)
+            })
+            .collect::<Result<Vec<_>, DapError>>()
+            .map_err(|e| fatal_error!(err = ?e, "failed to initialize a report"))?;
+
+        let replayed_reports_check = futures::stream::iter(reports_processed_request_data)
```
cjpatton, line 129: Since replay protection has now moved to AggregateStore, I don't think there's any sense in doing replay protection here as well. We will eventually delete this code --- any reason not to do so now?

cjpatton (approved):
Looks reasonable enough given the transitional phase we're in. I've left some suggestions and pointed out places that could use more clarification.

daphne/src/vdaf/mod.rs
```diff
@@ -185,6 +186,17 @@ impl<'req> EarlyReportStateConsumed<'req> {
...
+
+    pub fn into_initialized_rejected_due_to(
+        self,
+        failure: TransitionFailure,
+    ) -> EarlyReportStateInitialized<'req> {
+        let metadata = match self {
+            Self::Ready { metadata, .. } => metadata,
+            Self::Rejected { metadata, .. } => metadata,
+        };
+        EarlyReportStateInitialized::Rejected { metadata, failure }
```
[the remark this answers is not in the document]
mendess: I think we should override, principle of least surprise, if I pass a parameter I expect it to be used

daphne_worker/src/roles/aggregator.rs
cjpatton: Good catch. I suppose this will get simpler when we move the collected check to AggregateStore as well?

PR #416, opened 2023-10-31
Remove `DapVersion::Unknown`

mendess:
The code is littered with `unrecheable` and `panic` because it assumes one has checked the dap version before but we can just avoid this bug altogether if we don't have `DapVersion::Unknown`!

Also I think our error types might need a bit of a redesign but I'm not sure yet what that could be, but I found our current error types very annoying to work with, specially in the third commit of this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -710,7 +710,6 @@ dependencies = [
  "daphne",
  "futures",
  "hex",
- "matchit 0.7.3",
  "once_cell",
  "paste",
  "prio",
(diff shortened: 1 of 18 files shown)


mendess:
CI is failing on code that doesn't exist..... I'm honestly super confused as to what's happening

cjpatton (approved):
👍 thanks for taking the time to fix this.

mendess:
@cjpatton please review this again. I had to change the leader routes to make it pass the tests.

cjpatton (approved):
This seems fine. I would just add logic to make sure Draft07 is not used on these routes.

PR #420, opened 2023-11-10
Fix warnings provided by clippy::pendantic

mendess:
Each commit fixes one kind of warning included in the clippy::pendantic lint
group.

Some of the warnings in this group are too strict or provide too many false
positives but a lot of them, in my opinion, reduce code noise and improve
reliablity of the code. So I've enabled pendantic warning levels in all top
level files and disabled the ones that I think are more noise than benefit.

Some of these lints can be automatically fixed with `cargo clippy --all --fix -- -W clippy::WARNING_NAME`

### DIFF
--- a/daphne/src/vdaf/mod.rs
+++ b/daphne/src/vdaf/mod.rs
@@ -1367,16 +1367,11 @@ impl VdafConfig {
                 break;
             };
 
-            let leader_message = match &leader.var {
-                TransitionVar::Continued(message) => message,
-
-                // TODO Log the fact that the helper sent an unexpected message.
-                _ => {
-                    return Err(DapAbort::UnrecognizedMessage {
-                        detail: "helper sent unexpected message instead of `Continued`".to_string(),
-                        task_id: Some(*task_id),
-                    })
-                }
+            let TransitionVar::Continued(leader_message) = &leader.var else {
+                return Err(DapAbort::UnrecognizedMessage {
+                    detail: "helper sent unexpected message instead of `Continued`".to_string(),
+                    task_id: Some(*task_id),
+                });
             };
 
             let var = match report_status.get(&leader.report_id) {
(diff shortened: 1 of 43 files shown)


daphne/src/vdaf/mod.rs
```diff
@@ -1685,21 +1674,20 @@ mod test {
         fn unwrap_continued(self) -> (DapAggregationJobState, M) {
             match self {
                 Self::Continued(state, message) => (state, message),
-                _ => panic!("unexpected transition"),
+                Self::Finished(_, _) => panic!("unexpected transition"),
```
cjpatton, line 1677: I generally `..` when I don't want any parameters. But I guess clippy is calling the shots now.

daphne_worker/src/lib.rs
```diff
@@ -2,6 +2,22 @@
...
+#![allow(clippy::too_many_lines)]
+#![allow(clippy::needless_pass_by_value)]
+#![allow(clippy::ignored_unit_patterns)]
+#![allow(clippy::if_not_else)]
+#![allow(clippy::default_trait_access)]
+#![allow(clippy::items_after_statements)]
+#![allow(clippy::redundant_closure_for_method_calls)]
+#![allow(clippy::inconsistent_struct_constructor)]
+#![allow(clippy::similar_names)]
+#![allow(clippy::inline_always)]
```
cjpatton, line 20: This long list of allows makes me wonder if adding the aggressive clippy to CI is the right strategy. To be clear, I don't mind the changes that it suggested, it's just that I worry if this will be a huge pain. Maybe we should just run it once in a while?

cjpatton (approved):
LGTM. I'm a little concerned that "aggressive clippy" is going to be kind of annoying. I'm wondering if we really want it to be in CI. I'll defer to your judgement.

daphne_worker/src/lib.rs
cjpatton: Similar comment elsewhere.

daphne/src/vdaf/mod.rs
mendess: actually this was a bit of blind clippy following from me, since this is an unwrap method it should really be a let-else instead of a match

PR #427, opened 2023-11-20
draft07: Update `Report`

cjpatton:
Partially addresses #350.
Stacked on #426

### DIFF
--- a/daphne_worker/src/durable/mod.rs
+++ b/daphne_worker/src/durable/mod.rs
@@ -684,7 +684,7 @@ mod test {
         reports_pending::PendingReport,
     };
     use daphne::{
-        messages::{BatchId, Report, ReportId, ReportMetadata, TaskId},
+        messages::{BatchId, HpkeCiphertext, Report, ReportId, ReportMetadata, TaskId},
         test_versions, DapBatchBucket, DapVersion,
     };
     use prio::codec::{ParameterizedDecode, ParameterizedEncode};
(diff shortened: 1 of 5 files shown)


daphne_worker/src/durable/mod.rs
```diff
@@ -732,7 +732,18 @@ mod test {
...
+                    config_id: 0,
+                    enc: Vec::default(),
+                    payload: Vec::default(),
+                },
+                HpkeCiphertext {
+                    config_id: 0,
+                    enc: Vec::default(),
+                    payload: Vec::default(),
+                },
+            ],
```
mendess, line 746: Can't you use `Default::default()` here instead of writing out the array? I never know what kinds of array implement default but I think these can

mendess (approved):

daphne_worker/src/durable/mod.rs
cjpatton: I tried this:
```rust
HpkeCiphertext{ ..Default::default }
```

but `cargo` wasn't happy.

We could also implement `Default` for `HpkeCiphertext`, but I'm way of doing this here because it's not totally clear what "safe default" behavior would be.

mendess: Ah yeah, if `HpkeCipherText` doesn't implement `Default` than this is the only way to initialize this array. And I agree that a default cipher text does not make sense.

PR #428, opened 2023-11-20
More extension handling improvements

cjpatton:
Partially addresses #350.
Stacked on #427.

### DIFF
--- a/daphne/src/messages/mod.rs
+++ b/daphne/src/messages/mod.rs
@@ -1143,13 +1143,11 @@ impl Encode for PlaintextInputShare {
 
 impl Decode for PlaintextInputShare {
     fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
-        let share = Self {
-            extensions: decode_u16_items(&(), bytes)?,
-            payload: decode_u32_bytes(bytes)?,
-        };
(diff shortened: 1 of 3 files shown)


daphne/src/messages/mod.rs
```diff
@@ -1130,34 +1114,44 @@ impl Decode for HpkeCiphertext {
...
+                encode_u16_items(bytes, &(), extensions);
+                encode_u32_bytes(bytes, &self.payload);
+            }
+            (DapVersion::Draft07, None) => {
+                unreachable!("PlaintextInputShare with missing extensions")
+            }
+            (DapVersion::Draft02, ..) => {
+                unreachable!("cannot encode PlaintextInputShare in draft02")
+            }
+        }
```
mendess, line 1134: This feels wrong:
- you're taking a version parameter but it's never possible to use any variant other than `Draft07`, which means you're passing more information but only use it to panic
- the type allows for `None` extensions but this is an unencodable state, which leads me to think it's also an invalid state for the type

mendess (changes requested):

daphne/src/messages/mod.rs
cjpatton: Ugh, yeah you're absolutely right. I removed the decoding parameter and cleaned up, please take another look.

mendess (approved):

PR #429, opened 2023-11-20
draft07: Add ping-pong message framing as in draft-irtf-cfrg-vdaf-07

cjpatton:
Partially addresses #350.
Stacked on #428.

The payload of the PrepareInit message is prefixed by a `0` byte, indicating to the consumer that this is the ping-pong "initialize" message.

The payload of each transition in the AggregationJobResp is prefixed by a `2` byte, indicating to the consumer that this is the ping-pong "finish" message.

Since we only support 1-round VDAFs, it is not necessary to implement the "continue" message.


mendess (approved):
Seems fine but why not use some constants or an enum instead of the numbers `0` and `2`?

cjpatton:
Rebased.

mendess (approved):

Issue #413, opened 2023-10-17
Propose a change to libprio-rs such that `AggregateShare`s can be created from bytes without having to manually construct them


mendess:

cjpatton:
we should be good to go here: https://github.com/divviup/libprio-rs/pull/173/files

cjpatton:
Actually 173 was reverted at some point. This has added it back: https://github.com/divviup/libprio-rs/pull/857

PR #433, opened 2023-11-22
Remove ReportsProcessed DO

mendess:


daphne_worker/src/roles/aggregator.rs
```diff
@@ -123,31 +121,6 @@ impl DapReportInitializer for DaphneWorker<'_> {
...
-            })
-            .await
-            .map_err(|e| fatal_error!(err = ?e, "checking for replayed reports"))?;
-
-        for rep in &mut initialized_reports {
-            if replayed_reports_check.contains(&rep.metadata().id) {
-                rep.reject_due_to(TransitionFailure::ReportReplayed);
-            }
-        }
-
```
cjpatton, line 150: Please double check that the documentation for the `DapAggregator` trait is up-to-date. IIRC there is some language about checking for replays. Note that we also want to modify `MockAggregator` accordingly.

cjpatton (commented):
Great to see that we can basically rip this out. However I think I see at least one regression: The Leader no longer checks for replays in time to reject the reports during an aggregation job.

Also, and this is somewhat separate so feel free to punt: The `ReportsProcessed` DO instances are self-cleaning in the sense that, on creation, an alarm is set that deletes the instance some time later. Now would be a good time to implement this for `AggregateStore` as well, since we mainly just need to copy-paste the code from the `reports_processed` module to `aggregate_store`.

daphne/src/roles/leader.rs
```diff
@@ -98,6 +99,17 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
     /// Data type used to guide selection of a set of reports for aggregation.
     type ReportSelector;
 
+    /// Register these report ids as received by the leader. Returning all ids that have been
```
cjpatton, line 102: I think we're already using the term "processed" for what you mean by "registered" here?

daphne/src/roles/mod.rs
```diff
@@ -1103,49 +1103,6 @@ mod test {
 
     async_test_versions! { handle_agg_job_req_transition_continue }
```
cjpatton, line 1105: Why delete this test?

daphne/src/testing.rs
```diff
@@ -710,11 +710,10 @@ impl MockAggregator {
     /// Conducts checks on a received report to see whether:
     /// 1) the report falls into a batch that has been already collected, or
     /// 2) the report has been submitted by the client in the past.
-    fn check_report_early_fail(
```
cjpatton, line 713: The comment above needs to be updated.

daphne_worker/src/durable/reports_processed.rs
```diff
@@ -109,7 +107,7 @@ impl ReportsProcessed {
         .await?;
 
         match (req.path().as_ref(), req.method()) {
-            (DURABLE_REPORTS_PROCESSED_INITIALIZED, Method::Post) => {
+            (DURABLE_REPORTS_INITIALIZED_REGISTER, Method::Post) => {
```
cjpatton, line 110: What do we mean here by "initialized register"?

daphne_worker/src/roles/aggregator.rs
```diff
@@ -48,34 +46,14 @@ impl DapReportInitializer for DaphneWorker<'_> {
...
-                        .or_insert_with(Vec::new)
-                        .push(id);
-                }
-                agg_store_request_names.push((
-                    bucket,
-                    durable_name_agg_store(task_config.version, &task_id_hex, bucket),
-                ));
-            }
-
             // Send AggregateStore requests.
```
cjpatton, line 52: Do we still need to do this batch collected check here?

cjpatton (commented):
Looking good! One more question about the code itself, other comments are just related to documentation.

daphne/src/roles/aggregator.rs
```diff
@@ -26,8 +26,8 @@ use crate::{
 #[async_trait(?Send)]
 pub trait DapReportInitializer {
     /// Initialize a sequence of reports that are in the "consumed" state by performing the early
-    /// validation steps (check if the report was replayed, belongs to a batch that has been
-    /// collected) and initializing VDAF preparation.
+    /// validation steps (belongs to a batch that has been collected) and initializing VDAF
```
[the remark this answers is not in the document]
mendess: Yes, this check still happens

daphne/src/roles/helper.rs
```diff
@@ -517,6 +518,16 @@ async fn finish_agg_job_and_aggregate<S>(
                     }));
                     inc_restart_metric.call_once(|| metrics.agg_job_put_span_retry_inc());
                 }
+                // This bucket belongs to a collected aggregate share.
+                (Err(MergeAggShareError::AlreadyCollected), reports) => {
+                    report_status.extend(reports.into_iter().map(|(report_id, _)| {
```
cjpatton, line 523: Suggestion: We call `report_status.extend()` in every(?) arm of this match statement. What about having the match statement return a thing that we pass to `report_status.extend()`?

daphne/src/roles/leader.rs
```diff
@@ -439,8 +443,13 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
             .try_put_agg_share_span(task_id, task_config, agg_span)
```
cjpatton, line 443: Update the comment above ("if we do detect an error (at least one report was replayed, the span overlaps with a collected batch, etc.")

daphne/src/roles/leader.rs
```diff
@@ -439,8 +443,13 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
...
             .await
             .into_iter()
-            .map(|(_bucket, (result, _report_metadata))| {
-                result.map(|replayed_reports| replayed_reports.len())
+            .map(|(_bucket, (result, _report_metadata))| match result {
+                Ok(()) => Ok(0),
+                Err(MergeAggShareError::AlreadyCollected) => {
+                    panic!("aggregated to a collected agg share")
```
cjpatton, line 449: Why panic? Wouldn't we want to handle this similarly to detected replays, i.e., just output a log line with the warning?

daphne/src/roles/leader.rs
```diff
@@ -439,8 +443,13 @@ pub trait DapLeader<S>: DapAuthorizedSender<S> + DapAggregator<S> {
...
+                }
+                Err(MergeAggShareError::ReplaysDetected(replays)) => Ok(replays.len()),
+                Err(MergeAggShareError::Other(e)) => Err(e),
             })
             .sum::<Result<usize, _>>()?;
```
cjpatton, line 455: on line 457 below: Perhaps this should be `tracing::error!`?

daphne/src/roles/mod.rs
cjpatton: Bump: As discussed on Tuesday, we ought to rewrite this test rather than delete it. All that's changed is that we don't detect replays until just before completing the aggregation job. In draft02, this is on the continue request; in the latest draft, this is on the initial request.

daphne/src/testing.rs
cjpatton: Bump

daphne_worker/src/config.rs
```diff
@@ -121,10 +121,6 @@ pub(crate) struct DaphneWorkerConfig {
     /// configured by the Leader.
     pub(crate) helper_state_store_garbage_collect_after_secs: Option<Duration>,
```
cjpatton, line 123: Just a note that we'll want to reimplement something like this for the AggregateStore DO at some point (https://github.com/cloudflare/daphne/issues/413).

cjpatton (changes requested):
Looks great overall, but I think this needs a few minor changes. I've also bumped some comments that we missed in the previous round.

Also, I think we should delete this outdated comment we noticed on Tuesday (this caused some confusion about if/when the Leader needs to check for replays): https://github.com/cloudflare/daphne/blob/main/daphne_worker/src/roles/leader.rs#L105-L109

daphne/src/roles/helper.rs
mendess: it will be uglier, each of the things we are passing to extend is a different type so they won't unify. We could collect to a vec and return that or use dynamic dispatch, but both of those have performance costs and I don't think they improve readability that much

cjpatton (approved):
🦆  great work

PR #434, opened 2023-11-22
Make abort->error conversion one way

mendess:
Having DapError convertible into DapAbort means that an Error can be
converted into an abort and an abort converted into an error an infinite
number of times, meaning the type can grow in an unbounded way at
runtime. This is very confusing to reason about when reading the
signature of a method that returns a `DapAbort`. Now a method that
returns a `DapAbort` can only return due to a documented protocol error
and not another system related error (such as mis configuration or IO).

This PR removes the `impl From<DapError> for DapAbort` implementation and fixes
the code accordingly.

### DIFF
--- a/daphne/src/error/aborts.rs
+++ b/daphne/src/error/aborts.rs
@@ -35,10 +35,6 @@ pub enum DapAbort {
     #[error("batchOverlap")]
     BatchOverlap { detail: String, task_id: TaskId },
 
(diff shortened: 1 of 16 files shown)


cjpatton (approved):
Nice idea 👍. It doesn't look to me like this changes the HTTP responses, and it most definitely simplifies the code.

PR #435, opened 2023-11-22
Implement `XofHmacSha256Aes128`

cjpatton:
Add a new module, `daphne::vdaf::xof`, that implements a custom XOF based on HMAC-SHA256 and AES128 in CTR mode.


Cargo.toml
```diff
@@ -25,16 +25,17 @@ cap = "0.1.2"
...
-getrandom = { version = "0.2.11", features = ["js"] } # Required for prio
+getrandom = { version = "0.2.11", features = ["js", "std"] } # Required for prio
 hex = { version = "0.4.3", features = ["serde"] }
 hpke-rs = "=0.1.0"
 hpke-rs-crypto = "=0.1.1"
 hpke-rs-rust-crypto = "=0.1.1"
 matchit = "0.7.3"
 paste = "1.0.14"
-prio = "0.15.3"
+prio = { git = "https://github.com/divviup/libprio-rs", rev = "46e8b8bd3c9aed1a1b674e95c49fefeae737e06b" }
```
mendess, line 35: How long would we have to wait for a release of the `prio` lib so we don't have to have a git dependency?

daphne/src/vdaf/test_vec/XofHmacSha256Aes128.json
```diff
@@ -0,0 +1,8 @@
+{
```
mendess, line 1: maybe name this dir `xof_test_data`? `test_vec` doesn't mean anything to me

cjpatton: `test_vec` is consistent with the base draft and libprio: https://github.com/divviup/libprio-rs/tree/main/src/vdaf/test_vec

Cargo.toml
cjpatton: FYI, here's the change we need for this PR: https://github.com/divviup/libprio-rs/pull/846

cjpatton: https://github.com/divviup/libprio-rs/issues/839

mendess: Then maybe just create an issue to change this back to a normal dependency later

mendess (approved):

Cargo.toml
cjpatton: Done.

PR #436, opened 2023-11-23
Enable taskprov in draft07

cjpatton:
Stacked on #435.
Partially addresses #350.

Enable taskprov in draft07

Enable support for taskprov in the latest draft. The main technical
hurdle is to ensure the Leader can advertise the taskprov config to the
Helper in an HTTP Header. The simplest solution is to add the
information to `DapTaskConfig` that is needed in order to reconstruct
the serialized taskprov config, in particular the "task_info" field.

To do this, we add a new `method` field to the `DapTaskConfig` struct
that conveys the method by which the task was configured. The method has
type `enum DapTaskConfigMethod`, of which there are two variants. The
first is `Taskprov` and includes the "task_info". The second is
`Unknown` and is a catch-all for existing, manually configured tasks.

Accordingly, we deprecate the `taskprov` field of the `DapTaskConfig`
struct. We've kept the field in the struct for backwards compatibility:
there are some situations for which the indication is sufficient; in
situations where the "task_info" field is needed, we must abort.

This change also includes a number of related, but minor changes:

* Have `resolve_advertised_task_config()` return a `DapAbort` rather
  than a `DapError` (an error in this flow always leads to an abort and
  is never an internal error).

* Opt-out of the task unless (1) the "none" variant for "DpConfig" is
  indicated and (2) the indicated max batch query count is 1. Other
  values for these parameters are not currently supported by Daphne.

* Introduce a type `DapTaskParameters` that encapsulates the parameters
  used to configure a task and implement a method for generating a task
  config using the taskprov method. This change is intended to simplify
  the test code, but should be independently useful.

### DIFF
--- a/daphne/src/lib.rs
+++ b/daphne/src/lib.rs
@@ -79,8 +79,9 @@ use crate::{
 use constants::DapMediaType;
 pub use error::DapError;
 use hpke::{HpkeConfig, HpkeKemId};
+use messages::encode_base64url;
 use prio::{
-    codec::{Decode, Encode, ParameterizedDecode},
+    codec::{Decode, Encode, ParameterizedDecode, ParameterizedEncode},
     vdaf::Aggregatable as AggregatableTrait,
 };
 use rand::prelude::*;
@@ -95,13 +96,14 @@ use url::Url;
 use vdaf::{EarlyReportState, EarlyReportStateConsumed};
 
 /// DAP version used for a task.
-#[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+#[derive(Clone, Copy, Debug, Default, Deserialize, Eq, Hash, PartialEq, Serialize)]
 #[cfg_attr(any(test, feature = "test-utils"), derive(deepsize::DeepSizeOf))]
 pub enum DapVersion {
     #[serde(rename = "v02")]
     Draft02,
 
     #[serde(rename = "v07")]
+    #[default]
     Draft07,
 }
 
@@ -402,10 +404,22 @@ impl Extend<(DapBatchBucket, (ReportId, Time))> for DapAggregateSpan<()> {
     }
 }
 
-/// Per-task DAP parameters.
-#[derive(Clone, Deserialize, Serialize)]
+/// Method for configuring tasks.
+#[derive(Clone, Default, Deserialize, Serialize)]
 #[cfg_attr(test, derive(PartialEq, Debug))]
-pub struct DapTaskConfig {
+pub enum DapTaskConfigMethod {
+    /// draft-wang-ppm-dap-taskprov-06
+    Taskprov {
+        /// `TaskConfig.task_info`. If not set, then the task info is unknown.
+        info: Option<Vec<u8>>,
+    },
+
+    #[default]
+    Unknown,
+}
+
+/// Base parameters used to configure a DAP task.
+pub struct DapTaskParameters {
     /// The protocol version (i.e., which draft).
     pub version: DapVersion,
 
@@ -419,8 +433,8 @@ pub struct DapTaskConfig {
     /// constrain the batch interval of time=interval queries.
     pub time_precision: Duration,
 
-    /// The time at which the task expires.
-    pub expiration: Time,
+    /// The amount of time before the task should expire.
+    pub lifetime: Duration,
 
     /// The smallest batch permitted for this task.
     pub min_batch_size: u64,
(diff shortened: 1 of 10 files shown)


daphne/src/lib.rs
```diff
@@ -429,16 +445,113 @@ pub struct DapTaskConfig {
 
     /// The VDAF configuration for this task.
     pub vdaf: VdafConfig,
+}
+
+impl DapTaskParameters {
+    /// Construct a new task config using the taskprov extension. Return the task ID, the taskprov
+    /// advertisement (if applicable), and the payload of the report extension.
+    pub fn to_config_with_taskprov(
```
mendess, line 453: this method seems to only be used in tests, that mean it can actually be moved into a test block in this file so it's out of the way of the main code. Other test modules in the library can still access it
```rs
#[cfg(test)]
mod test {
  impl DapTaskParameters {
    // ...
  }
}
```

mendess (changes requested):

daphne/src/lib.rs
cjpatton: It's also needed by `daphne_worker_test`. I added `#[cfg(any(test, feature = "test-utils"))]` and left it where it is. We could move it to `testing` if that makes sense?

daphne/src/lib.rs
```diff
@@ -429,16 +445,113 @@ pub struct DapTaskConfig {
...
+            lifetime: 86400 * 14, // two weeks
+            min_batch_size: 10,
+            query: DapQueryConfig::TimeInterval,
+            vdaf: VdafConfig::Prio2 { dimension: 10 },
+        }
+    }
+}
+
+/// Per-task DAP parameters.
+#[derive(Clone, Deserialize, Serialize)]
```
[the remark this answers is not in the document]
cjpatton: Yes, we write it to KV.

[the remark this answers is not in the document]
cjpatton: `DapTaskParameters` is not ... I removed the derived traits from there.

daphne/src/lib.rs
```diff
@@ -429,16 +444,114 @@ pub struct DapTaskConfig {
...
-    /// If true, then the taskprov extension was used to configure this task.
+    /// Deprecated. Indicates that the task was configured via draft-wang-ppm-taskprov.
+    #[serde(default, rename = "taskprov")]
+    pub deprecated_taskprov: bool,
+
+    /// Method by which the task was configured.
     #[serde(default)]
-    pub taskprov: bool,
+    pub method: DapTaskConfigMethod,
 }
```
mendess, line 555: It would be nice if these two fields were somehow coalesced. There seems to be quite a fair number of invalid states that are possible between these fields and `version` and they are all `pub`. Could we encapsulate this a bit more so this type is more obvious in how it's supposed to be used?

mendess (changes requested):

daphne/src/lib.rs
cjpatton: Which invalid states?

Note that one requirement I'm trying to fullfill with this change is maintaining backwards compatibility with `DapTaskConfig`s that are already in KV. This means the old JSON object has a field called `taskprov` and the new one has a field called `method`. This change has code that uses the old field if available.

daphne/src/lib.rs
```diff
@@ -573,6 +686,30 @@ impl DapTaskConfig {
...
         Ok(report_count >= self.min_batch_size)
     }
+
+    /// Leader: Resolve taskprov advertisement to send in a request to the Helper.
+    pub(crate) fn resolve_taskprove_advertisement(&self) -> Result<Option<String>, DapError> {
+        if let DapTaskConfigMethod::Taskprov { info: _ } = &self.method {
+            let encoded_taskprov_config = messages::taskprov::TaskConfig::try_from(self)?
+                .get_encoded_with_param(&self.version);
+            Ok(Some(encode_base64url(encoded_taskprov_config)))
+        } else if self.deprecated_taskprov && self.version != DapVersion::Draft02 {
```
mendess, line 696: This if signals to me there `deprecated_taskprov` can only be `true` if the version is `Draft2`

daphne/src/lib.rs
mendess: See [this comment](https://github.com/cloudflare/daphne/pull/436#discussion_r1406308160)

daphne/src/lib.rs
cjpatton: That's not true (see code comment below). The intent here is to abort loudly if we're in a situation where we can't proceed.

A simpler alternative would be to fail here if `self.version != DapVersion::Draft02`, regardless of the value of `self.deprecated_taskprov`.

mendess (approved):

PR #438, opened 2023-11-27
draft07, taskprov: Align `QueryConfig` with the latest draft

cjpatton:
Partially addresses #350.
Stacked on #436.

PR #439, opened 2023-11-28
draft07, taskprov: Align `VdafConfig` and clean up

cjpatton:
Partially addresses #350.
Stacked on #438.

Align serialization of `VdafConfig` with the latest draft. Along the way, remove the `VdafType` and simplify. This struct is used only for aligning the VDAF verify key derivation with `ring`'s HKDF API. It's sufficient (and simpler) to implement the necessary trait on our own `VdafVerifyKey` type.

### DIFF
--- a/daphne/src/vdaf/mod.rs
+++ b/daphne/src/vdaf/mod.rs
@@ -49,6 +49,7 @@ use prio::{
 };
 use rand::prelude::*;
 use replace_with::replace_with_or_abort;
+use ring::hkdf::KeyType;
 use serde::{Deserialize, Serialize, Serializer};
 use std::{
     borrow::Cow,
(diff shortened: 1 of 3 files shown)


daphne/src/vdaf/mod.rs
```diff
@@ -127,6 +137,15 @@ impl AsRef<[u8]> for VdafVerifyKey {
     }
 }
 
+impl AsMut<[u8]> for VdafVerifyKey {
+    fn as_mut(&mut self) -> &mut [u8] {
+        match self {
+            Self::Prio3(ref mut bytes) => &mut bytes[..],
+            Self::Prio2(ref mut bytes) => &mut bytes[..],
```
mendess, line 144: are these `ref mut` necessary? I think you can omit them

mendess (approved):

cjpatton:
Rebased.

daphne/src/vdaf/mod.rs
cjpatton: YOu're right, removed.

PR #441, opened 2023-11-28
taskprov: Enforce param length prefixes

cjpatton:
Partially addresses #350.
Stacked on #439.

Structs `VdafConfig`, `DpConfig`, and `QueryConfig` all include a length prefix for the type-specific parameters so that the struct can be decoded even if the type is not recognized. If the parameters don't fill the space indicated by the length prefix, then decoding should fail. This change ensures the length is enforced for recognized types. It unifies the decoding logic across draft02 and draft07 (the former does not have the length prefix).

### DIFF
--- a/daphne/src/messages/mod.rs
+++ b/daphne/src/messages/mod.rs
@@ -1214,6 +1214,61 @@ pub fn decode_base64url_vec<T: AsRef<[u8]>>(input: T) -> Option<Vec<u8>> {
     URL_SAFE_NO_PAD.decode(input).ok()
 }
 
+fn encode_u16_item_for_version<E: ParameterizedEncode<DapVersion>>(
+    bytes: &mut Vec<u8>,
+    version: DapVersion,
+    item: &E,
+) {
+    match version {
+        DapVersion::Draft07 => {
+            // Cribbed from `decode_u16_items()` from libprio.
+            //
+            // Reserve space for the length prefix.
+            let len_offset = bytes.len();
+            0_u16.encode(bytes);
+
+            item.encode_with_param(&version, bytes);
+            let len_bytes = std::mem::size_of::<u16>();
+            let len = bytes.len() - len_offset - len_bytes;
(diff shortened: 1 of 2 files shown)


daphne/src/messages/mod.rs
```diff
@@ -1214,6 +1214,61 @@ pub fn decode_base64url_vec<T: AsRef<[u8]>>(input: T) -> Option<Vec<u8>> {
...
+            //
+            // Reserve space for the length prefix.
+            let len_offset = bytes.len();
+            0_u16.encode(bytes);
+
+            item.encode_with_param(&version, bytes);
+            let len = bytes.len() - len_offset - 2;
+            for (offset, byte) in u16::to_be_bytes(len.try_into().unwrap()).iter().enumerate() {
+                bytes[len_offset + offset] = *byte;
+            }
```
mendess, line 1234: This can probably be written using [copy_from_slice](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice)

daphne/src/messages/mod.rs
```diff
@@ -1214,6 +1214,61 @@ pub fn decode_base64url_vec<T: AsRef<[u8]>>(input: T) -> Option<Vec<u8>> {
...
+    match version {
+        DapVersion::Draft07 => {
+            // Cribbed from `decode_u16_items()` from libprio.
+            //
+            // Reserve space for the length prefix.
+            let len_offset = bytes.len();
+            0_u16.encode(bytes);
+
+            item.encode_with_param(&version, bytes);
+            let len = bytes.len() - len_offset - 2;
```
mendess, line 1231:
```suggestion
            let len = bytes.len() - len_offset - std::mem::size_of<u16>();
```

mendess (changes requested):

cjpatton:
Addressed your comments and force-pushed the commit.

mendess (approved):

PR #443, opened 2023-11-29
Clean up and move to draft09

### LINKED ISSUE #350: DAP-07

### LINKED ISSUE #386: Disable all VDAFs except Prio2 in draft-ietf-ppm-dap-02
If codepoints change across VDAF versions, then taskprov will end up with different codepoints in different versions. Right now we only intend to support draft-02 and the latest version. Since draft-02 is only compatible with draft-ietf-ppm-dap-02 and we don't need any new VDAFs for this version, let's put in a feature freeze.

cjpatton:
Closes #350.
Closes #386.
Stacked on #442.

* Update draft references. The next draft set we're targeting is {draft-ietf-ppm-dap-09, draft-irtf-cfrg-vdaf-08, draft-wang-ppm-dap-taskprov-06}.
* Align version tags with the latest draft.
* taskprov: Use version to resolve VDAF config


daphne/src/taskprov.rs
```diff
@@ -223,15 +223,19 @@ impl DapQueryConfig {
...
                 dimension: dimension.try_into().map_err(|_| DapAbort::InvalidTask {
                     detail: "dimension is larger than the system's word size".to_string(),
                     task_id: *task_id,
                 })?,
             }),
-            VdafTypeVar::NotImplemented { typ, .. } => Err(DapAbort::InvalidTask {
+            (.., VdafTypeVar::NotImplemented { typ, .. }) => Err(DapAbort::InvalidTask {
                 detail: format!("unimplemented VDAF type ({typ})"),
                 task_id: *task_id,
             }),
```
mendess, line 241: You're matching on the version but you never use it's value :thinking:

daphne/src/lib.rs
```diff
@@ -102,17 +106,17 @@ pub enum DapVersion {
     #[serde(rename = "v02")]
     Draft02,
 
-    #[serde(rename = "v07")]
+    #[serde(rename = "v09")]
```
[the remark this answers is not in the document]
cjpatton: This is used to construct the aggregator URL by daphne_worker, e.g., `/v09/hpke_config`

daphne/src/taskprov.rs
cjpatton: This will get used when we add new VDAFs for taskprov. We want to be able to opt-out for any VDAF other than Prio2 in draft02.

If you like I can remove this change from this PR. (I'll add it back in a future PR, probably later this week.)

mendess: if it's gonna be used soon then let it be, don't want to give you any more work

PR #452, opened 2023-12-04
Upgrade dependencies (including breaking ones) 

cjpatton:
~`serde-wasm-bindgen` is being upgraded, so we'll need to make sure we validate this change in Testing before merging downstream. This is marked as draft for now; I'll mark it as reviewable once I've tested it.~

EDIT: `serde-wasm-bindgen` is not being upgraded.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,23 +19,12 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aead"
-version = "0.4.3"
+version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0b613b8e1e3cf911a086f53f03bf286f52fd7a7258e4fa606f0ef220d39d8877"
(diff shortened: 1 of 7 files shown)


daphne_worker/src/durable/aggregate_store.rs
```diff
@@ -374,7 +374,8 @@ impl AggregateStore {
                 js_sys::Reflect::set(
                     &chunks_map,
                     &JsValue::from_str(METADATA_KEY),
-                    &serde_wasm_bindgen::to_value(&meta)?,
+                    &serde_wasm_bindgen::to_value(&meta)
+                        .map_err(|e| worker::Error::RustError(e.to_string()))?,
```
mendess, line 378: Is this because `worker` depends on an earlier version of serde-wasm-bindgen? If that's the case it would be better to make a PR to them to bump `serde-wasm-bindgen` first or not bumping the version of `serde-wasm-bindgen` ourselves yet.

I checked what changed and it doesn't seem relevant to us

cjpatton: Done.

cjpatton:
I've validated this in Testing.

mendess (approved):

PR #455, opened 2023-12-06
Create generic Base64Encode trait

mendess:
This will be necessary to simplify some code in the native implementation


cjpatton (approved):
Great idea 👍

PR #458, opened 2023-12-06
Add skeleton of the daphne server

mendess:


daphne_server/examples/configuration.toml
cjpatton, line 1: Nice, this will be easier to work with than a JSON blob.

daphne_server/examples/service.rs
cjpatton, line 1: Add copyright info

daphne_server/examples/service.rs
```diff
@@ -0,0 +1,82 @@
...
+            )?
+            .add_source(config::Environment::with_prefix("DAP"))
+            .build()?
+            .try_deserialize()
+    }
+}
+
+#[derive(clap::Parser)]
+struct Args {
+    #[arg(short, long)]
```
cjpatton, line 47: CLI documentation should indicate that a toml file is expected.

daphne_service_utils/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// Maximum number of aggregation jobs to process at once.
+    pub max_agg_jobs: u64,
+
+    /// Maximum number of reports to drain for each aggregation job.
+    pub max_reports: u64,
+}
+
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum Role {
```
cjpatton, line 21: 1. For consistency with `DapTaskConfig` et al
2. Daphne may, in the future, implement other protocols besides DAP.
```suggestion
pub enum DapRole {
```

daphne_server/examples/service.rs
```diff
@@ -0,0 +1,82 @@
+use std::path::PathBuf;
+
```
cjpatton, line 2: It's not very clear what this is an example of. More documentation is needed.

Minimally, add a docucomment describing what this does (or will do, once it's complete).

daphne_server/examples/service.rs
```diff
@@ -0,0 +1,82 @@
...
+    #[arg(short, long)]
+    configuration: Option<PathBuf>,
+    #[arg(short, long)]
+    role: Option<String>,
+    #[arg(short, long)]
+    port: Option<u16>,
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error + Sync + Send>> {
```
cjpatton, line 56: Presumably this is meant to be copy-pastable example of a Daphne service deployment. To make this more useful, I would add a ton of comments to this function explaining what each step does. To understand the system, it should be sufficient to just go through this file.

daphne_server/src/router/mod.rs
```diff
@@ -0,0 +1,455 @@
...
+        let (task_id, resource) = match version {
+            DapVersion::Draft02 => {
+                // Parse the task ID from the front of the request payload and use it to look up the
+                // expected bearer token.
+                //
+                // TODO(cjpatton) Add regression tests that ensure each protocol message is prefixed by the
+                // task ID.
+                //
+                // TODO spec: Consider moving the task ID out of the payload. Right now we're parsing it
+                // twice so that we have a reference to the task ID before parsing the entire message.
```
cjpatton, line 236: No longer relevant
```suggestion
```

daphne_server/src/worker_connection.rs
cjpatton, line 1: This definitely needs a lot of documentation. Somewhere in this crate should be an architecture diagram showing how all the pieces fit together.

By the way, it would be useful to be able to run `daphne_server` without a Workers backend and instead keep everything in memory. That way we can use it to drive tests (and perhaps replace `MockAggregator` eventually?)

daphne_server/src/worker_connection.rs
```diff
@@ -0,0 +1,108 @@
...
+    MetaAggregationJobId,
+};
+use serde::{de::DeserializeOwned, Serialize};
+use tokio::io;
+use url::Url;
+
+pub(crate) enum Error {
+    Serde(serde_json::Error),
+}
+
```
cjpatton, line 21: Add documenetation (here and the methods in the implementation)

daphne_service_utils/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "daphne_service_utils"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
```
cjpatton, line 7: Remove boilerplate

```suggestion
```

daphne_service_utils/src/lib.rs
cjpatton, line 1: Add copyright

daphne_service_utils/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+    Leader,
+    Helper,
+}
+
+impl Role {
+    pub fn is_leader(self) -> bool {
+        self == Self::Leader
+    }
+
+    #[allow(dead_code)]
```
cjpatton, line 31: If this is pub then we don't need this allow here, correct?

daphne_service_utils/src/test_route_types.rs
cjpatton, line 1: Add copyright

cjpatton (commented):
Looks great overall! My main request is to please add more documentation. (See inline comments.)

One high level idea for `daphne_server_utils` crate: From the commit message it sounds like the intent is to factor out code shared by `daphne_server` and `daphne_worker`. Since this situation is temporary (`daphne_worker` will be deprecated) we could just have `daphne_server_utils` be a submodule of `daphne_server` and have `daphne_worker` depend on it. What do you think?

check-all.sh
```diff
@@ -0,0 +1,35 @@
...
+        r cargo clippy -p daphne --features send-traits --tests
+        r cargo clippy --bin dapf
+        r cargo clippy -p daphne_worker --tests
+        r cargo clippy -p daphne-worker-test --tests
+        r cargo clippy -p daphne-worker-test --tests --features test_e2e
+        r cargo clippy -p daphne_server --tests
+        r cargo clippy --example service
+        ;;
+
+    test)
```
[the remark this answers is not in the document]
mendess: I looked around and asked around and didn't find a way to do this. Doing this goes against the compilation architecture of cargo/rustc sadly

check-all.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+r() {
+    echo "🔄 '$*'"
```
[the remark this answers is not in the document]
mendess: I use emojis more for their color than for their emojiness haha. Painting text is very hard to make portable. And the color codes show up in raw logs in github actions. It's a tradeoff

check-all.sh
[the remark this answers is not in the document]
cjpatton: Ah no worries. at least there is a script I can run.

mendess:
> Looks great overall! My main request is to please add more documentation. (See inline comments.)
> 
> One high level idea for `daphne_server_utils` crate: From the commit message it sounds like the intent is to factor out code shared by `daphne_server` and `daphne_worker`. Since this situation is temporary (`daphne_worker` will be deprecated) we could just have `daphne_server_utils` be a submodule of `daphne_server` and have `daphne_worker` depend on it. What do you think?

I think it's a good idea but cargo won't like it if you make `d_worker` depend on `d_server` because it will activate the `send-traits` feature for `d_worker` implicitly :/

daphne_server/src/roles/aggregator.rs
```diff
@@ -0,0 +1,222 @@
...
+        &self.metrics.daphne
+    }
+
+    /// Access the audit log.
+    fn audit_log(&self) -> &dyn AuditLog {
+        &NoopAuditLog
+    }
+
+    fn host(&self) -> &str {
+        "unspecified-host"
```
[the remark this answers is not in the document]
mendess: you're right, idk what I was thinking here

daphne_server/src/worker_connection.rs
mendess: On the second point. I did think about that, that's why you'll see the `router` module is agnostic of the `App` type!

daphne_server/src/worker_connection.rs
mendess: I'm expecting this module to change a lot still so I think I'll postpone documenting it for now

daphne_server/examples/service.rs
mendess: This is actually the test binary. I intend to write a readme for this later on

daphne_server/examples/service.rs
mendess: that's actually not true :smirk: you can use any kind of file as long as it's called `configuration` :eyes:

daphne_server/src/router/mod.rs
mendess: I had a feeling it wasn't relevant, but wasn't 100% sure.

daphne_server/src/router/mod.rs
```diff
@@ -0,0 +1,455 @@
...
+            .map_err(|e| (StatusCode::BAD_REQUEST, e.to_string()))?;
+
+        let sender_auth = DaphneAuth {
+            bearer_token: parts
+                .headers
+                .get("DAP-Auth-Token")
+                .and_then(|v| v.to_str().ok())
+                .map(BearerToken::from),
+            // TODO(mendess) figure out tls
+            cf_tls_client_auth: None,
```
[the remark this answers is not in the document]
mendess: How does this trait solve the problem? We just have to parse the certificate and store it in the `cf_tls_client_auth` field.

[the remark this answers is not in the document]
cjpatton: What I'm trying to avoid here is having to re-implement the `verified` string passed from FL. (i.e., "SUCCESS" if a cert was presented and was valid, "FAILED(...)" if there was some sort of error, something else if there was no certificate at all.) Frankly this is some Cloudflare-specific awkwardness that we can and should avoid if we're terminating TLS ourselves.

cjpatton (approved):

oliy (approved):

PR #460, opened 2023-12-19
Finish implementation native service

mendess:


oliy:
For the DO bindings, seems like you would want to make a trait to represents the DO operations.  Then the main implementation for them would be to make HTTP requests to the DO bindings you defined.  Feels like the primary interface to the actual endpoints themselves but a Rust interface/trait to the operations desired.  The fact that they are HTTP endpoints is just an implementation detail.

oliy:
This PR seems to explicitly tie the server implementation with the http implementation.  The original interfaces/traits seem like they would be a cleaner separation between the systems, to identify the DO operations.  For unit testing, this means that the server implementation could be tested with a pure mock interface.  Why was this removed?

daphne_service_utils/src/config.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Leader: Method for authorizing Collector requests.
+    pub collector_auth: Option<DaphneWorkerAuthMethod>,
+}
+
+pub type HpkeRecieverConfigList = Vec<HpkeReceiverConfig>;
+
+/// Daphne-Worker configuration, including long-lived parameters used across DAP tasks.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct DaphneServiceConfig {
+    pub env: String,
```
cjpatton, line 32: Missing documentation (not clear from the "some-machine-identifer" example what the intended semantics is).

daphne_worker/src/config.rs
```diff
@@ -77,57 +64,130 @@ struct MetricsPushConfig {
...
+}
+
 /// Daphne-Worker configuration, including long-lived parameters used across DAP tasks.
 pub(crate) struct DaphneWorkerConfig {
-    pub(crate) env: String,
-
-    /// Indicates if DaphneWorker is used as the Leader.
-    pub(crate) is_leader: bool,
+    service_config: DaphneServiceConfig,
+    pub do_config: DaphneWorkerDoConfig,
```
cjpatton, line 86: Elsewhere we use "durable" instead of "do" (resp. "Durable" instead of "Do"). This is precisely for the ambiguity we have here: "do" reads like the verb "do", not the acronym "D.O."

If you want to go with "do" that's fine, but let's be consistent.

daphne/src/lib.rs
```diff
@@ -1118,6 +1118,10 @@ pub struct DapRequest<S> {
     /// requests may apply to all tasks, e.g., the request for the HPKE configuration.
     pub task_id: Option<TaskId>,
 
+    /// ID of the collection job. This field is optional, since some
+    /// requests may apply to all tasks.
```
cjpatton, line 1122: What do you mean by "some requests may apply to all tasks"?

In what situations would the collection job ID not be part of the DAP request?

daphne/src/lib.rs
```diff
@@ -1173,7 +1178,9 @@ impl<S> DapRequest<S> {
     /// the version in use is not draft02. If the caller is not the Collector, or draft02 is in
     /// use, we exepct the collection job ID to be missing.
     pub fn collection_job_id(&self) -> Result<&CollectionJobId, DapAbort> {
-        if let DapResource::CollectionJob(ref collection_job_id) = self.resource {
+        if let DapResource::CollectionJob(collection_job_id) = &self.resource {
+            Ok(collection_job_id)
+        } else if let Some(collection_job_id) = &self.collection_job_id {
```
cjpatton, line 1183: Why make this change? There's no explanation in the commit message.

Why leave the `CollectionJob` variant of `DapResource` intact?

Note: `DapResource` is meant to encapsulate an idea from the DAP spec, namely that a "collection job" is a "resource". If there's a practical reason to delete this, I think it's fine, but I don't understand why we wouldn't also change `DapResource` in the same way.

cjpatton (commented):
(I'm not all the way through but I'm posting my current set of comments because I'm afraid of GitHub losing them.)

daphne_worker/src/storage_proxy.rs
cjpatton, line 1: Missing copyright (I'm tired of typing this, please try to remember).

daphne_worker_test/wrangler.storage_proxy.toml
cjpatton, line 1: Missing copyright

daphne_worker_test/wrangler.storage_proxy.toml
```diff
@@ -0,0 +1,65 @@
+name = "daphne_worker_test"
```
cjpatton, line 1: copy-pasta

daphne_worker_test/wrangler.storage_proxy.toml
```diff
@@ -0,0 +1,65 @@
...
+main = "build/worker/shim.mjs"
+compatibility_date = "2022-10-14"
```
cjpatton, line 3: copy-pasta

daphne_worker/src/storage_proxy.rs
```diff
@@ -0,0 +1,136 @@
+use daphne_service_utils::do_requests::{DoRequest, IdFrom, DO_PATH_BASE, KV_PATH_BASE};
+use url::Url;
+use worker::{js_sys::Uint8Array, Env, Request, RequestInit, Response};
+
```
cjpatton, line 4: Add a docucomment explaining what a "storage proxy" is. Also, please make sure this role is documented in the functional spec (the name should also match).

Related: Does this code share any infrastructure with `daphen_worker` other than `DoRequest`?

daphne_worker/src/storage_proxy.rs
```diff
@@ -0,0 +1,136 @@
...
+const KV_BINDING_DAP_CONFIG: &str = "DAP_CONFIG";
+
+pub async fn handle_request(
```
cjpatton, line 7: Is this the service that the edge service talk to?

daphne_worker/src/storage_proxy.rs
```diff
@@ -0,0 +1,136 @@
...
+    req: Request,
+    env: Env,
+    _ctx: worker::Context,
+) -> worker::Result<Response> {
+    let path = req.path();
+    if let Some(uri) = path.strip_prefix(KV_PATH_BASE) {
```
cjpatton, line 13: What are we using KV for? If DAP tasks (i.e., control plane) We discussed replacing it with D1. Where are we on that discussion?

cjpatton (changes requested):
I made it half-way through the sixth commit. So far these changes mostly look great/useful, but I think this is too big of a change to review in one PR. Some of the commits have design questions that need to discuss.

I suggest we proceed as follows:
1. Rebase and merge #458 
2. Address current feedback (this will make review faster)
3. Split the commits as follows: Make the first 5 commits into 5 PRs (these are all separable and easy to review on their own); make the "storage proxy" worker another PR; and make the end-to-end daphne server bits (basically the remaining commits) their own PR. 
4. Close this PR.

You can continue working on your commit stack, but you'll have to rebase as downstack PRs are approved and merged.

daphne_worker/src/storage_proxy.rs
mendess: you don't have to comment for every file! I'll try to remember but if you just mention it once I'll look at every new file.

daphne_worker/src/storage_proxy.rs
mendess: This service is simply a proxy for kv. We are using kv just as much in this implementation as we were. I don't think it makes sense to make any architecture changes in this rewrite to keep it as simple and testable as possible.

daphne_worker/src/config.rs
```diff
@@ -856,9 +830,9 @@ impl<'srv> DaphneWorker<'srv> {
 
         let future_delete_durable = durable
             .post::<_, ()>(
-                BINDING_DAP_GARBAGE_COLLECTOR,
-                DURABLE_DELETE_ALL,
-                "garbage_collector".to_string(),
+                DoGarbageCollector::BINDING,
+                DoGarbageCollector::DeleteAll.to_route(),
+                DoGarbageCollector::name().unwrap_from_name(),
```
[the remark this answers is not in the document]
mendess: That's true but the reason it's designed this way is to cater to the daphne_server at the expense of some awkwardness on the `daphne_worker`, so I chose to do this weird unwraping pattern that should disappear once we drop the worker version

daphne_server/src/roles/aggregator.rs
```diff
@@ -66,21 +144,97 @@ impl<S: Sync> DapAggregator<S> for crate::App {
...
+        // subject are trusted.
+        if let Some(ref cf_tls_client_auth) = sender_auth.cf_tls_client_auth {
+            // TODO(cjpatton) Add support for TLS client authentication for non-Taskprov tasks.
+            let Some(ref taskprov_config) = self.service_config.taskprov else {
+                return Ok(Some(
+                    "TLS client authentication is currently only supported with Taskprov.".into(),
+                ));
+            };
+
+            // Check that that the certificate is valid. This is indicated bylLiteral "SUCCESS".
```
cjpatton, line 175:
```suggestion
            // Check that that the certificate is valid. This is indicated by literal "SUCCESS".
```

daphne_server/src/roles/aggregator.rs
```diff
@@ -66,21 +144,97 @@ impl<S: Sync> DapAggregator<S> for crate::App {
     async fn unauthorized_reason(
```
cjpatton, line 144: The implementation appears to be line-for-line the same as for daphne_worker. As we discussed previously, the TLS certificate bits are specific to Workers. In Oxy this is going to work differently, and I don't think we want to force ourselves to match Workers here. For example, `verified == "SUCCESS"` is a super awkward way to check a valid certificate.

I'd suggest that we put a pin in this. For now I would replace the TLS cert code path here with a `todo!()` so that we only support bearer tokens for now.

In any case, this code is very delicate. I think we need to refactor so that all of the code shared by daphne_worker and daphne_server is implemented by the same function.

cjpatton (commented):
Not quite all the way through but I wanted to dump my current comments so you can address them.

daphne_server/src/roles/aggregator.rs
```diff
@@ -93,17 +247,32 @@ impl<S: Sync> DapAggregator<S> for crate::App {
...
+        Ok(())
     }
 
     async fn get_task_config_for<'req>(
         &'req self,
         task_id: &'req TaskId,
     ) -> Result<Option<Self::WrappedDapTaskConfig<'req>>, DapError> {
-        todo!()
+        self.kv()
+            .get::<kv::prefix::TaskConfig>(task_id)
```
cjpatton, line 273: Does this check cache for (and hit KV in case of miss)?

daphne_server/src/roles/aggregator.rs
```diff
@@ -118,18 +287,93 @@ impl<S: Sync> DapAggregator<S> for crate::App {
...
+            .send()
+            .await
+            .map_err(|e| fatal_error!(err = ?e))?;
+
+        match res {
+            LeaderBatchQueueResult::Ok(batch_id) => Ok(batch_id),
+
+            // TODO spec: If we end up taking the current batch semantics of
+            // https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/313, then we'll need to
+            // define an error type for this case.
```
cjpatton, line 374: This has been settled.
```suggestion
            // TODO Return 400 Bad Request (with problem detail "empty batch queue") here instead 500 Internal Error
```

daphne_server/src/roles/aggregator.rs
```diff
@@ -141,7 +385,7 @@ impl<S: Sync> DapAggregator<S> for crate::App {
     }
 
     fn host(&self) -> &str {
-        todo!()
+        &self.service_config.env
```
cjpatton, line 388: This will usually "oxy", correct? Is that what we want, or should this be a proper hostname?

daphne_server/src/roles/aggregator.rs
```diff
@@ -155,7 +399,25 @@ impl DapReportInitializer for crate::App {
...
+                            &vdaf_verify_key,
+                            &vdaf_config,
+                            consumed_report,
+                        )
+                    })
+                    .collect::<Result<Vec<EarlyReportStateInitialized>, _>>()
+            }
+        })
+        .await
+        .map_err(|e| fatal_error!(err = ?e, "initialzing reports panicked"))?
```
cjpatton, line 420: It didn't really panic, did it?
```suggestion
        .map_err(|e| fatal_error!(err = ?e, "encountered error while initializing reports"))?
```

daphne_server/src/roles/aggregator.rs
```diff
@@ -184,6 +477,78 @@ impl HpkeDecrypter for crate::App {
...
+                    })
+            })
+            .await
+            .map_err(|e| fatal_error!(err = ?e))?
+            .ok_or(DapError::Transition(TransitionFailure::HpkeUnknownConfigId))?
+    }
+}
+
+#[async_trait]
+impl BearerTokenProvider for crate::App {
```
cjpatton, line 502: I think you already `use` this (if not, might as well?)
```suggestion
impl BearerTokenProvider for App {
```

daphne_server/src/roles/helper.rs
```diff
@@ -31,9 +49,32 @@ impl<S: Sync> DapHelper<S> for crate::App {
...
     {
-        self.worker
-            .durable_objects()
-            .get_helper_state(task_id, &agg_job_id.into())
+        let task_config = self
+            .get_task_config_for(task_id)
+            .await?
+            .ok_or(DapError::Abort(DapAbort::UnrecognizedTask))?;
+        // TODO(cjpatton) Figure out if retry is safe, since the request is not actually
+        // idempotent. (It removes the helper's state from storage if it exists.)
```
cjpatton, line 57: Worth double checking if this comment is still true.

daphne_server/src/router/mod.rs
```diff
@@ -99,28 +109,37 @@ impl AxumDapResponse {
...
-        let error = error.into();
-        let status = if matches!(error, DapError::Fatal(_)) {
+        // trigger abort if transition failures reach this point.
+        let error = match error.into() {
+            DapError::Transition(failure) => DapAbort::report_rejected(failure),
+            e @ DapError::Fatal(..) => Err(e),
+            DapError::Abort(abort) => Ok(abort),
+        };
+        let status = if let Err(_e) = &error {
+            // TODO(mendess)
```
cjpatton, line 125: Probably a good idea to tell your future self what you actually need to do here.

daphne_server/src/storage_proxy_connection/cache.rs
cjpatton, line 1: Is this module meant to be KV specific? Perhaps this module should be called `kv_cache`?

daphne_server/src/storage_proxy_connection/cache.rs
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+use std::{any::Any, collections::HashMap};
+
+use super::kv::KvPrefix;
+
+#[derive(Default, Debug)]
+pub struct Cache {
+    kv: HashMap<&'static str, HashMap<String, Box<dyn Any + Send + Sync + 'static>>>,
```
cjpatton, line 10: Document this structure. What are the keys for the outer hashmap and what are the keys for the inner hashmap?

daphne_server/src/roles/aggregator.rs
```diff
@@ -3,60 +3,138 @@
...
+        let durable = self.durable();
+
+        futures::stream::iter(agg_share_span)
+            .map(|(bucket, (agg_share, report_metadatas))| async {
+                let result = durable
+                    .request(
+                        bindings::AggregateStore::Merge,
+                        (task_config.version, &task_id_hex, &bucket),
+                    )
+                    .bin_encoding(AggregateStoreMergeReq {
```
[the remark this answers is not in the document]
mendess: Every DO expects the body of requests to use bincode

daphne_server/src/roles/aggregator.rs
mendess: It did actually panic. This error happens because whatever was passed into `tokio::task::spawn_blocking` panicked. This panic is caught and returned here in the Err variant

daphne_server/src/router/mod.rs
mendess: I did, it's to uncomment the line below. But I'll make it more obvious

daphne_server/src/roles/aggregator.rs
mendess: All `get`s go through the cache

daphne_server/src/roles/aggregator.rs
mendess: This method is for logging purposes only, but it's also one of the "loose ends" that still needs some design work.

daphne_server/src/storage_proxy_connection/cache.rs
mendess: Change the hierarchy so that the cache is a submodule of kv

cjpatton (approved):
Stamping this so that you can continue to make progress while I'm on PTO. This is heading in the right direction and is certainly good enough shape to iterate on and experiment with in Oxy.

Things to pay attention to:
1. I don't quite understand the tracing stuff, it would be good for someone to take a closer look at this.
2. The TLS client auth bits are copy-pasted from daphne_worker. It's going to end up looking a bit different ... I think we should just put a pin on this until we're closer to needing to turn this on.
3. Make sure DO requests are retryable. (I missed this in my initial pass but @mendess made me aware that this is not fully implemented.)
4. Sooner rather than later we should deploy daphne_worker in Testing and make sure we haven't regressed. I'd say after merging this would be a good time.


cc/ @captain-mota, who I assume will be taking over review from here.

PR #463, opened 2023-12-20
Refactor DO Bindings and extract them to utils

mendess:
Instead of using global constants to encode the various bindings and methods, define a more intuitive and typesafe api for interacting with DOs.

Also add an abstraction that makes use of these types to implement the communication with the future storage proxy

### DIFF
--- a/daphne_service_utils/src/durable_requests/mod.rs
+++ b/daphne_service_utils/src/durable_requests/mod.rs
@@ -1,0 +1,294 @@
+// Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+//! This module implements the communication between the daphne service and the storage proxy.
+//!
+//! Communication with the storage proxy happens through HTTP. The URI of the request is the URI
+//! to use when talking to the durable object, prefixed by [`DO_PATH_PREFIX`]. The body of this
+//! request will contain the metadata necessary for the storage proxy to locate the appropriate
+//! durable object, followed by the opaque body that should be forwarded to the durable object
+//! itself.
+//!
+//! # Metadata encoding
+//!
+//! The metadata is encoded using capnproto as the binary protocol, it contains:
+//! - The `binding` which identifies the durable object implementation to target.
+//! - The `id` which identifies the specific instance of the durable object.
+//!
+//! # Body encoding
+//!
+//! The body that follows this piece of metadata is opaque and to be understood by the Durable
+//! Object itself. So the proxy simply forwards it along without any processing.
+//!
+//! # The response
+//!
+//! The HTTP response is simply forwarded back from the Durable Object to the daphne service.
+//!
+//!```text
+//! +----------------+                 +---------------+            +----------------+
+//! | Native Service |                 | Storage Proxy |            | Durable Object |
+//! +----------------+                 +---------------+            +----------------+
+//!    | Http Request:                         |                            |
+//!    | POST /v1/`DO_PATH_PREFIX`/path/to/call   |                            |
+//!    | DurableRequest:                       |                            |
+//!    |   {binding, id}[body]                 |                            |
+//!    |-------------------------------------->| use `binding` and `id`     |
+//!    |                                       | to locate the durable      |
+//!    |                                       | object.                    |
+//!    |                                       |--------+                   |
+//!    |                                       |        |                   |
+//!    |                                       |        |                   |
+//!    |                                       |<-------+                   |
+//!    |                                       |                            |
+//!    |                                       |                            |
+//!    |                                       | Http Request:              |
+//!    |                                       | POST `/path/to/call`       | do some
+//!    |                                       | `body`                     | storage work
+//!    |                                       |--------------------------> |---------+
+//!    |                                       |                            |         |
+//!    |                                       |              Http Response |         |
+//!    |<--------------------------------------|<---------------------------|<--------+
+//!```
+
+pub mod bindings;
+
+use std::io;
+
+use bindings::DurableMethod;
+use daphne::messages::constant_time_eq;
+use durable_request_capnp::durable_request;
+use serde::{Deserialize, Serialize};
+
+use crate::durable_request_capnp;
+
+/// The base of a request path that points to a key in KV.
+pub const KV_PATH_PREFIX: &str = "/v1/kv";
+/// The base of a request path that points to a durable object.
+pub const DO_PATH_PREFIX: &str = "/v1/do";
+#[cfg(feature = "test-utils")]
+/// The path of the purge request, which wipes all storage. This is meant for tests only.
+pub const PURGE_STORAGE: &str = "/v1/purge";
+
+/// The way the target object's id will be obtained.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum ObjectIdFrom {
+    /// Derive a unique id from a name string.
+    ///
+    /// [see also](https://docs.rs/worker/latest/worker/durable/struct.ObjectNamespace.html#method.id_from_name)
+    Name(String),
+
+    /// Parse a previously stringified id from a hex string.
+    ///
+    /// [see also](https://docs.rs/worker/latest/worker/durable/struct.ObjectNamespace.html#method.id_from_string)
+    Hex(String),
+}
+
+impl ObjectIdFrom {
+    /// Assert that the object is being referenced by name.
+    ///
+    /// # Note
+    /// This is here for convenience of the daphne_worker and can be removed in the future.
+    pub fn unwrap_from_name(self) -> String {
+        let Self::Name(name) = self else {
+            panic!("unwraped a {self:?} when a Name was expected");
+        };
+        name
+    }
+
+    /// Assert that the object is being referenced by id.
+    ///
+    /// # Note
+    /// This is here for convenience of the daphne_worker and can be removed in the future.
+    pub fn unwrap_from_hex(self) -> String {
+        let Self::Hex(name) = self else {
+            panic!("unwraped a {self:?} when a Hex was expected");
+        };
+        name
+    }
+}
+
+/// A durable object request meant for the storage proxy.
+///
+/// It's generic over the payload which can be any type that implements AsRef<[u8]>. This payload
+/// is what is ultimately passed to the durable object.
+#[derive(Debug, Clone)]
+pub struct DurableRequest<P: AsRef<[u8]>> {
+    /// A binding name, as configured in the wrangler.toml.
+    ///
+    /// [see also](https://developers.cloudflare.com/durable-objects/get-started/#5-configure-durable-object-bindings)
+    pub binding: String,
+
+    /// The method by which the object id will be obtained in the storage proxy.
+    ///
+    /// [see also](https://developers.cloudflare.com/durable-objects/get-started/#4-instantiate-and-communicate-with-a-durable-object)
+    pub id: ObjectIdFrom,
+
+    /// The body of the request.
+    body: P,
+}
+
+// This needs to be very general to facilitate tests.
+impl<T, U> PartialEq<DurableRequest<U>> for DurableRequest<T>
+where
+    T: AsRef<[u8]>,
+    U: AsRef<[u8]>,
+{
+    fn eq(&self, other: &DurableRequest<U>) -> bool {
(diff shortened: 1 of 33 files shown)


daphne_worker/src/storage_proxy.rs
```diff
@@ -0,0 +1,213 @@
+// Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+//! This is a worker implementation that serves as a proxy to the storage that workers has access
```
cjpatton, line 4: nit: Worker/Workers are proper nouns
nit: More concise
```suggestion
//! This is a Worker that proxies requests to the storage that Workers has access
```

daphne_worker/src/storage_proxy.rs
```diff
@@ -0,0 +1,213 @@
...
+//! to, i.e., KV and Durable Objects.
+//!
+//! Comunication with this worker is done via http.
```
cjpatton, line 7:
```suggestion
//! Comunication with this Worker is done via HTTP.
```

daphne_worker/src/storage_proxy.rs
```diff
@@ -0,0 +1,213 @@
...
+//!
+//! # KV
+//!
+//! The base of all kv request uris is [`KV_PATH_BASE`].
```
cjpatton, line 11: nit: capitalize acronyms
nit: s/base/prefix? Feel free to ignore if you prefer
```suggestion
//! The prefix of all KV request URIs [`KV_PATH_BASE`].
```

daphne_worker/src/storage_proxy.rs
```diff
@@ -0,0 +1,213 @@
...
+//! ## Deleting a key
+//!
+//! Make a `DELETE` request with uri `{KV_PATH_BASE}/path/to/key`.
+//!
+//!
+//! # Durable Objects
+//!
+//! The base of all durable object requests is [`DO_PATH_BASE`].
+//!
+//! To interact with a durable object create an instance of [`DurableRequest`] which will be the
```
cjpatton, line 36:
```suggestion
//! To interact with a durable object, create an instance of [`DurableRequest`], which will be the
```

daphne_worker/src/storage_proxy.rs
```diff
@@ -0,0 +1,213 @@
...
+//!         &daphne::DapBatchBucket::TimeInterval { batch_window: 50 }
+//!     ),
+//! );
+//!
+//! let worker_url = Url::parse("https://example-worker.com")
+//!     .unwrap()
+//!     .join(uri)
+//!     .unwrap();
+//!
+//!
```
cjpatton, line 65:
```suggestion
```

daphne_worker/src/storage_proxy.rs
```diff
@@ -0,0 +1,213 @@
...
+//! let _send_future = reqwest::Client::new()
+//!     .post(worker_url)
+//!     .body(durable_request.into_bytes())
+//!     .send();
+//! ```
+//!
+//! [to_uri]: daphne_service_utils::durable_requests::bindings::DurableMethod::to_uri
+//!
```
cjpatton, line 73:
```suggestion
```

daphne_worker/src/storage_proxy.rs
```diff
@@ -0,0 +1,213 @@
+// Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
```
cjpatton, line 3: Beautiful!

daphne_worker/src/storage_proxy.rs
```diff
@@ -0,0 +1,213 @@
...
+use daphne_service_utils::durable_requests::{
+    DurableRequest, ObjectIdFrom, DO_PATH_BASE, KV_PATH_BASE,
+};
+use url::Url;
+use worker::{js_sys::Uint8Array, Env, Request, RequestInit, Response};
+
+const KV_BINDING_DAP_CONFIG: &str = "DAP_CONFIG";
+
+/// Handle a proxy request. This is the entry point of the worker.
```
cjpatton, line 83:
```suggestion
/// Handle a proxy request. This is the entry point of the Worker.
```

daphne_worker/src/storage_proxy.rs
```diff
@@ -0,0 +1,213 @@
...
+//!
+//! Comunication with this worker is done via http.
+//!
+//! # KV
+//!
+//! The base of all kv request uris is [`KV_PATH_BASE`].
+//!
+//! ## Getting a key
+//!
+//! Make a `GET` request with uri `{KV_PATH_BASE}/path/to/key`.
```
cjpatton, line 15: We should consider versioning the API, i.e., add to each request.  Here for example:
```suggestion
//! Make a `GET` request with uri `/storage_proxy/v1/{KV_PATH_BASE}/path/to/key`.
```

cjpatton (commented):
Beautiful work! I think the main thing that's left to do is to add a bit of versioning. I'm also not super happy with the custom serialization format for the storage proxy. Ideally we don't have a mixture of serialization formats unless we absolutely have to.

cjpatton (changes requested):
I meant to request changes to make sure we at least address versioning.

daphne_service_utils/src/durable_requests/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+    ///
+    /// [see also](https://docs.rs/worker/latest/worker/durable/struct.ObjectNamespace.html#method.id_from_string)
+    Hex(String),
+}
+
+impl ObjectIdFrom {
+    /// Assert that the object is being referenced by name.
+    ///
+    /// # Note
+    /// This is here for convenience of the daphne_worker and can be removed in the future.
```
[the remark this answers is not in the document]
mendess: Isn't that wrong? Two `for`s so close to eachother?

daphne_service_utils/src/durable_requests/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use daphne::{DapBatchBucket, DapVersion};
+
+    use crate::durable_requests::bindings::AggregateStore;
+
+    use super::{bindings, DurableRequest};
```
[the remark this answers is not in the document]
mendess: in the test module it's custom to use `super` as much as possible because you're testing the `super` module

daphne_service_utils/src/durable_requests/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+impl<T: AsRef<[u8]>, U: AsRef<[u8]>> PartialEq<Bytes<U>> for Bytes<T> {
+    fn eq(&self, other: &Bytes<U>) -> bool {
+        // safeguard against struct shape change
+        let Self(bytes) = self;
+        bytes.as_ref() == other.0.as_ref()
```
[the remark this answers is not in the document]
mendess: Made it constant time. It's only needed for tests but there's no harm in having it available for all as long as it's safe.

daphne_worker/src/durable/reports_pending.rs
```diff
@@ -4,58 +4,22 @@
...
-impl PendingReport {
-    pub(crate) fn report_id_hex(&self) -> Option<&str> {
-        match self.version {
-            DapVersion::Draft02 if self.report_hex.len() >= 96 => Some(&self.report_hex[64..96]),
-            DapVersion::DraftLatest if self.report_hex.len() >= 32 => Some(&self.report_hex[..32]),
-            _ => None,
-        }
-    }
-}
+use super::{worker_method_to_http, DapDurableObject, GarbageCollectable};
```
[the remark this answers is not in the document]
mendess: I prefer `super` when there is very close coupling between modules. As in, it's very unlikely that `reports_prending` wouldn't be a sub module of `durable`. This makes the import path more resistant to the `durable` module being moved.

This is all very *minor* stuff and I mostly leave it to rust-analyzer to pick.

daphne_service_utils/src/durable_requests/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// The http method used in the request.
+    #[serde(with = "serialize_method")]
+    pub method: http::Method,
+
+    /// The body of the request.
+    #[serde(skip)]
+    body: P,
+}
+
+// This needs to be very general to facilitate tests.
```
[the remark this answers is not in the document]
mendess: [it's here](https://github.com/cloudflare/daphne/blob/f82291bc7db43c815ba07fa8edf547f62b86fea3/daphne_service_utils/src/durable_requests/mod.rs#L367-L371)

```rs
let want: DurableRequest<Bytes<Vec<u8>>;
let got:  DurableRequest<Bytes<&[u8]>>;
```

daphne_service_utils/src/durable_requests/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+//! # Metadata encoding
+//!
+//! The metadata is encoded using capnproto as the binary protocol, it contains:
+//! - The `binding` which identifies the Durable Object implementation to target.
+//! - The `id` which identifies the specific instance of the durable object.
+//!
+//! # Body encoding
+//!
+//! The body that follows this piece of metadata is opaque and to be understood by the Durable
+//! Object itself. So the proxy shall simply forward it along without any processing
```
cjpatton, line 21:
```suggestion
//! Object itself. So the proxy simply forwards it along without any processing.
```

daphne_service_utils/src/durable_requests/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+//!
+//! # The response
+//!
+//! The HTTP response is simply forwarded back from the Durable Object to the Native Service.
```
cjpatton, line 25: "Native Service": "native" means "bare metal", as in "not on Workers", right? Perhaps "DAP Aggregator" would be better? Or "Edge Service" if you want to make it all Cloudflarey

daphne_service_utils/src/durable_requests/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+//!
+//!```text
+//! +----------------+                 +---------------+            +----------------+
+//! | Native Service |                 | Storage Proxy |            | Durable Object |
+//! +----------------+                 +---------------+            +----------------+
+//!    | Http Request:                         |                            |
+//!    | POST /`DO_PATH_PREFIX`/path/to/call   |                            |
```
cjpatton, line 32: Suggest versioning this in case we need to change something later on.
```suggestion
//!    | POST /v1/`DO_PATH_PREFIX`/path/to/call   |                            |
```

daphne_service_utils/src/durable_requests/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+//! Communication with the storage proxy happens through HTTP. The URI of the request is the URI
+//! to use when talking to the durable object, prefixed by [`DO_PATH_PREFIX`]. The body of this
+//! request will contain the metadata necessary for the storage proxy to locate the appropriate
+//! durable object, followed by the opaque body that should be forwarded to the durable object
+//! itself.
+//!
+//! # Metadata encoding
+//!
+//! The metadata is encoded using capnproto as the binary protocol, it contains:
+//! - The `binding` which identifies the Durable Object implementation to target.
```
cjpatton, line 15: nit: Consistency
```suggestion
//! - The `binding` which identifies the durable object implementation to target.
```

daphne_service_utils/src/durable_requests/mod.rs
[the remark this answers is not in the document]
cjpatton: Agreed.

daphne_service_utils/src/durable_requests/mod.rs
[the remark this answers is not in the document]
cjpatton: OOooohhhh that makes sense! Thanks

cjpatton (approved):
Looks good to me! Great work.

My one suggestion would be to add a version prefix to the *path* of the request from the edge service to the storage proxy. capnproto only solves the versioning problem if we decide to always use capnproto. It's still a good idea I think to version the path so that we can change to something else if we needed to. (Hopefully we don't.)

PR #464, opened 2023-12-20
Add collection_job_id to DapRequest

mendess:
This is needed in order for the `DapRequestExtractor` to provide a way for draft02 requests to access the collection_job_id, this is because [this route](https://github.com/cloudflare/daphne/blob/890f8a0170cd24c10f90f829cfd4b2a0b9d1f8de/daphne_worker/src/router/leader.rs#L83-L90) manually parses the collection job id for draft 02.

By adding this field we can make remove some duplicate code in the server implementation.

### DIFF
--- a/daphne/src/lib.rs
+++ b/daphne/src/lib.rs
@@ -1099,8 +1099,10 @@ pub enum DapResource {
     /// The resource of a DAP request is undetermined if its identifier could not be parsed from
     /// request path.
     ///
-    /// draft02 compatibility: In draft02, the resource of a DAP request is undetermined until the
-    /// request payload is parsed. Defer detrmination of the resource until then.
+    /// draft02 compatibility: In draft02, the resource of a DAP request is usually undetermined
+    /// until the request payload is parsed, so defer determination of the resource until then. The
+    /// exception is the GET request to the collection URI. The format of this endpoint is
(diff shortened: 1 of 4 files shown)


cjpatton:
Note to self: Make sure it's safe for draft02 to set `DapResource::CollectionJob`.

Cargo.lock
```diff
@@ -636,7 +636,7 @@ dependencies = [
 
 [[package]]
 name = "daphne_server"
-version = "0.1.0"
+version = "0.3.0"
```
cjpatton, line 639: nit: Bump the version in a seaparte commit

cjpatton (approved):
I appreciate the care you've taken in not to change the semantics of `DapResource`. However in this case I think we're OK to do so.

In either draft02 or draft09, the collect URI (the endpoint the Collector polls to get the result of the collection job) encodes the collection job ID. In fact, the format of the collect URI is the same in both versions. (In draft02 the format is unspecified, but our implementation matches the latest version).

One important thing to keep in mind: in draft09, the Collector sets the collection job ID in the request path; in draft02, the Leader sets the collection job ID in the response. In the former case, `DapResource` should be set; but in the latter it should be undefined.

I've tacked on a commit that sets `DapResource` in draft02. With that I think this is OK to merge.

PR #465, opened 2023-12-20
Split DaphneWorkerConfig into DaphneWorkerConfig and DaphneWorkerDoConfig

mendess:
To run exclusively DOs we don't need to load all the configuration values needed for a full worker. As such we split these two in order to make configuring the storage proxy worker simpler.

### DIFF
--- a/daphne_server/examples/configuration.toml
+++ b/daphne_server/examples/configuration.toml
@@ -1,9 +1,28 @@
+storage = "http://localhost:4000"
 port = 3000
-#role = "helper"
 
-[global]
+[service]
+env = "oxy"
+role = "helper"
 max_batch_duration = 360000
 min_batch_interval_start = 259200
 max_batch_interval_end = 259200
 supported_hpke_kems = ["x25519_hkdf_sha256"]
+report_shard_key = "2e0a99e5a047510243e08289811db0386c620644ae53bafdd50604678f3ec537" # SECRET
+report_shard_count = 4
+default_version = "v09"
+report_storage_epoch_duration = 300000
+base_url = "http://127.0.0.1:8787"
 allow_taskprov = true
+
+[service.taskprov]
+vdaf_verify_key_init = "b029a72fa327931a5cb643dcadcaafa098fcbfac07d990cb9e7c9a8675fafb18" # SECRET
+leader_auth.bearer_token = "I am the leader!" # SECRET
+collector_auth.bearer_token = "I am the collector!" # SECRET
+
+[service.taskprov.hpke_collector_config]
+id = 23
+kem_id = "p256_hkdf_sha256"
+kdf_id = "hkdf_sha256"
+aead_id = "aes128_gcm"
+public_key = "047dab625e0d269abcc28c611bebf5a60987ddf7e23df0e0aa343e5774ad81a1d0160d9252b82b4b5c52354205f5ec945645cb79facff8d85c9c31b490cdf35466"
--- a/daphne_service_utils/src/config.rs
+++ b/daphne_service_utils/src/config.rs
@@ -1,0 +1,83 @@
+// Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+use daphne::{
+    hpke::{HpkeConfig, HpkeReceiverConfig},
+    DapGlobalConfig, DapVersion,
+};
+use serde::{Deserialize, Serialize};
+use url::Url;
+
+use crate::{auth::DaphneWorkerAuthMethod, DapRole};
+
+/// draft-wang-ppm-dap-taskprov: Long-lived parameters for the taskprov extension.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct TaskprovConfig {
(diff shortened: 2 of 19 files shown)


daphne_server/examples/configuration.toml
```diff
@@ -1,9 +1,28 @@
...
 
-[global]
+[service]
+env = "oxy"
+role = "helper"
 max_batch_duration = 360000
 min_batch_interval_start = 259200
 max_batch_interval_end = 259200
 supported_hpke_kems = ["x25519_hkdf_sha256"]
+report_shard_key = "2e0a99e5a047510243e08289811db0386c620644ae53bafdd50604678f3ec537"
```
cjpatton, line 11:
```suggestion
report_shard_key = "2e0a99e5a047510243e08289811db0386c620644ae53bafdd50604678f3ec537" # SECRET
```

daphne_server/examples/configuration.toml
cjpatton, line 1: (Initially discussed in #460.) In a real deployment we'll need to be able to make changes to this file without having to touch secrets. What is our plan for this?

For daphne_worker the story is pretty simple: For testing we can leave the "secrets" in wrangler.toml; in the deployed code, we set the secret with `wrangler secret`. This was possible because the Worker treats them like environment variables either way. It doesn't seem like this could work this same way.

I'm not sure how the new service infra works, but if this were a standard edge service, we would do one of two things:
1. Store each secret in a file and load them either here (with some toml magic) or on startup.
2. At startup, request the keys from some key management service, like Quicksilver.

I'm not sure we should land this code until we've figured this out.

daphne_server/examples/configuration.toml
```diff
@@ -1,9 +1,28 @@
...
+report_shard_count = 4
+default_version = "v09"
+report_storage_epoch_duration = 300000
+base_url = "http://127.0.0.1:8787"
 allow_taskprov = true
+
+[service.taskprov]
+vdaf_verify_key_init = "b029a72fa327931a5cb643dcadcaafa098fcbfac07d990cb9e7c9a8675fafb18" # SECRET
+leader_auth.bearer_token = "I am the leader!" # SECRET
```
cjpatton, line 20: If feasible, I'd suggest using values here that look more like bearer tokens. I believe [something like](https://swagger.io/docs/specification/authentication/bearer-authentication/)
```
Bearer <randdom 16 byte hex string>
```

daphne_service_utils/src/config.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// Leader, Helper: Method for authorizing Leader requests.
+    pub leader_auth: DaphneWorkerAuthMethod,
+
+    /// Leader: Method for authorizing Collector requests.
+    pub collector_auth: Option<DaphneWorkerAuthMethod>,
+}
+
+pub type HpkeRecieverConfigList = Vec<HpkeReceiverConfig>;
+
+/// Daphne-Worker configuration, including long-lived parameters used across DAP tasks.
```
cjpatton, line 32: Perhaps we want a generic term that applies to either daphne_worker or daphne_server
```suggestion
/// Daphne server configuration, including long-lived parameters used across DAP tasks.
```

daphne_worker/src/durable/reports_pending.rs
```diff
@@ -2,15 +2,18 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 use crate::{
-    config::DaphneWorkerConfig,
+    config::DaphneWorkerDoConfig,
```
cjpatton, line 5: Consistency
```suggestion
    config::DaphneWorkerDurableConfig,
```

cjpatton (changes requested):
This is an improvement over all, but it creates a new problem: we now mingle secrets with non-secret configuration parameters in the same configuration file, configuration.toml. (Previously this was not an issue because Workers treats both as environment variables.)

That said, I appreciate if this is intended as an incremental step and you'd like to address this problem later.

daphne_server/examples/configuration.toml
cjpatton: Update: @mendess clarified that this config mechanism is for the example only -- in the real deployment we'll have to use Vault any way, which means something like this won't be an option.

daphne_server/examples/configuration.toml
cjpatton: Ignore this since it would break the end to end tests

cjpatton (commented):
I've resolved a couple of comments based on our discussion today.

mendess:
> This is an improvement over all, but it creates a new problem: we now mingle secrets with non-secret configuration parameters in the same configuration file, configuration.toml. (Previously this was not an issue because Workers treats both as environment variables.)
> 
> That said, I appreciate if this is intended as an incremental step and you'd like to address this problem later.

This configuration file is just an example, all of the values that are present in it can be passed through environment variables as well. For example, the `report_shard_key` can be provided through the `DAP_SERVICE_REPORT_SHARD_KEY` environment variable in a production setting. Either way the `examples/service.rs` is just for e2e testing and is not meant to mimic a real production server

cjpatton (approved):

PR #466, opened 2023-12-22
Implement DO request retrying in the storage proxy

mendess:


oliy (approved):

oliy:
Already discussed, but only concern was around considering retry: true as the default.

cjpatton (approved):

Issue #437, opened 2023-11-27
Upgrade to the latest version of `prio`


cjpatton:
We're in the midst of upgrading `prio` to the latest VDAF draft: https://github.com/divviup/libprio-rs/issues/839

The next major release will include support for the new draft. In the meantime, we're checking out specific revisisions on main as new features are implemented.

This is issue is to track pointing to the new release. See https://github.com/cloudflare/daphne/pull/435#discussion_r1402510306_

cjpatton:
Next version will be 0.16.0 and will include a variety of breaking changes. Namely:

- [ ] "prio2" feature moved to "experimental"
- [ ] `encode()` is now fallible
- [ ] Implementation of `XofHmacSha256Aes128` was upstreamed
- [ ] Other stuff I don't know off the top of my head.

PR #473, opened 2024-01-14
Upgrade prio to 0.16.0

cjpatton:
Partially addresses #437.


daphne_worker/src/tracing_utils/wasm_timing.rs
```diff
@@ -38,6 +38,7 @@ impl TryFrom<&str> for MeasuredSpanName {
     }
 }
 
+#[allow(clippy::struct_field_names)]
```
cjpatton, line 41: All of the fields in this struct have the same suffix.

daphne/src/vdaf/mod.rs
```diff
@@ -364,7 +366,7 @@ where
     S: Serializer,
     T: Encode,
 {
-    s.serialize_str(&hex::encode(x.get_encoded()))
+    s.serialize_str(&hex::encode(x.get_encoded().expect("XXX")))
```
cjpatton, line 369: @mendess how should we handle this encoding failure? Previously we would have panicked.

daphne_server/src/router/leader.rs
```diff
@@ -145,7 +145,9 @@ where
             daphne::DapResponse {
                 version: req.version,
                 media_type: DapMediaType::Collection,
-                payload: collect_resp.get_encoded_with_param(&req.version),
+                payload: collect_resp
+                    .get_encoded_with_param(&req.version)
+                    .expect("XXX"),
```
cjpatton, line 150: @mendess how should we handle this encoding failure? Previously we would have panicked.

daphne_server/src/router/mod.rs
```diff
@@ -445,7 +445,7 @@ mod test {
                 .uri("/v02/parse-version")
                 .body({
                     let mut v = Vec::new();
-                    task_id.encode(&mut v);
+                    task_id.encode(&mut v).expect("XXX");
```
cjpatton, line 448: @mendess how should we handle this encoding failure? `.unwrap()` is probably fine here since `Id::encode()` is actually infallible.

daphne_worker/src/router/leader.rs
```diff
@@ -97,7 +97,7 @@ pub(super) fn add_leader_routes(router: DapRouter<'_>) -> DapRouter<'_> {
                     Ok(DapCollectJob::Done(collect_resp)) => dap_response_to_worker(DapResponse {
                         version: DapVersion::Draft02,
                         media_type: DapMediaType::Collection,
-                        payload: collect_resp.get_encoded_with_param(&version),
+                        payload: collect_resp.get_encoded_with_param(&version).expect("XXX"),
```
cjpatton, line 100: @mendess how should we handle this encoding failure? Previously we would have panicked.

daphne_worker/src/router/leader.rs
```diff
@@ -172,7 +172,9 @@ pub(super) fn add_leader_routes(router: DapRouter<'_>) -> DapRouter<'_> {
                     Ok(DapCollectJob::Done(collect_resp)) => dap_response_to_worker(DapResponse {
                         version: req.version,
                         media_type: DapMediaType::Collection,
-                        payload: collect_resp.get_encoded_with_param(&req.version),
+                        payload: collect_resp
+                            .get_encoded_with_param(&req.version)
+                            .expect("XXX"),
```
cjpatton, line 177: @mendess how should we handle this encoding failure? Previously we would have panicked.

daphne/src/error/mod.rs
```diff
@@ -57,15 +58,27 @@ impl FatalDapError {
...
+        // Assume this an encoding error and treat as fatal.
+        match e {
+            CodecError::Other(..) => {
+                DapError::Fatal(FatalDapError(format!("encoding failed: {e}")))
+            }
+            unhandled => unreachable!("unhandled error: {unhandled}"),
+        }
+    }
+}
+
```
mendess, line 79: if you're assuming something here than this `From` implementation is only correct for as long as that comment there stays true. Meaning it's holding an invariant over the entire codebase. I'd rather you make something like
```rs
impl DapError {
    fn encoding(e: CodecError) -> Self { ... }
}
```

mendess: Also, why is only the `Other` variant handled?

daphne/src/vdaf/mod.rs
mendess: `.map_err(|e| S::Error::custom(e))?)`

daphne_server/src/router/leader.rs
mendess, line 150:
```suggestion
                payload: match collect_resp.get_encoded_with_param(&req.version) {
                    Ok(payload) => payload,
                    Err(e) => return AxumDapResponse::new_error(DapError::encoding(e), app.server_metrics()),
                }
```

daphne_server/src/router/mod.rs
mendess: prefer unwrapping in tests if you don't expect the call to fail. Makes the tests more useful as they are more likely to panic in the event of code changes

daphne_worker/src/router/leader.rs
mendess: See previous suggestion on the other routing method. But here use `daph.state.dap_abort_to_worker_response(e)`

daphne_worker/src/router/leader.rs
mendess: same as other routing method

mendess (changes requested):

daphne_server/src/router/leader.rs
cjpatton: Bump :)

daphne/src/error/mod.rs
cjpatton: **Much better**, done!

daphne_server/src/router/leader.rs
mendess: I had already answered here https://github.com/cloudflare/daphne/pull/473#discussion_r1452172468

cjpatton: Yup, I saw! I resolved this comment but also should have commented. Sorry about that

mendess (approved):

PR #476, opened 2024-01-19
Add e2e for daphne_server to CI

mendess:


daphne_server/docker-compose-e2e.yaml
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) 2022 Cloudflare, Inc. All rights reserved.
```
cjpatton, line 1:
```suggestion
# Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
```

daphne_server/docker/example-service.Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+        capnproto
+
+WORKDIR /dap
+
+COPY Cargo.toml Cargo.lock .
+COPY daphne daphne
+COPY daphne_service_utils daphne_service_utils
+COPY daphne_server daphne_server
+RUN cargo new --lib daphne_worker
+RUN cargo new --lib daphne_worker_test
```
cjpatton, line 14: As I understand it, creating this empty lib is necessary because the workspace defines these crates. This is a little awkwward -- it would be nice if there was a cleaner way to do this.

cjpatton (approved):
Please add more context to the commit message. You answer the question: "Does CI now run the end to end test against daphne-server?"

daphne_server/docker/example-service.Dockerfile
mendess: we either sed script away the specification (tried it, went over badly) or we copy each one over just for it to be unused or we do this, either way, the number of lines in the dockerfile is the same and this way we can clearly see which ones matter to the build and which ones don't

mendess:
> Please add more context to the commit message. You answer the question: "Does CI now run the end to end test against daphne-server?"

I don't understand what kind of context you think should be added. I've added some but I felt the message was pretty self describing :thinking:

PR #478, opened 2024-01-22
Move the legacy worker implementation to a submodule

mendess:
Move the legacy worker into it's own module to make cleanup later easier to do.


daphne_worker/src/legacy/mod.rs
cjpatton, line 1: Missing copyright

mendess: done

daphne_worker_test/wrangler.storage_proxy.toml
```diff
@@ -15,7 +15,7 @@ fallthrough = false
 # as they are here. See
 # https://developers.cloudflare.com/workers/wrangler/commands/#secret.
 [vars]
-DAP_PROXY = "true"
+MODE = "storage-proxy"
```
cjpatton, line 18: I think environment variables specific to our service should be named as such.
```suggestion
DAP_WORKER_MODE = "storage-proxy"
```

daphne_worker/src/legacy/mod.rs
```diff
@@ -0,0 +1,125 @@
+// Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
```
cjpatton, line 3: Add a docucomment saying what this is. (It's the old daphne_worker.)

daphne_worker/src/legacy/mod.rs
cjpatton, line 1: Instead of `legacy` I'd give this a distinctive name that reflects what this used to do and add `#[deprecated]` to the module.

mendess: It makes sense, but that will just produce warnings and not be very useful because we are eventually going to remove it

cjpatton (approved):

PR #490, opened 2024-02-01
Add metric to track the reason a request was unauthorized

mendess:


cjpatton (commented):
Why add the failure "reason" to metrics? An alternative would be to just count the total number of authorization failures and look at the logs to understand what's happening.

We need to be careful here. Prometheus is not meant to handle a large number of labels: if the "reason" is just a string, then the number of distinct labels is potentially huge. Imagine for example that we added the hash of the the client's TLS certificate to this string.

If adding the reason to metrics is necessary, then I would recommend we replace the "failure reason" `String` with an `enum FailureReason`. It should also be non-exhaustive, in case we want to define variants in the downstream repo. That way way it's harder for us to blow up prometheus.

mendess:
cc @cbranch since this is technically his code :p

cbranch:
Thinking harder about this, the specific reason isn't actionable unless you've made a change; the client can cause any of these reasons at any time, so why bother distinguishing the cases? If you are deploying a new change and there is a problem with authorization, logs are sufficient.

daphne_server/src/roles/aggregator.rs
```diff
@@ -144,18 +144,22 @@ impl DapAggregator<DaphneAuth> for crate::App {
         task_config: &DapTaskConfig,
         req: &DapRequest<DaphneAuth>,
     ) -> Result<Option<String>, DapError> {
+        let unauth_reason = |reason: String| {
+            self.metrics().unauthorized_request(&reason);
+            Ok(Some(reason))
+        };
```
cbranch, line 150: Alternatively we could wrap this entire function in another function that increments the metric when we have `Ok(Some(x))`. The goal is to avoid having to increment the metric at each callsite.

mendess:
Closing this since:
- we already have a metric counting aborts, which counts unauthorized requests
- we already log aborts, which includes unauthorized requests

PR #496, opened 2024-02-02
Reject reports that fall outside the allowed time span

mendess:


daphne_server/src/roles/aggregator.rs
```diff
@@ -396,19 +396,31 @@ impl DapReportInitializer for crate::App {
...
-                            consumed_report,
-                        )
+                        let metadata = consumed_report.metadata();
+                        let initialized = if metadata.time < valid_report_range.start {
+                            consumed_report
+                                .into_initialized_rejected_due_to(TransitionFailure::ReportDropped)
+                        } else if metadata.time > valid_report_range.end {
+                            consumed_report
+                                .into_initialized_rejected_due_to(TransitionFailure::ReportTooEarly)
+                        } else {
```
cjpatton, line 415: Do we also need to check if the timestamp is passed the tasks expiration date, or do we check this elsewhere?

daphne_server/src/roles/aggregator.rs
```diff
@@ -396,19 +396,31 @@ impl DapReportInitializer for crate::App {
...
-                            &vdaf_config,
-                            consumed_report,
-                        )
+                        let metadata = consumed_report.metadata();
+                        let initialized = if metadata.time < valid_report_range.start {
+                            consumed_report
+                                .into_initialized_rejected_due_to(TransitionFailure::ReportDropped)
+                        } else if metadata.time > valid_report_range.end {
+                            consumed_report
+                                .into_initialized_rejected_due_to(TransitionFailure::ReportTooEarly)
```
cjpatton, line 414: This code looks correct, but it took me a while to see what's going on. (If the timestamp is after the end of the validity range, it seems like it would be "too late" rather than "too early". But it's clear that "too early" is correct once you understand how the `valid_report_range` is computed.) Can you add a comment or comments here helping the reader understanding which reports we're rejecting?

Also, nit: Does it make sense to do this check earier, perhaps before calling initialize reports?

daphne_server/src/roles/mod.rs
```diff
@@ -1,10 +1,26 @@
...
 
+impl crate::App {
+    pub(crate) fn valid_report_time_range(&self) -> Range<u64> {
+        let now = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("now should always be after unix epoch")
+            .as_secs();
+
+        let start = now.saturating_sub(self.service_config.report_storage_epoch_duration);
+        let end = now.saturating_add(self.service_config.report_storage_max_future_time_skew);
```
cjpatton, line 18: What is a reasonable value for `report_storage_max_future_time_skew`?

cjpatton (approved):
Looks correct, but we can make the code a little more clear.

PR #502, opened 2024-02-19
Plumb aggregation parameter through the code

### LINKED ISSUE #25: Scaling up the Leader

### LINKED ISSUE #68: Leader: Consider flushing all reports before responding to CollectReq
This came up when testing against Janus. Their tests expect that the Leader waits to respond to a collect request until are reports pertaining to that request are processed. While not required by spec, this might be a useful thing to consider implementing. One thing to consider is that reports pertaining to the batch may be uploaded at the same time, so we'll need to figure out a cut-off mechanism.

cjpatton:
Based on #500 (merge that first).
Closes #25.
Closes #68.

This PR has three major changes in it.

**First commit**

Replace the hard-coded aggregation parameter for testing Mastic with one plumbed through the collection job. The change to the protocol code is relatively minor, but it entails a major rework of the Leader architecture.

The current architecture assumes there is no aggregation parameter, hence aggregation can begin as soon as reports arrive. Aggregation job creation is currently facilitated by `DapLeader::get_reports()`. The Leader begins its processing loop (`leader::process()`) by calling this function to get a set of reports to aggregate. It then issues a number of aggregation jobs, then cycles through the queue of collection jobs to complete any that can be completed.

This assumption no longer: with Mastic, the Leader must wait for a collection job before it can begin scheduling aggregation jobs.

The new processing loop is centered around an object called the "work queue". The work queue is a mixture of aggregation jobs and collection jobs. The queue is manipulated with `DapLeader::enqueue_work()` and `DapLeader::dequeue_work()`. The processing loop begins by dequeueing a number of work items and processing them in order.

`MockAggregator` has been updated to wait to aggregate reports until a collection job is created.

* Various minor improvements:

    * Clean up some buggy protocol logic around the creation of batches
      for fixed-size tasks. In particular, don't override the "current
      batch" query of the Collector's request.

    * Align input type of `DapAbort::batch_overlap()` with
      `query_mismatch()`

    * Add the batch ID to implementation of `Display` for `BatchSelector`

    * Implement `Display` for `Query`

    * Test multiple tasks in flight at once for `MockAggregator`

    * Align variable names for collection jobs with the latest draft

    * Resolve some TODOs

**Second commit**

Align `daphne_server`'s Leader with new work queue semantics

Update `daphne_server` to match new the changes introduced by the
previous commit. Right now, `DapLeader` for `daphne_server::App` is used
for testing only. Thus, to keep this change as simple as possible,
refactor `daphne::MockAggregator` to share code with `daphne_server::App`.

As a result, most of the Leader's operating memory, including the work
queue, pending reports, and pending collection jobs, is no longer stored
in durable objects. This is good enough for now, but once it's time to
implement the production Leader, we'll need to devise a storage scheme,
using durable objects or something else.

Minor changes:

* Since removing the "report selector", the Leader can no longer control
  the "aggregation rate" in the way it did before.

* `DapLeader::put_report()` no longer tries to detect replays.
  Previously it was "best effort" and didn't catch all edge cases. (The
  definitive replay check happens when running an aggregation job.)

* Move `DapAggregator::current_batch()` to `DapLeader::current_batch()`
  (the Helper doesn't need to implement this)

* Add reports processed back to the telemtry returned by
  `leader::process()` (this was erroneously removed from the previous
  commit)

**Third commit**

Delete the DO code we no longer need.


daphne/src/messages/mod.rs
```diff
@@ -1378,24 +1402,30 @@ pub fn decode_base64url_vec<T: AsRef<[u8]>>(input: T) -> Option<Vec<u8>> {
...
+            bytes[len_offset..len_offset + len_bytes]
+                .copy_from_slice(&$type::to_be_bytes(len.try_into().unwrap()));
+            Ok(())
+        }
+    };
 }
 
+make_encode_len_prefixed!(u16, encode_u16_prefixed);
+make_encode_len_prefixed!(u32, encode_u32_prefixed);
+
```
mendess, line 1428: This can probably be implemented as a generic function
```rust
where
  T: Default + Encode
  ```
and instead of using std::mem::size_of you can do
```rust
let len_bytes = bytes.len() - len_offset;
```

mendess: thinking back on it you can't be generic over `to_be_bytes` but at the same time, should you be using that at all? Shouldn't you be using `len.encode`? I understand that the `encode` api doesn't let you encode in the middle, but at the same time, doesn't this code only work for as long as the encode implementation of u16 and u32 doesn't change?

cjpatton: That might work, I'll give it a shot.

daphne/src/messages/mod.rs
```diff
@@ -522,20 +521,6 @@ impl Default for BatchSelector {
...
-    fn try_from(query: Query) -> Result<Self, DapError> {
-        match query {
-            Query::TimeInterval { batch_interval } => Ok(Self::TimeInterval { batch_interval }),
-            Query::FixedSizeByBatchId { batch_id } => Ok(Self::FixedSizeByBatchId { batch_id }),
-            Query::FixedSizeCurrentBatch => Err(fatal_error!(
-                err = "tried to make a BatchSelector from a FixedSizeCurrentBatch query",
-            )),
-        }
-    }
-}
```
cjpatton, line 534: Reviewer note: We expect to hit this error even when things are fine, but `fatal_error` logs the error. I've replaced with `Query::into_batch_sel()`, defined below.

daphne_worker_test/tests/e2e/e2e.rs
```diff
@@ -191,18 +190,6 @@ async fn leader_upload(version: DapVersion) {
     )
     .await;
 
-    // Try uploading the same report a second time (expect failure due to repeated ID.
```
cjpatton, line 194: Reviewer note: Replay is no longer enforced here. The Leader will them during an aggregation job.

daphne_worker_test/tests/e2e/e2e.rs
```diff
@@ -499,79 +479,46 @@ async fn internal_leader_process(version: DapVersion) {
         .await;
     }
 
-    let agg_telem = t.internal_process(&client, &report_sel).await;
+    let collect_req = CollectionReq {
```
cjpatton, line 482: Reviewer note: The collection job request needs to be initialized before aggregation jobs start getting queued.

daphne_worker_test/tests/e2e/e2e.rs
```diff
@@ -499,79 +479,46 @@ async fn internal_leader_process(version: DapVersion) {
...
 
     // There should be nothing left to aggregate.
-    let agg_telem = t.internal_process(&client, &report_sel).await;
+    let agg_telem = t.internal_process(&client).await;
     assert_eq!(agg_telem.reports_processed, 0, "reports processed");
     assert_eq!(agg_telem.reports_aggregated, 0, "reports aggregated");
     assert_eq!(agg_telem.reports_collected, 0, "reports collected");
 }
 
 async_test_versions! { internal_leader_process }
```
cjpatton, line 521: Reviewer note: This test no longer applies because there is no report selector.

daphne/src/messages/mod.rs
mendess: yeah, I agree with the downsides you've listed, they might not be worth it. Keep the macro for now

cjpatton:
> Make sure to remove unused bindings from here as well https://github.com/cloudflare/daphne/blob/main/daphne_service_utils/src/durable_requests/bindings.rs

Done, and rebased on #500.

mendess (approved):

PR #505, opened 2024-02-22
Reduce number of report share clones

mendess:


daphne/src/protocol/aggregator.rs
```diff
@@ -81,15 +81,13 @@ pub trait EarlyReportState {
...
-#[serde(rename_all = "snake_case")]
+#[derive(Clone)]
 pub enum EarlyReportStateConsumed {
     Ready {
         metadata: ReportMetadata,
-        #[serde(with = "serialize_bytes")]
         public_share: Vec<u8>,
-        #[serde(with = "serialize_bytes")]
         input_share: Vec<u8>,
+        draft_latest_payload: Option<Vec<u8>>,
```
cjpatton, line 90: What is this value, and why is it only set in the latest draft?

cjpatton: Ah, right this is "initial prep message". Consider giving it a more meaningful name, like "draft_latest_prep_init_payload".

In any case, add a code comment explaining what this is and why it's not needed in draft02.

cjpatton (approved):

cjpatton:
> Okay so, new idea. Instead of tacking on a new field I made a `ReportState` struct that delimits the values that need to be threaded through the entire flow from start to finish. This way it's more obvious what 's the main concern of the `EarlyReportState*` enums (the cypher text) and what is auxiliary data that must be passed along. Tell me what you think

SGTM 👍

PR #511, opened 2024-02-26
Check that code builds with multiple combinations of features

mendess:


daphne/src/mock_leader_memory.rs
cjpatton, line 1: Add copyright

cjpatton, line 1: Instead of creating a new module for this, what about sticking it in `src/roles/leader.rs`?

daphne/src/mock_leader_memory.rs
```diff
@@ -0,0 +1,264 @@
...
+use crate::{
+    error::DapAbort,
+    fatal_error,
+    messages::{Base64Encode, BatchId, BatchSelector, Collection, CollectionJobId, Report, TaskId},
+    roles::leader::WorkItem,
+    DapAggregationParam, DapBatchBucket, DapCollectionJob, DapError, DapQueryConfig, DapTaskConfig,
+};
+
+#[derive(Default)]
+pub struct MockLeaderMemory {
```
cjpatton, line 22: As discussed earlier today. Don't forget to rename the module as well. 
```suggestion
pub struct InMemoryLeaderState {
```

cjpatton (approved):
Looks good as-is, but:
1. Rename the struct as discussed
2. Consider sticking this in the daphne::roles::leader module instead of creating a new module for it.

PR #516, opened 2024-03-01
Reduce peak memory usage in aggregator protocol functions

mendess:
By making sure the pre allocated vectors and hashmaps/sets are live as
little as possible, as well as, delaying the pre allocation of these we
can reduce the peak allocated memory of the system while running the
protocol.


cjpatton (approved):
Nice idea!

PR #520, opened 2024-03-04
Add hpke config signing

mendess:


.github/workflows/daphneci.yml
```diff
@@ -39,8 +39,35 @@ jobs:
...
+          } >> "$GITHUB_OUTPUT"
+
       - name: Run integration tests
         uses: isbang/compose-action@v1.4.1
         with:
           compose-file: "./daphne_server/docker-compose-e2e.yaml"
           up-flags: "--build --abort-on-container-exit --exit-code-from test"
+        env:
+          DAP_SIGNING_KEY: ${{ steps.hpke_signing_key.outputs.hpke_signing_key }}
+          E2E_HPKE_SIGNING_CERTIFICATE: ${{ steps.hpke_signing_cert.outputs.hpke_signing_cert }}
```
cjpatton, line 73: Since these are conceptually related, they should have similar names. Also "E2E" is a little vague.
```suggestion
          DAP_SIGNING_KEY: ${{ steps.hpke_signing_key.outputs.hpke_signing_key }}
          E2E_TEST_DAP_SIGNING_CERTIFICATE: ${{ steps.hpke_signing_cert.outputs.hpke_signing_cert }}
```

daphne_worker_test/tests/e2e/test_runner.rs
```diff
@@ -254,6 +254,7 @@ impl TestRunner {
             res.status, res.error
         );
 
+        println!("############ starting test ############");
```
mendess, line 257: this println is intentional. It makes parsing the test output a lot easier by separating logs produced by the initialization of the test from logs produced by the test itself

.github/workflows/daphneci.yml
```diff
@@ -39,8 +39,35 @@ jobs:
     steps:
       - name: Checking out
         uses: actions/checkout@v3
+
+      - name: Generate HPKE Signing Key
+        id: hpke_signing_key
```
cjpatton, line 44: For consistency with `DAP_SIGNING_KEY`. If you prefer HPKE signing key, that's fine, let's just be consistent.

Here and below.
```suggestion
      - name: Generate DAP Signing Key
        id: dap_signing_key
```

.github/workflows/daphneci.yml
```diff
@@ -39,8 +39,35 @@ jobs:
...
+            echo "EEEEEOF"
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Generate HPKE Signing Certificate
+        id: hpke_signing_cert
+        run: |
+          echo "${{ steps.hpke_signing_key.outputs.hpke_signing_key }}" > private-key
+          cert=$(openssl req -key private-key \
+                      -new -x509 -days 365 -out - \
+                      -subj '/C=PT/L=Braga/O=Cloudflare Lda/CN=cloudflare.com')
```
cjpatton, line 59: What does "Lda" stand for?

daphne/src/taskprov.rs
```diff
@@ -227,7 +227,7 @@ impl VdafConfig {
         var: VdafTypeVar,
     ) -> Result<Self, DapAbort> {
         match (version, var) {
-            (.., VdafTypeVar::Prio2 { dimension }) => Ok(VdafConfig::Prio2 {
+            (_, VdafTypeVar::Prio2 { dimension }) => Ok(VdafConfig::Prio2 {
```
cjpatton, line 230: Is this clippy's preference or yours? Or, is this just matching the internal repo?

daphne_server/docker-compose-e2e.yaml
```diff
@@ -46,6 +48,9 @@ services:
       target: helper
     depends_on:
       - helper_storage
+    environment:
+      - DAP_SERVICE__SIGNING_KEY=${DAP_SIGNING_KEY}
```
cjpatton, line 52: Why use a different variable name here? Why not just go with `DAP_SIGNING_KEY`? Also, what's with the double underscore?

daphne_server/src/router/aggregator.rs
```diff
@@ -107,4 +148,81 @@ mod test {
...
+
+        const PAYLOAD: &[u8] = b"dummy HPKE configuration";
+        let signing_key = SigningKey::from({
+            let mut buf = String::new();
+            signing_key_file.read_to_string(&mut buf).unwrap();
+            p256::SecretKey::from_sec1_pem(&buf).unwrap()
+        });
+
+        let resp = DapResponse {
+            version: DapVersion::Draft02,
```
cjpatton, line 208: Just to give us a little less work when it's time to rip out draft 02.
```suggestion
            version: DapVersion::Draft09,
```

daphne_server/src/router/aggregator.rs
```diff
@@ -107,4 +148,81 @@ mod test {
...
+        let mut signing_key_file = tempfile::NamedTempFile::new().unwrap();
+        let status = Command::new("openssl")
+            .args([
+                os("ecparam"),
+                os("-name"),
+                os("prime256v1"),
+                os("-genkey"),
+                os("-noout"),
+                os("-out"),
+                os(signing_key_file.path()),
```
cjpatton, line 168: Instead of writing to a temporary file, you could write to stdout and capture it here.

daphne_server/src/router/aggregator.rs
```diff
@@ -107,4 +148,81 @@ mod test {
...
+
+    // Check that a signature produced by Daphne-Worker will be verified properly by the Clients.
+    #[test]
+    fn rondtrip_sign_hpke_config() {
+        fn os<S: AsRef<OsStr> + ?Sized>(s: &S) -> &OsStr {
+            OsStr::new(s)
+        }
+
+        let mut signing_key_file = tempfile::NamedTempFile::new().unwrap();
+        let status = Command::new("openssl")
```
cjpatton, line 160: It's not ideal to add "openssl" as a dependency for running `cargo test`.

I would probably just generate a test cert and signing key and stick them in this file. Note that It's not uncommon for unit tests to have certificates. It's a bummer that you have to update the test certs once in a while, but you could make the lifetime like 10 years or something.

An alternative to calling openssl from the shell is to use the rust bindings: https://github.com/sfackler/rust-openssl/blob/master/openssl/examples/mk_certs.rs

If you prefer to call openssl from the shell, let's make sure to document somewhwere that `openssl` is required to run tests.

daphne/src/taskprov.rs
mendess: My preference and to align with the rest of the match and the internal repo

daphne_server/docker-compose-e2e.yaml
mendess: This is a consequence of the [`config`](https://docs.rs/config/latest/config/struct.Environment.html) crate.

`DAP_` is the [env prefix](https://github.com/cloudflare/daphne/blob/6e1d98b70244ba14d6bf8eedaf384a0644651eef/daphne_server/examples/service.rs#L39)
`__` is [equivalent to the `.` operator in rust](https://github.com/cloudflare/daphne/blob/6e1d98b70244ba14d6bf8eedaf384a0644651eef/daphne_server/examples/service.rs#L41)

Hence, `DAP_SERVICE__SIGNING_KEY` maps to `config.`[`service`](https://github.com/cloudflare/daphne/blob/mendess/upstream-hpke-config-signing/daphne_server/examples/service.rs#L17)`.`[`signing_key`](https://github.com/cloudflare/daphne/blob/mendess/upstream-hpke-config-signing/daphne_service_utils/src/config.rs#L96)

daphne_server/src/router/aggregator.rs
mendess: No point in doing that since I need the key file for the next command anyway

.github/workflows/daphneci.yml
mendess: changed all `DAP_SIGNING_KEY` to `HPKE_SIGNING_KEY`

cjpatton:
Rebased.

.github/workflows/daphneci.yml
cjpatton, line 59: CN=cloudflare.com is probably not a great DN to use. In practice, this would be a test Leader, correct?

cjpatton (approved):

PR #523, opened 2024-03-05
Simplify `DapReportInitializer`

cjpatton:
Long ago, an implementation of this trait was responsible for some of the early report validation checks, including checking for replays and batch overlaps. These days it is used solely for initializing preparation for a set of reports.

However, `MockAggregator` currently does a bit more, in particular what `DapAggregator::try_put_agg_share_span()` already does. After removing this code, we can simplify the `DapReportInitializer` API so that it does only what it's supposed to do.

### DIFF
--- a/daphne/src/roles/aggregator.rs
+++ b/daphne/src/roles/aggregator.rs
@@ -11,9 +11,7 @@ use crate::{
     constants::DapMediaType,
     error::DapAbort,
     hpke::{HpkeConfig, HpkeDecrypter},
-    messages::{
-        BatchId, BatchSelector, HpkeConfigList, PartialBatchSelector, ReportId, TaskId, Time,
-    },
+    messages::{BatchId, BatchSelector, HpkeConfigList, ReportId, TaskId, Time},
     metrics::{DaphneMetrics, DaphneRequestType},
     protocol::aggregator::{EarlyReportStateConsumed, EarlyReportStateInitialized},
     DapAggregateShare, DapAggregateSpan, DapAggregationParam, DapError, DapGlobalConfig,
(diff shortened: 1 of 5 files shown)


daphne/src/roles/mod.rs
```diff
@@ -1222,27 +1228,21 @@ mod test {
...
-            &helper::handle_agg_job_req(&*t.helper, &req)
-                .await
-                .unwrap()
-                .payload,
-        )
-        .unwrap();
-        let transition = &agg_job_resp.transitions[0];
+        let query = task_config.query_for_current_batch_window(t.now);
+        let req = t.gen_test_coll_job_req(query, task_id).await;
+        leader::handle_coll_job_req(&*t.leader, &req).await.unwrap();
```
cjpatton, line 1233: We need to change this test because, in draft02, the report now gets rejected on the second request for the job rather than the first. This behavior is expected.

daphne/src/roles/aggregator.rs
```diff
@@ -26,12 +24,10 @@ use crate::{
 pub trait DapReportInitializer {
     /// Initialize a sequence of reports that are in the "consumed" state by initializing VDAF
     /// preparation.
-    async fn initialize_reports<'req>(
+    async fn initialize_reports(
         &self,
         is_leader: bool,
```
mendess, line 29: I want to eventually get rid of these `is_leader: bool` parameters. I know they closely model the draft but I think we should use `DapRole` enum instead

mendess (approved):
great PR 
image

daphne/src/roles/aggregator.rs
cjpatton: SGTM, but let's save for a future PR.

PR #524, opened 2024-03-06
Improvements to test code

cjpatton:
Stacked on #523 

In preparation for supporting heavy hitters, make some changes to `daphne::testing::MockAggregator`:
1. Make aggregate storage look more like `daphne_server::App`
2. Rename to `InMemoryAggregator`

### DIFF
--- a/daphne/src/lib.rs
+++ b/daphne/src/lib.rs
@@ -269,6 +269,15 @@ pub struct DapAggregateSpan<T> {
     span: HashMap<DapBatchBucket, (T, Vec<(ReportId, Time)>)>,
 }
 
+impl std::fmt::Display for DapBatchBucket {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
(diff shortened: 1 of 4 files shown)


daphne_service_utils/src/durable_requests/mod.rs
```diff
@@ -323,4 +323,23 @@ mod test {
...
+                DapBatchBucket::FixedSize {
+                    batch_id: BatchId([17; 32])
+                }
+            )
+        );
+        assert_eq!(
+            "window/1337",
+            format!("{}", DapBatchBucket::TimeInterval { batch_window: 1337 })
+        );
+    }
```
mendess, line 344: Why put this test in this file instead of the `bindings.rs` file? Since the code that actually uses this display implementation is there

mendess (approved):

daphne_service_utils/src/durable_requests/mod.rs
cjpatton: Good idea, moved.

PR #527, opened 2024-03-08
Add acceptance and load tests

mendess:
With this PR `dapf` now also doubles as a library to store various utility functions that can be used to interact with DAP implementations.

If you find any function that could be there that's used in e2e or acceptance tests please point it out so we can reduce duplicate code.


daphne_service_utils/dapf/src/lib.rs
```diff
@@ -0,0 +1,77 @@
+use std::{io::Cursor, path::Path};
```
cjpatton, line 1: copyright

daphne_service_utils/dapf/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+    DapVersion,
+};
+use daphne_service_utils::http_headers;
+use prio::codec::Decode;
+use reqwest::Client;
+use url::Url;
+use webpki::{EndEntityCert, ECDSA_P256_SHA256};
+use x509_parser::pem::Pem;
+
+pub trait HttpClientExt {
```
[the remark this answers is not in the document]
mendess: Ext is the idiomatic way to define an extension trait.

This trait is only used to add new methods to an external type (`reqwest::Client`). It's not supposed to be implemented for more types (hence the `Ext` suffix)

tbh I want to be a little bit fancy and add methods to HttpClient so that I can do stuff like [this](https://github.com/cloudflare/daphne/blob/12daef8e518a812f288552c968957ccf4b607236/daphne_service_utils/dapf/src/main.rs#L143-L144)

daphne_service_utils/dapf/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+    fn get_hpke_config(
+        &self,
+        base_url: &Url,
+        certificate_file: Option<&Path>,
+    ) -> impl std::future::Future<Output = anyhow::Result<HpkeConfig>> + Send;
```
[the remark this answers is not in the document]
cjpatton: If you could use aysnc_trait here, then I'd suggest doing so just for consistency with the rest of the repo. But it's op to you.

daphne_worker_test/src/acceptance.rs
```diff
@@ -0,0 +1,993 @@
+// Copyright (c) 2023 Cloudflare, Inc. All rights reserved.
+
+//! Acceptance tests for live Daphne deployments. These tests assume the following
+//! environment variables are defined:
+//!
+//! * `$HELPER_URL` - The base URL for the Helper.
+//! * `$LEADER_BEARER_TOKEN` - The bearer token (a string)
+//! * `$VDAF_VERIFY_INIT` - The hex encoded VDAF verification key initializer, as specified in the
+//!   Task prov extension.
```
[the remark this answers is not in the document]
mendess: This variable was never here :eye:

mendess:
@cjpatton addressed all comments. Moved the code into dapf. Let me know what you think. I think it could probably still be cleaned up a bit but I don't have time right now.

daphne_service_utils/dapf/src/test_durations.rs
cjpatton, line 1: copyright

daphne_service_utils/dapf/src/main.rs
```diff
@@ -431,6 +452,50 @@ async fn main() -> Result<()> {
...
+
+            tracing::info!("using vdaf: {:?}", t.vdaf_config);
+
+            let res = t
+                .test_helper(
+                    &TestOptions {
+                        bearer_token: t.leader_bearer_token.clone(),
+                        ..Default::default()
+                    },
+                    if t.helper_url.path().contains("/v02") {
```
cjpatton, line 466: There is already a function for resolving the `DapVersion` for a given path. I'm not sure off hand what it's called.

daphne_service_utils/dapf/src/main.rs
```diff
@@ -431,6 +452,50 @@ async fn main() -> Result<()> {
             println!("ID of the new HPKE config: {hpke_config_id}");
             Ok(())
         }
+        Action::Aggregate {} => {
+            let t = dapf::acceptance::Test::from_env()?;
```
cjpatton, line 456: A mixture of environment variables and command line arguments is confusing from a UX perspective. Let's consider which of these we want to be environment variables and which we don't.

daphne_service_utils/dapf/src/main.rs
```diff
@@ -431,6 +452,50 @@ async fn main() -> Result<()> {
             println!("ID of the new HPKE config: {hpke_config_id}");
             Ok(())
         }
+        Action::Aggregate {} => {
```
cjpatton, line 455: In the documentation we should say that this uses taskprov.

cjpatton (approved):

PR #528, opened 2024-03-10
Various improvements for the `protocol` module

cjpatton:
See commit messages.

### DIFF
--- a/daphne/src/protocol/aggregator.rs
+++ b/daphne/src/protocol/aggregator.rs
@@ -89,8 +89,10 @@ pub trait EarlyReportState {
 pub struct ReportState {
     pub metadata: ReportMetadata,
     pub public_share: Vec<u8>,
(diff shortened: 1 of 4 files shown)


daphne/src/protocol/aggregator.rs
```diff
@@ -1,6 +1,3 @@
-// Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
-// SPDX-License-Identifier: BSD-3-Clause
-
```
mendess, line 3: Why remove copywrite?

mendess (approved):

daphne/src/protocol/aggregator.rs
cjpatton: Whoopsie daisy, no idea how that happened. Will restore.

PR #533, opened 2024-03-12
Make argument parsing in dapf more consistent.

mendess:
Also make use of the `env` atribute to make the CLI easier to script and
use. All options annotated with `env` can be set in the environment
instead of being passed as command line arguments.

And finally removed the option to pass a bearer token through the
commmand line to avoid accidentally leaking secrets to a user's bash
history.


cjpatton (approved):
Very good idea!

PR #534, opened 2024-03-12
Align report state naming with VDAF draft

cjpatton:


daphne/src/protocol/aggregator.rs
```diff
@@ -277,9 +277,12 @@ pub enum EarlyReportStateInitialized {
     Ready {
         metadata: ReportMetadata,
         public_share: Vec<u8>,
-        leader_prep_share: Option<Vec<u8>>,
-        vdaf_state: VdafPrepState,
-        message: VdafPrepMessage,
+        // Set by the Helper.
+        //
+        // draft02 compatibility: This is only set set in the latest draft.
```
mendess, line 282:
```suggestion
        // draft02 compatibility: This is only set in the latest draft.
```

mendess (approved):

PR #538, opened 2024-03-14
I forgot to actually wait for the service in the test_runner

mendess:
This should have been done in #536, but I forgot....


cjpatton (changes requested):
The commit message needs work.

PR #539, opened 2024-03-14
Fix bugs in draft-dcook-ppm-dap-interop-test-design

cjpatton:
Running our Helper against Janus' Leader revealed a few minor bugs in our implementation of the interop API:

- POST /internal/test/ready instead of GET
- POST /internal/test/add_task was returning 500s

Also, add a method for Janus's test runner to conveniently configure an HPKE config: POST /internal/test/add_hpke_config

### DIFF
--- a/daphne_server/src/router/test_routes.rs
+++ b/daphne_server/src/router/test_routes.rs
@@ -46,7 +46,7 @@ where
 
     router
         .route("/internal/delete_all", post(delete_all))
-        .route("/internal/test/ready", get(check_storage_readyness))
+        .route("/internal/test/ready", post(check_storage_readyness))
         .route(
             "/internal/test/endpoint_for_task",
             post(endpoint_for_task_default),
@@ -55,9 +55,12 @@ where
             "/:version/internal/test/endpoint_for_task",
             post(endpoint_for_task),
         )
-        // TODO: could be removed after we add the divviup api
-        .route("/internal/test/add_task", post(add_task))
+        .route("/internal/test/add_task", post(add_task_default))
         .route("/:version/internal/test/add_task", post(add_task))
+        .route(
+            "/internal/test/add_hpke_config",
+            post(add_hpke_config_default),
+        )
         .route(
             "/:version/internal/test/add_hpke_config",
             post(add_hpke_config),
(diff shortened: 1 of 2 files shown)


daphne_server/src/router/test_routes.rs
```diff
@@ -46,7 +46,7 @@ where
 
     router
         .route("/internal/delete_all", post(delete_all))
-        .route("/internal/test/ready", get(check_storage_readyness))
+        .route("/internal/test/ready", post(check_storage_readyness))
```
mendess, line 49: Why must this be a POST?

daphne_server/src/router/test_routes.rs
```diff
@@ -145,6 +148,15 @@ async fn add_task(
...
 
+#[tracing::instrument(skip(app, cmd))]
+async fn add_task_default(
+    State(app): State<Arc<App>>,
+    Json(cmd): Json<InternalTestAddTask>,
+) -> impl IntoResponse {
+    let version = app.service_config.default_version;
+    add_task(State(app), Path(version), Json(cmd)).await
+}
+
```
mendess, line 159: nit: No need to destructure the Json extractor if you're just going to pass it along

 ```rust
async fn add_task_default(
    State(app): State<Arc<App>>,
    cmd: Json<InternalTestAddTask>,
) -> impl IntoResponse {
    let version = app.service_config.default_version;
    add_task(State(app), Path(version), cmd).await
}
```

File: daphne_server/src/router/test_routes.rs [hunk 3]
```diff
@@ -160,3 +172,12 @@ async fn add_hpke_config(
...
     }
 }
+
+#[tracing::instrument(skip(app, hpke))]
+async fn add_hpke_config_default(
+    State(app): State<Arc<App>>,
+    Json(hpke): Json<HpkeReceiverConfig>,
+) -> impl IntoResponse {
+    let version = app.service_config.default_version;
+    add_hpke_config(State(app), Path(version), Json(hpke)).await
```
mendess, line 182: same here

daphne_server/src/router/test_routes.rs
cjpatton: https://datatracker.ietf.org/doc/html/draft-dcook-ppm-dap-interop-test-design#section-4.3.1

daphne_server/src/router/test_routes.rs
cjpatton: Done.

mendess (approved):

PR #541, opened 2024-03-15
Remove draft-ietf-ppm-dap-02

### LINKED ISSUE #537: Remove draft-ietf-ppm-dap-02

cjpatton:
Partially addresses #537.

Drop support for DAP draft 02 and prune some code that is no longer
needed:

1. Remove `AggregationJobContinueReq`, as this is only needed to support
   multi-round VDAFs. For now we only support 1-round VDAFs.

2. Remove code for draft-wang-ppm-dap-taskprov-02, including processing
   of `TaskConfig` struct carried by report extensions.

2. Remove `MetaAggregationJobId`. This was used to unify shared
   semantics across draft 02 and 09, which have a different syntax for
   aggregation job IDs.

3. Remove the code in draft 02 for generating a collection job ID.

As a result of dropping draft 02, it is possible to further simplify
control flow and some protocol semantics (e.g., returning multiple HPKE
configs). To keep the commit as small as possible, we leave most of
these changes as TODOs.

### DIFF
--- a/daphne/src/constants.rs
+++ b/daphne/src/constants.rs
@@ -6,14 +6,6 @@
 use crate::{DapSender, DapVersion};
 
 // Media types for HTTP requests.
-const DRAFT02_MEDIA_TYPE_AGG_CONT_REQ: &str = "application/dap-aggregate-continue-req";
-const DRAFT02_MEDIA_TYPE_AGG_CONT_RESP: &str = "application/dap-aggregate-continue-resp";
-const DRAFT02_MEDIA_TYPE_AGG_INIT_REQ: &str = "application/dap-aggregate-initialize-req";
-const DRAFT02_MEDIA_TYPE_AGG_INIT_RESP: &str = "application/dap-aggregate-initialize-resp";
-const DRAFT02_MEDIA_TYPE_AGG_SHARE_RESP: &str = "application/dap-aggregate-share-resp";
-const DRAFT02_MEDIA_TYPE_COLLECT_RESP: &str = "application/dap-collect-resp";
-const DRAFT02_MEDIA_TYPE_HPKE_CONFIG: &str = "application/dap-hpke-config";
-const MEDIA_TYPE_AGG_JOB_CONT_REQ: &str = "application/dap-aggregation-job-continue-req";
 const MEDIA_TYPE_AGG_JOB_INIT_REQ: &str = "application/dap-aggregation-job-init-req";
 const MEDIA_TYPE_AGG_JOB_RESP: &str = "application/dap-aggregation-job-resp";
 const MEDIA_TYPE_AGG_SHARE_REQ: &str = "application/dap-aggregate-share-req";
@@ -25,13 +17,10 @@ const MEDIA_TYPE_REPORT: &str = "application/dap-report";
 
 /// Media type for each DAP request. This is included in the "content-type" HTTP header.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[cfg_attr(test, derive(strum::EnumIter))]
 pub enum DapMediaType {
     AggregationJobInitReq,
     AggregationJobResp,
-    AggregationJobContinueReq,
-    /// draft02 compatibility: the latest draft doesn't define a separate media type for initialize
-    /// and continue responses, but draft02 does.
-    Draft02AggregateContinueResp,
     AggregateShareReq,
     AggregateShare,
     CollectReq,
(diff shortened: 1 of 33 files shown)


daphne/src/taskprov.rs
```diff
@@ -32,19 +31,15 @@ pub(crate) const TASKPROV_SALT: [u8; 32] = [
...
 
-fn compute_task_id_draft02(serialized: &[u8]) -> TaskId {
+/// Compute the task id of a serialized task config.
+pub fn compute_task_id(_version: DapVersion, serialized: &[u8]) -> TaskId {
     let d = digest::digest(&digest::SHA256, serialized);
     let dref = d.as_ref();
     let mut b: [u8; 32] = [0; 32];
     b[..32].copy_from_slice(&dref[..32]);
     TaskId(b)
 }
```
mendess, line 42: Remove the version parameter?

daphne/src/lib.rs
```diff
@@ -755,11 +737,9 @@ impl DapTaskConfig {
     /// Leader: Resolve taskprov advertisement to send in a request to the Helper.
     pub(crate) fn resolve_taskprove_advertisement(&self) -> Result<Option<String>, DapError> {
         if let DapTaskConfigMethod::Taskprov { info } = &self.method {
-            if info.is_none() && self.version != DapVersion::Draft02 {
+            if info.is_none() {
                 // The task config indicates that the configuration method was taskprov, but we
-                // don't have enough information to construct the advertisement. This is not a
-                // problem for draft02, however, because the task config was encoded by the report
-                // extensions in that version.
+                // don't have enough information to construct the advertisement.
```
mendess, line 742: Is there a situation in which `info` being `None` makes sense? if not maybe it's type shouldn't be `Option`

mendess (commented):
Why leave the DapVersion in the `ParameterizedDecode` implementations? Is it for future proofing?

daphne/src/constants.rs
```diff
@@ -46,186 +34,54 @@ impl DapMediaType {
...
-            | Self::AggregateShare => DapSender::Helper,
+            Self::AggregationJobResp | Self::AggregateShare => DapSender::Helper,
             Self::Report => DapSender::Client,
             Self::CollectReq => DapSender::Collector,
         }
     }
 
     /// Parse the media type from the content-type HTTP header.
-    pub fn from_str_for_version(version: DapVersion, content_type: &str) -> Option<Self> {
+    pub fn from_str_for_version(_version: DapVersion, content_type: &str) -> Option<Self> {
```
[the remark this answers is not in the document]
cjpatton: The version is useful here for the same reason noted elsewhere.

daphne/src/error/aborts.rs
```diff
@@ -149,15 +149,26 @@ impl DapAbort {
...
-                "unexpected content-type: got {got_str}; want {want_str}"
-            ))
-        } else {
-            Self::BadRequest(format!("missing content-type: expected {want_str}"))
-        }
+        let Some(want_content_type) = expected.as_str_for_version(req.version) else {
+            return Self::BadRequest(format!(
+                "unexpected content-type for DAP version {:?}",
+                req.version
+            ));
```
mendess, line 156: If this else branch happens, doesn't that mean that we expected a media type that's unsupported by the version? Shouldn't this be a 500 instead? If we expect a media type that this version could never have supported than this was a mistake by our code

mendess (changes requested):

cjpatton (commented):
> Why leave the DapVersion in the `ParameterizedDecode` implementations? Is it for future proofing

I've mostly kept this plumbing in for now. I'd like to see what happens in the next DAP version before we start removing it.

daphne/src/error/aborts.rs
cjpatton: Yes, 500 is more appropriate, however since the return type is `DapAbort`, we can't force this here. I also don't want to add another variant of `DapAbort`, since we recently moved from handling 500s via `DapError`.

Another option would be to just panic, as we're already doing. Would you prefer that?

daphne/src/lib.rs
mendess: Alright, defer this to an issue. I might pick it up

daphne/src/error/aborts.rs
mendess: I'd prefer a panic (which will be turned into a 500 anyway) than this. Panics are okay for code that should *never* happen. And we don't expect this to ever happen.

daphne/src/lib.rs
cjpatton: Done.

daphne/src/error/aborts.rs
```diff
@@ -149,22 +149,19 @@ impl DapAbort {
...
-                "missing or unexpected content type for DAP version {:?}",
-                req.version
-            ));
-        };
+            .unwrap_or_else(|| {
+                unreachable!(
+                    "missing or unexpected content type for DAP version {:?}",
+                    req.version
+                )
+            });
```
mendess, line 164: If this is `None` then the user requested a media type that this version doesn't support, right? This would mean the previous `BadRequest` code was correct

cjpatton: Good catch, done.

cjpatton:
Rebased.

mendess (approved):

cjpatton:
Squashed.

PR #543, opened 2024-03-18
Rework hpke config decryption

mendess:


daphne/src/testing.rs
```diff
@@ -703,7 +706,23 @@ impl HpkeDecrypter for InMemoryAggregator {
...
+                tokio::task::spawn_blocking({
+                    let info = info.to_vec();
+                    let aad = aad.to_vec();
+                    let ciphertext = ciphertext.clone();
+                    let config = hpke_receiver_config.clone();
+                    move || config.decrypt(&info, &aad, &ciphertext)
+                })
+                .await
+                .unwrap()
+            }
```
cjpatton, line 720: Why do this during testing?

cjpatton: Or more to the point: why change the behavior in testing?

cjpatton (approved):
LGTM, great ideas!

PR #546, opened 2024-03-19
Optimize load tests by streaming reports

mendess:
- Improve usablity of dapf around vdaf configs
- Generate reports dinamically in load tests

### DIFF
--- a/daphne_service_utils/dapf/src/acceptance/mod.rs
+++ b/daphne_service_utils/dapf/src/acceptance/mod.rs
@@ -17,6 +17,7 @@
 //!
 
 pub mod load_testing;
+pub mod report_generator;
 
 use crate::{test_durations::TestDurations, HttpClientExt};
 use anyhow::{anyhow, bail, Context, Result};
@@ -27,7 +28,7 @@ use daphne::{
     error::aborts::ProblemDetails,
     hpke::{HpkeConfig, HpkeKemId, HpkeReceiverConfig},
     messages::{
-        self, AggregateShareReq, AggregationJobId, AggregationJobResp, Base64Encode, BatchId,
+        AggregateShareReq, AggregationJobId, AggregationJobResp, Base64Encode, BatchId,
         BatchSelector, PartialBatchSelector, ReportId, TaskId,
     },
     metrics::{prometheus::DaphnePromMetrics, DaphneMetrics},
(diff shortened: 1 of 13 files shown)


daphne_service_utils/dapf/src/acceptance/mod.rs
```diff
@@ -878,3 +832,99 @@ pub fn now() -> Now {
...
+/// An iterator that yields the size of chunks based on a length and a maximum chunk size.
+///
+/// # Example
+/// ```
+/// use dapf::acceptance::Chunker;
+///
+/// assert_eq!(Chunker::of_size(200, 51).collect::<Vec<_>>(), [51, 51, 51, 47]);
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct Chunker {
```
Noah-Kennedy, line 845: Where is this used?

daphne_service_utils/dapf/src/acceptance/report_generator.rs
```diff
@@ -0,0 +1,154 @@
...
+impl ReportGenerator {
+    pub(super) fn new(
+        test_task_config: &TestTaskConfig,
+        reports_per_batch: usize,
+        measurement: &DapMeasurement,
+        version: DapVersion,
+        now: Now,
+    ) -> Self {
+        let (tx, rx) = mpsc::channel(4);
+        thread::spawn({
```
Noah-Kennedy, line 67: Why not just use `rayon::spawn` on each job?

daphne_service_utils/dapf/src/acceptance/mod.rs
mendess: It's used in this same file to produce an iterator to split the number of report among jobs. Tbh there's probably a shorter way to write this and I'm just suffering from acute skill issue :rofl:

Noah-Kennedy: There probably is, but never let the perfect be the enemy of the good.

The jank will do just fine.

cjpatton (approved):
Changes to the `daphne` crate look good to me. As for the changes to `dapf`, consider my approval a stamp to unblock you. At a high level it seems like a reasonable way to speed up report generation. The changes to the CLI are a little clunky but I don't have any better ideas.

I would recommend adding some code comments to guide the reader. I'm thinking in particular of the `ReportGenerator` module: what we're doing here is simulating a bunch of Clients generating reports, which is just really really expensive to do on a single machine. In production, this will be spread across actual Client devices and we expect the cost per-device to be low.

PR #548, opened 2024-03-19
testing: Add aggregation parameter validation logic

cjpatton:
Each VDAF specifies constraints on the aggregation parameter used to aggregate reports:

1. For IDPF-based VDAFs like Mastic, the sequence of aggregation parameters must determine a valid tree traversal.

2. For simple VDAFs like Prio3, only the empty aggregation parameter is valid.

This commit adds code to `InMemoryAggregator` for enforcing aggregation parameter validity.

Note that this logic will eventually need to be implemented by `daphne_server::App`.

### DIFF
--- a/daphne/src/testing.rs
+++ b/daphne/src/testing.rs
@@ -453,37 +453,104 @@ impl AuditLog for MockAuditLog {
     }
 }
 
-#[derive(Default)]
+/// Aggregate share and associated book-keeping data for a bucket of reports.
+//
+// TODO heavy hitters: Move this code into a space where `daphne_server` can use it. We need the
+// same logic there.
 #[cfg_attr(any(test, feature = "test-utils"), derive(deepsize::DeepSizeOf))]
-pub(crate) struct AggregateStoreForCollection {
-    /// The aggregate share.
-    pub(crate) agg_share: DapAggregateShare,
-    /// Indication of whether the aggregate share has been collected.
-    pub(crate) collected: bool,
-    /// The reports included in the aggregate share.
-    pub(crate) reports: HashSet<ReportId>,
+pub struct AggregateStore {
+    pub agg_share: DapAggregateShare,
+
+    /// Indication of whether the bucket has been collected at least once. If set, new reports for
+    /// this bucket will be rejected.
+    pub collected: bool,
+
+    /// The reports included in the current aggregate share. If a report wants to be aggregated is
+    /// already in this set, it will be rejected.
+    pub reports: HashSet<ReportId>,
+
+    /// The current aggregation parameter.
+    pub(crate) agg_param: DapAggregationParam,
+}
+
+impl AggregateStore {
+    /// Check if the given aggregation parameter can be used to aggregate the reports in this bucket.
+    ///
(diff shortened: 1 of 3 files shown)


daphne/src/testing.rs
```diff
@@ -467,37 +467,104 @@ impl AuditLog for MockAuditLog {
...
         task_id: &TaskId,
         bucket: &DapBatchBucket,
         agg_param: &DapAggregationParam,
-    ) -> &mut AggregateStoreForCollection {
-        let agg_level = agg_param.level();
-        // NOTE(cjpatton) `daphne_server::App` will use a similar naming scheme for instances of
-        // the `AggregateStore` DO. However, for backwards compatibility, if `agg_level == 0`, then
-        // we won't include it in the DO name.
-        self.0
-            .entry(format!("{task_id}/{bucket}/{agg_level}"))
```
cjpatton, line 499: We can't split the buckets by level anymore because we need to know the last agg param to determine if the next one is valid. In fact, we only need one `AggregateStore` for each bucket: it will store the agg share for the current agg param.

daphne/src/testing.rs
```diff
@@ -467,37 +467,104 @@ impl AuditLog for MockAuditLog {
...
+    /// Check if the given aggregation parameter can be used to aggregate the reports in this bucket.
+    ///
+    /// For IDPF-based VDAFs (e.g., Mastic), this returns true only if the aggregation parameter
+    /// results in a valid tree traversal, that is, each prefix encoded by the aggregation
+    /// parameter is equal to or a child of a prefix in the current aggregation parameter.
+    ///
+    /// For simple VDAFs (e.g., Prio3), only the empty aggregation parameter is valid.
+    //
+    // TODO heavy hitters: Enforce the maximum tree depth.
+    pub fn is_valid(&self, next_agg_param: &DapAggregationParam) -> bool {
```
mendess, line 500: Is there any reason for this method to be pub? I don't like methods that return `bool` in public APIs as they tend to be misused and distract from the most appropriate methods (in this case goto).

Another point is that maybe this function should instead return `Result<(), DapAbort>` so that `goto` can return `Result<&mut Self, DapAbort>` so we don't lose the information of why you couldn't progress

mendess (changes requested):

daphne/src/testing.rs
cjpatton: No reason to be public, removed.

Propagating an error might be useful.  A `DapAbort` isn't appropriate because the caller wouldn't want to handle this generically as an abort. The correct action depends on the context. For instance, if this error occurs while processing an aggregation job, then the reports for this bucket should be rejected, rather than abort the aggregation job altogether.

Also, in the rejection case, we won't inform the Leader why the aggregation parameter is bad, just that we rejected the report because of it. (This is an indication of a misbehaving Leader.)

cjpatton:
TODO: Split logs into files: https://ietf.slack.com/archives/C02P9620XT6/p1710868689386369?thread_ts=1710868077.859259&cid=C02P9620XT6

mendess (approved):

PR #550, opened 2024-03-20
Cleanup http header related code

mendess:
- Move dap-auth-token and dap-taskprov headers to constants
- Use standard bearer token api in reqwest instead of manually implementing it


cjpatton (approved):
👍

PR #551, opened 2024-03-20
Add metrics to storage proxy

mendess:


daphne_worker/src/storage_proxy/metrics.rs
```diff
@@ -0,0 +1,109 @@
...
+                        tracing::error!("failed to configure metrics push: missing server URL");
+                        None
+                    }
+                }
+            })
+            .as_ref()
+    }
+}
+
+pub struct Metrics {
```
cjpatton, line 57: For consistency with the `daphne_server` crate? 
```suggestion
pub struct StorageProxyMetrics {
```

daphne_worker/src/storage_proxy/metrics.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+impl Metrics {
+    pub fn new(env: &Env) -> Option<Self> {
+        let config = Config::from_env(env)?;
+
+        let durable_request_retry_count = register_int_counter_vec!(
+            "durable_request_retry_count",
+            "The number of times a request to durable objects was retried before succeeding or giving up",
+            &["count"],
```
cjpatton, line 72: It might make sense to label this counter by the DO name and request path.

cjpatton (approved):
🦆, with one suggestion.

daphne_worker/src/storage_proxy/metrics.rs
mendess: it's already inside of the `storage_proxy` module, maybe it doesn't need that qualification

cjpatton: Up to you, just a suggestion.

daphne_worker/src/storage_proxy/metrics.rs
```diff
@@ -0,0 +1,148 @@
...
+        let registry = Registry::new_custom(
+            Option::None,
+            Option::Some(HashMap::from([("host".to_string(), host.to_string())])),
+        )
+        .unwrap();
+
+        let durable_request_retry_count = register_int_counter_vec_with_registry!(
+            "durable_request_retry_count",
+            "The number of times a request to durable objects was retried before succeeding or giving up",
+            &["count", "object", "path"],
```
cjpatton, line 85: Actually, what variants of the "count" label will there be? Don't we just want a counter for each "object"/"path" pair?

mendess: Updated the description of the metric to hopefully make it more obvious.

`count` it will be a number in the range [-1, 4].
- `-1`: the request failed
- `0`: no retries were done
- `1`..`4`: X retires were done but we succeeded

cjpatton: Ahhhh, it keepts track of the number of times it was retired. That works!

cjpatton (approved):

PR #554, opened 2024-03-22
Refactor durable object definitions

mendess:
- Rework durable object definitions
- Rename DapDurableObject and GarbageCollector


daphne_worker/src/durable/helper_state_store.rs
```diff
@@ -103,26 +71,24 @@ impl HelperStateStore {
...
-    #[inline(always)]
-    fn deployment(&self) -> DaphneWorkerDeployment {
-        self.config.deployment
-    }
-}
+        let duration = SELF_DELETE_AFTER.get_or_init(|| {
+            Duration::from_secs(
+                self.env
+                    .var(VAR_NAME)
+                    .unwrap_or_else(|e| panic!("{VAR_NAME} not defined: {e}"))
```
mendess, line 83: Should we have a default value here like I did for `AggregateStore`? I kept the old behavior but it might not be desirable.

cjpatton: Setting a default sounds good to me.

daphne_worker/src/durable/mod.rs
```diff
@@ -209,97 +190,185 @@ impl<'srv> DurableConnector<'srv> {
...
+    /// Handle a durable object request.
+    async fn handle(&mut self, req: Request) -> Result<Response>;
 
-    fn deployment(&self) -> DaphneWorkerDeployment;
+    /// When this durable object should self cleanup.
+    ///
+    /// The default implementation of this function returns None, signaling that this object should
+    /// not be automatically cleaned up.
+    fn should_cleanup_at(&self) -> Option<ScheduledTime> {
+        None
```
cjpatton, line 208: Consider not providing a default implementation. That way we force ourselves to think this through for each DO.

daphne_worker/src/durable/mod.rs
```diff
@@ -209,97 +190,185 @@ impl<'srv> DurableConnector<'srv> {
...
-            Some(GarbageCollector::DeleteAll) => {
-                self.state().storage().delete_all().await?;
-                *self.touched() = false;
-                return Ok(std::ops::ControlFlow::Break(()));
+        #[worker::durable_object]
+        impl DurableObject for $name {
+            fn new(state: ::worker::State, env: ::worker::Env) -> Self {
+                $crate::tracing_utils::initialize_tracing(&env);
+                <Self as $crate::durable::GcDurableObject>::new(state, env)
             }
```
cjpatton, line 254: It's odd to shadow `new()` here. Can we use a different function name for the `GcDurableObject` constructor?

daphne_worker/src/durable/mod.rs
```diff
@@ -209,97 +190,185 @@ impl<'srv> DurableConnector<'srv> {
...
+                use ::tracing::Instrument;
+
+                // Ensure this DO instance is garbage collected eventually.
+                if let Some(lifetime) = self.should_cleanup_at() {
+                    ensure_alarmed(&self.state, lifetime, deployment(&self.env)).await?;
+                };
+
+                if matches!(deployment(&self.env), Dev) {
+                    // Try to handle a delete all request.
+                    req = match setup_and_handle_garbage_collector_requests::<Self>(
```
cjpatton, line 275: Should we change "garbage collector" to "test cleaner" here?

daphne_worker/src/durable/mod.rs
```diff
@@ -209,97 +190,185 @@ impl<'srv> DurableConnector<'srv> {
...
+            // the experimental-local dev environment doesn't have
+            // working get_alarm() and set_alarm() yet, so we want to
+            // ignore errors in that case.
+            return Err(e);
+        }
+    }
+    Ok(())
+}
+
+/// Register a do to be deleted by the [`GarbageCollector`](garbage_collector::GarbageCollector).
```
cjpatton, line 323:
```suggestion
/// Register a DO to be deleted by the [`GarbageCollector`](garbage_collector::GarbageCollector).
```

cjpatton, line 323: Need to "GarbageCollector". Give this comment another pass.

daphne_worker/src/durable/mod.rs
```diff
@@ -209,97 +190,185 @@ impl<'srv> DurableConnector<'srv> {
...
+    // the else branch was already done before.
+    if let Some(TestStateCleaner::DeleteAll) = TestStateCleaner::try_from_uri(&req.path()) {
+        state.storage().delete_all().await?;
+        state
+            .storage()
+            .put("gc_delete_all_is_setup", &false)
+            .await?;
+        Ok(ControlFlow::Break(()))
+    } else {
+        // if this wasn't a delete all request then we must ensure that we have registered
```
cjpatton, line 351: Don't we only want to do this in testing?
```suggestion
        // If this wasn't a delete all request then we must ensure that we have registered
```

cjpatton (commented):
I'm about half way through. A couple of high level things:

1. Consider putting the `TestStateCleaner` DO behind "test-utils".
2. Can you clarify if any functional changes are indended in this PR? In particular, are we changing how garbage colleciton will work in deployment?

daphne_worker/src/durable/mod.rs
mendess: We check that before calling the function at all, you can see that in the macro. In production we don't even call this.

daphne_worker/src/durable/mod.rs
```diff
@@ -2,305 +2,169 @@
...
+async fn ensure_alarmed(state: &State, env: &Env, lifetime: ScheduledTime) -> Result<()> {
+    if let Err(e) = state.storage().set_alarm(lifetime).await {
+        if matches!(deployment(env), DaphneWorkerDeployment::Dev) {
+            warn!("ignoring get_alarm() failure in a dev environment until --experimental-local implements it: {e}");
+        } else {
+            // We only return an error if not in the "dev" deployment as
+            // the experimental-local dev environment doesn't have
+            // working get_alarm() and set_alarm() yet, so we want to
+            // ignore errors in that case.
+            return Err(e);
```
cjpatton, line 162: Here we unconditionally set the alarm, whereas previously we only set the alarm we haven't already. This is intended to change GC semantics such that the DO instance is deleted some time after last touch, rather than on first touch as we did before. I think this is a great idea, but two points:
1. We need to document the change of semantics in the commit message. This is an important functional change.
2. Are there side-effects of calling `set_alarm()` multiple times? I'm wondering if (a) all calls after the first **are ignored** or (b) if calling `set_alarm()` on every DO request is inefficient. Is there any documentation that can help shed light?

cjpatton (commented):
Looks good, just want to make sure to nail down the details for the new garbage collection functionality.

daphne_worker/src/durable/mod.rs
mendess: I tested this implementation and set_alarm does clear the previous alarm before creating a new one. On the topic of efficiency I have no idea tbh but I can't imagine it's that expensive, the docs don't mention anything.

docs: https://developers.cloudflare.com/durable-objects/api/alarms/

The docs don't explicitly mention this "replace last alarm" behavior but `getAlarm` seems to suggest it and experimentation confirms it. Speaking of which, running this locally didn't trigger the error that the code seems to suggest it would :thinking: I think this branch can be removed.

daphne_worker/src/durable/test_state_cleaner.rs
```diff
@@ -0,0 +1,389 @@
...
+    /// Ties are broken by a random nonce tacked on to the key. The format of the ordinal is:
+    ///
+    /// ```text
+    ///     time/<time>/nonce/<nonce>
+    /// ```
+    ///
+    /// where <time> is the timestamp and <nonce> is a random nonce.
+    pub(crate) fn new_roughly_ordered(item: T, prefix: &str) -> Self {
+        let mut rng = thread_rng();
+        let time = Date::now().as_millis() / 1000;
```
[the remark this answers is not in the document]
cjpatton: Sad!

daphne_worker/src/durable/mod.rs
cjpatton: We're using an old version of miniflare for the end-to-end tests, not the workers runtime itself. The behavior might be different in deployment

I'm not too worried about efficiency right now -- let's just deploy and monitor. Load testing should point out any issues.

I think the only concern I have left is ensuring that setting the alarm multiple times has the expected effect. If documentation supports this great. Otherwise, I'd propose testing it out before merging.

cjpatton (approved):

PR #556, opened 2024-03-26
Small optimization o batch overlap check

mendess:
By iterating the stream instead of collecting to a vector we get the same amount
of concurrency but abort all in flight requests as soon as we get a 'collected'
response. And we avoid allocating a vector.


cjpatton (approved):
There is a typo in the commit message.

PR #558, opened 2024-03-27
Add metric to measure aggregate store merge request times

mendess:


daphne/src/vdaf/mod.rs
```diff
@@ -332,6 +332,20 @@ pub enum VdafAggregateShare {
     FieldPrio2(prio::vdaf::AggregateShare<FieldPrio2>),
 }
 
+impl VdafAggregateShare {
+    pub fn len(&self) -> usize {
+        match self {
+            Self::Field64(fields) => fields.as_ref().len(),
```
cjpatton, line 338: Might be a better name
```suggestion
            Self::Field64(field_vec) => fields.as_ref().len(),
```

cjpatton (approved):

PR #561, opened 2024-04-02
Remove unused draft 02 routes

mendess:


daphne_worker_test/tests/e2e/test_runner.rs
```diff
@@ -353,6 +353,7 @@ impl TestRunner {
         get_raw_hpke_config(client, self.task_id.as_ref(), &self.helper_url, "helper").await
     }
 
+    #[allow(dead_code)]
```
cjpatton, line 356: What code does cargo think is dead? Can we solve this by removing the "dead code"? In the past, cargo had decided that some code was dead which was actually needed to drive tests.

cjpatton (approved):
Suggestion for the commit title: I think the title can be read as modifying daphne_server, in particular removing HTTP routes from the app. I would add a "component" being modified to the commit title so that it's more clear what's being modified. Suggestion:

```
daphne_worker_test: Remove unused draft02 routes
```

daphne_worker_test/tests/e2e/test_runner.rs
mendess: This is actually dead, but since it might be useful to post to the leader in the future I kept it. The dead code is the entire function

mendess:
> Suggestion for the commit title: I think the title can be read as modifying daphne_server, in particular removing HTTP routes from the app. I would add a "component" being modified to the commit title so that it's more clear what's being modified. Suggestion:
> 
> ```
> daphne_worker_test: Remove unused draft02 routes
> ```

But the daphne_server is what's being modified :thinking:

cjpatton:
> > Suggestion for the commit title: I think the title can be read as modifying daphne_server, in particular removing HTTP routes from the app. I would add a "component" being modified to the commit title so that it's more clear what's being modified. Suggestion:
> > ```
> > daphne_worker_test: Remove unused draft02 routes
> > ```
> 
> But the daphne_server is what's being modified 🤔

You're right, I totally misread. That was sloppy of me, I apologize.

cjpatton (approved):

PR #562, opened 2024-04-03
Move unwraps to tests instead of test_runner

mendess:
This way, when the panic happens it shows the line of the test that
paniced instead of the line of test runner that paniced. Hopefuly
improving the developer experience of finding out why e2e tests failed.


cjpatton (approved):
Good idea 👍

PR #565, opened 2024-04-05
Feature gate daphne_service_utils dependency on capnproto

mendess:
This lets dapf not depend on capnproto which helps keep CI builds simpler.

### DIFF
--- a/daphne_service_utils/Cargo.toml
+++ b/daphne_service_utils/Cargo.toml
@@ -17,7 +17,7 @@ readme = "../README.md"
 
 [dependencies]
 async-trait.workspace = true
-capnp.workspace = true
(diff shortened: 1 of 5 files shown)


daphne_service_utils/Cargo.toml
```diff
@@ -42,7 +42,7 @@ capnpc = "0.18.1"
 [features]
 test-utils = ["dep:prometheus", "daphne/prometheus", "daphne/test-utils"]
 prometheus = ["dep:prometheus", "daphne/prometheus"]
-test_acceptance = ["dep:prometheus"]
```
mendess, line 45: the `test_acceptance` feature was unused so I deleted it

PR #571, opened 2024-04-09
Improve visibilty over report rejections

mendess:


daphne/src/protocol/aggregator.rs
```diff
@@ -206,11 +206,12 @@ impl EarlyReportStateConsumed {
             .transpose()
         {
             Ok(peer_prep_share) => peer_prep_share.map(|bytes| bytes.to_vec()),
-            Err(_) => {
+            Err(e) => {
+                tracing::warn!(error = ?e, "rejecting report");
```
cjpatton, line 210: What are information are we getting with this change that we don't get from metrics? I ask because `warn!` would currently be quite noisy in logs (see the reports/s panel of the dashboard). 

Here and below.

mendess: we get the agg job id and the task id where the reports have been rejected (something we've been asked for) and we get the actual failure instead of just `VdafPrepError` which is what the leader will see. I'd like to know more

cjpatton: If we reject 200 reports in a job, then that's 200 log lines. From experience, it's quite likely that if we're going to reject 1 report in a job, then we're also going to reject a lot more. How about this instead: Just before completing the aggregation job, check if there any rejected reports. If there are, then log the aggregation job ID, the task ID, and how many times each rejection occurred.

cjpatton: Hmm, OK.

cjpatton (approved):

PR #572, opened 2024-04-10
Remove some dead code and refactor

cjpatton:
The goal of this PR is to remove some tech debt. It mostly removes some dead code, but there's some refactoring and a few additional tests.


mendess (approved):
Good one!

PR #575, opened 2024-04-11
Initial heavy hitters support via Mastic

cjpatton:
~Based on #574 (merge that first).~

This is a prototype of proposal 1 from:
https://docs.google.com/document/d/1ZjXz-1kGsTDf2Vn2u-fwYqR8BSc3tOYlELVHAYvAfjk

Daphne will use the Mastic VDAF rather than Poplar1 because we only support 1-round VDAFs. (Poplar1 requires two rounds for preparation.) For now we're working with a dummy version of Mastic, as its implementation is still in progress:
https://github.com/divviup/libprio-rs/issues/947

This is not ready to merge. We need to:
- [ ] Align with the changes we settle on in the DAP specification (some changes will be required if we take proposal 2)
- [ ] Consider fencing changes to `protocol` and `messages` behind the "test-utils" feature
- [ ] Plumb the new code path through `roles` module
- [ ] Additional tests
     - Leader tries to override an existing report
     - Stored report does not exist
     - ...
- [ ] Replace dummy Mastic with the real one
- [ ] Implement VIDPF caching

### DIFF
--- a/crates/dapf/src/acceptance/mod.rs
+++ b/crates/dapf/src/acceptance/mod.rs
@@ -471,7 +471,7 @@ impl Test {
                 .put(url)
                 .body(
                     agg_job_init_req
-                        .get_encoded_with_param(&task_config.version)
+                        .get_encoded_with_param(&(task_config.version, false))
                         .unwrap(),
                 )
                 .headers(headers),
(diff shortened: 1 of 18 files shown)


cjpatton:
DAP will not support heavy hitters: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/554

I've closed this, but created a branch (`poc/cjatton/dap-with-heavy-hitters`) to refer back to later if needed.

PR #579, opened 2024-04-16
Reduce the size of the interop test image

cjpatton:
The main goal is to copy the service binary and storage proxy worker into the final layer. We also observe that we don't need wrangler until we're ready to invoke the worker; all we need to build the worker is `worker-build`.

### DIFF
--- a/daphne_worker_test/wrangler.storage_proxy.toml
+++ b/daphne_worker_test/wrangler.storage_proxy.toml
@@ -5,11 +5,12 @@ name = "daphne_storage_proxy"
 main = "build/worker/shim.mjs"
 compatibility_date = "2023-12-21"
 
-# This Worker requires `worker-build` in order to package Rust for WASM. In
-# most cases it is preferable to install it ahead of time so that `wrangler
-# dev` starts faster.
-[build]
-command = "cargo install --git https://github.com/cloudflare/workers-rs && worker-build --dev"
+# Don't ask to send metrics to Cloudflare. The worker may be run from a container.
+send_metrics = false
+
+# Before starting the worker, run `worker-build`.
+#[build]
+#command = "cargo install --git https://github.com/cloudflare/workers-rs && worker-build --dev"
 
 [[rules]]
 globs = ["**/*.wasm"]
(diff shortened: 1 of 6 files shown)


daphne_worker_test/wrangler.storage_proxy.toml
```diff
@@ -5,11 +5,12 @@ name = "daphne_storage_proxy"
...
-# most cases it is preferable to install it ahead of time so that `wrangler
-# dev` starts faster.
-[build]
-command = "cargo install --git https://github.com/cloudflare/workers-rs && worker-build --dev"
+# Don't ask to send metrics to Cloudflare. The worker may be run from a container.
+send_metrics = false
+
+# Before starting the worker, run `worker-build`.
+#[build]
+#command = "cargo install --git https://github.com/cloudflare/workers-rs && worker-build --dev"
```
cjpatton, line 13: Expect CI to fail due to this change. We need to modify the other images so that `worker-build` is called prior to invoking the worker.

interop/Dockerfile.interop_helper
```diff
@@ -1,55 +1,53 @@
...
+# Build the service.
+FROM build-deps-common AS builder-service
+WORKDIR /build
+COPY Cargo.toml Cargo.lock /build/
+COPY daphne /build/daphne
+COPY daphne_server /build/daphne_server
+COPY daphne_service_utils /build/daphne_service_utils
+RUN cargo new --lib daphne_worker  # mock workspace dependency
+RUN cargo new --lib daphne_worker_test  # mock workspace dependency
+RUN cargo build --example service --features test-utils
```
divergentdave, line 24: This is missing `--release`, I get a file not found error from a later COPY.

divergentdave (approved):
FWIW, the separation between build-deps-storage-proxy and builder-storage-proxy is not necessary, since you could combine the steps under one name without issue, but I see how the extra split improves self-documentation.

interop/Dockerfile.interop_helper
```diff
@@ -1,55 +1,53 @@
...
+COPY daphne_service_utils /build/daphne_service_utils
+COPY daphne_worker /build/daphne_worker
+COPY daphne_worker_test /build/daphne_worker_test
 WORKDIR /build/daphne_worker_test
+RUN worker-build --dev
+
+# Prepare the environment in which the service and storage proxy will run.
+FROM node:bookworm AS final
+RUN apt update && apt install -y colorized-logs
+RUN npm install -g wrangler@3.50.0
```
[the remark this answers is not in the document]
cjpatton: I'll take it!

PR #583, opened 2024-04-19
Reorganize crates inside a crates directory

mendess:


cjpatton (approved):

mendess:
I made a few new commits since you reviewed so I'm asking for your review again.

- I made the toml simpler and pull more info from the workspace toml, this way we have more consistency and less redundancy of information
- I renamed every crate to use `-` instead of `_`. This seems to be the more common rust way to name packages and `cargo deb` also replaces `_` with `-` which can be confusing

cjpatton:
> I made a few new commits since you reviewed so I'm asking for your review again.
> 
>     * I made the toml simpler and pull more info from the workspace toml, this way we have more consistency and less redundancy of information
> 
>     * I renamed every crate to use `-` instead of `_`. This seems to be the more common rust way to name packages and `cargo deb` also replaces `_` with `-` which can be confusing

I say everything but the kebab-case change. LGTM!

cjpatton (approved):

PR #595, opened 2024-05-16
Bump worker to 0.2.0

mendess:


cjpatton:
wow, that wasn't bad at all!

cjpatton (approved):

PR #598, opened 2024-05-17
Downgrade prio and cleanup some dependencies

mendess:
- Upgrade to prio 0.16.5
- Move e2e to daphne-server


mendess:
I thought moving the tests to daphne-server would let us get rid of the getrandom hack, turns out it doesn't. I kept it anyway because I think it still makes sense.

When/if https://github.com/divviup/libprio-rs/pull/1059 gets merged we can finally get rid of the getrandom hack

Cargo.toml
```diff
@@ -56,7 +57,7 @@ matchit = "0.7.3"
 p256 = { version = "0.13.2", features = ["ecdsa-core", "ecdsa", "pem"] }
 paste = "1.0.14"
 pin-project = "1.1.5"
-prio = "0.16.4"
+prio = "0.16.3"
```
cjpatton, line 60: Consider pinning the version, and perhaps the getrandom version.
```suggestion
prio = "=0.16.3"
```

cjpatton (approved):
Stamped to unblock, but we should try to avoid downgrading if we can. The speed up will be helpful.

PR #599, opened 2024-05-20
Drop bincode in favor of capnproto for DO message serialization

mendess:


crates/daphne-worker-test/wrangler.storage-proxy.toml
```diff
@@ -71,3 +70,7 @@ tag = "v2"
 renamed_classes = [
     { from = "GarbageCollector", to = "TestStateCleaner" },
 ]
+
+[[migrations]]
+tag = "v3"
+deleted_classes = ["HelperStateStore"]
```
cjpatton, line 76: I don't think we need to document the migration, since it's not actually part of a deployment.

crates/daphne-worker/Cargo.toml
```diff
@@ -16,6 +16,7 @@ description = "Workers backend for Daphne"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
+# TODO: delete
```
cjpatton, line 19:
```suggestion
# TODO(mendess): delete
```

crates/daphne-service-utils/src/durable_requests/bindings/aggregate_store.rs
```diff
@@ -0,0 +1,264 @@
...
+        ))
+    }
+}
+
+fn durable_name_task(version: DapVersion, task_id_hex: &str) -> String {
+    format!("{}/task/{}", version.as_ref(), task_id_hex)
+}
+
+#[derive(Debug, PartialEq, Eq)]
+// TODO(mendess): delete. Only here to suport `legacy_req_parse`
```
cjpatton, line 45: I can't find anything called "legacy_req_parse" in this PR.

crates/daphne-service-utils/src/durable_requests/durable_request.capnp
```diff
@@ -23,3 +23,28 @@ struct DurableRequest @0xfbd55b93d47690b9 {
...
+struct DapAggregateShare @0xb34ce529a4a66aed {
+    reportCount @0 :UInt64;
+    minTime @1 :UInt64;
+    maxTime @2 :UInt64;
+    checksum @3 :Data;
+    data :union {
+        field64 @4 :Data;
+        field128 @5 :Data;
+        fieldPrio2 @6 :Data;
+        none @7 :Void;
```
cjpatton, line 41: Is this to cover the fact that `data` is an `Option`?

cjpatton (commented):
This isctually a pretty major refactor, yet there is no documentation of what is changing in the commit message. The title itself is also only about 30% of what is actually happening here. Some specific things I think need to be documented:
* the fact that we're trying to be backwards compatible. You should also say something about how fallback works. (This isn't immediately clear to me from looking at the code.)
* HelperStateStore is being deleted. You should also say why it's being deleted.

crates/daphne-worker/src/durable/mod.rs
```diff
@@ -199,10 +199,19 @@ pub(crate) async fn state_set_if_not_exists<T: for<'a> Deserialize<'a> + Seriali
...
+{
     let bytes = req.bytes().await?;
-    bincode::deserialize(&bytes)
-        .map_err(|e| Error::RustError(format!("failed to deserialize bincode: {e:?}")))
+    T::decode_from_bytes(&bytes)
+        .map_err(|e| Error::RustError(e.to_string()))
+        .or_else(|_| {
+            bincode::deserialize(&bytes)
+                .map_err(|e| Error::RustError(format!("failed to deserialize bincode: {e:?}")))
+        })
```
cjpatton, line 214: It might be helpful to temporarily log the cpanproto decode failure.

cjpatton (approved):

crates/daphne-worker/src/durable/mod.rs
mendess: I'll just include both in the error message instead.

PR #596, opened 2024-05-16
taskprov: Ignore unrecognized DP mechanisms

cjpatton:
For now we don't implement any DP mechanism.

PR #603, opened 2024-05-21
Revert bad changes from upstream PR

mendess:
revert bad changes made in #600

### DIFF
--- a/crates/daphne/src/taskprov.rs
+++ b/crates/daphne/src/taskprov.rs
@@ -557,19 +557,65 @@ mod test {
             .unwrap()
             .config;
 
-        match (
-            version,
-            resolve_advertised_task_config(&req, &[0; 32], &collector_hpke_config, &task_id),
-        ) {
-            (
-                DapVersion::Draft09 | DapVersion::Latest,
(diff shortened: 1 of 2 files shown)


crates/daphne/src/taskprov.rs
```diff
@@ -572,4 +572,57 @@ mod test {
     }
 
     test_versions! { resolve_advertised_task_config_expect_abort_unrecognized_vdaf }
+
```
cjpatton, line 575: We already had this test: https://github.com/cloudflare/daphne/pull/596/files#diff-e8ba4a25150d4a6f4c52e3735a7b4b04a034a096b35163063bcbe9d4914b8dadR512

It looks like we deleted it in https://github.com/cloudflare/daphne/pull/600. Let's double check that we didn't inadvertenly remove other code in that PR.

"we_accept_all_dp_configuration_parameters" is an odd name for a test. (Who is "we"?)

I think it would be better to test resolve_advertised_task_config().

cjpatton (approved):

cjpatton (approved):

PR #610, opened 2024-05-28
Create runtime override to disable replay protection

mendess:


cjpatton:
Looks like this PR is stacked on #599? I'd suggest changing base.

crates/daphne-server/src/roles/aggregator.rs
```diff
@@ -38,6 +40,19 @@ impl DapAggregator<DaphneAuth> for crate::App {
...
+        let skip_replay_protection = self
+            .kv()
+            .get_cloned::<kv::prefix::GlobalConfigOverride<bool>>(
+                &kv::prefix::GlobalOverrides::SkipReplayProtection,
+                &KvGetOptions {
+                    cache_not_found: true,
+                },
+            )
+            .await
+            .ok() // treat error as false
```
cjpatton, line 52: Would it be better to abort? We should at least log the error.

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -39,13 +39,30 @@ pub mod prefix {
     use super::KvPrefix;
 
     pub struct GlobalConfigOverride<V>(PhantomData<V>);
+    #[derive(Debug)]
+    #[cfg_attr(feature = "test-utils", derive(serde::Serialize, serde::Deserialize))]
```
cjpatton, line 43:
```suggestion

    /// XXX <- documentation
    #[derive(Debug)]
    #[cfg_attr(feature = "test-utils", derive(serde::Serialize, serde::Deserialize))]
```

crates/daphne-service-utils/src/durable_requests/bindings/aggregate_store.rs
```diff
@@ -47,6 +47,14 @@ fn durable_name_task(version: DapVersion, task_id_hex: &str) -> String {
...
     pub contained_reports: Vec<ReportId>,
     pub agg_share_delta: DapAggregateShare,
+    pub options: AggregateStoreMergeOptions,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+// TODO(mendess): delete. Only here to suport `legacy_req_parse`
+#[derive(serde::Deserialize)]
+pub struct AggregateStoreMergeOptions {
+    pub skip_replay_protection: bool,
```
cjpatton, line 57: It's notable that skipping replay protection degrades privacy by allowing a malicious co-Aggregator to replay reports. I think this needs to be reflected in how we document this feature. We should note (1) privacy is degraded and (2) it's intended to be used in an incident and needs to be re-enabled after the incident is mitigated.

crates/daphne-server/src/roles/aggregator.rs
mendess: Abort could cause an incident. This is not fatal, if we can't fetch from kv:
- and we abort we create an incident
- we assume true we have a safety hole where disruptions in KV service lower our privacy guarantees
- we assume false we keep the safety guarantees even if that means crashing later anyway because we are in an incident

Logging makes sense though. Added

cjpatton (approved):

PR #612, opened 2024-06-03
Expose wait_until method on instantiated objects

mendess:


crates/daphne-worker/src/durable/mod.rs
```diff
@@ -321,6 +326,15 @@ macro_rules! instantiate_durable_object {
                     self.inner.alarm().await
                 }
             }
+
+            impl $name {
+                pub fn wait_until<F>(&self, future: F)
```
cjpatton, line 331: What does this do?

cjpatton (approved):

PR #613, opened 2024-06-05
Don't use the "wasm-compat" feature for prio

cjpatton:
This feature is intended to ensure compatibility with WASM. Instead, add `getrandom` to the workspace dependencies with the "js" feature enabled. This has the same effect according to:
https://docs.rs/getrandom/0.2.15/getrandom/#webassembly-support

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -858,6 +858,7 @@ dependencies = [
  "daphne",
  "daphne-service-utils",
  "futures",
+ "getrandom",
  "hex",
  "paste",
  "prio",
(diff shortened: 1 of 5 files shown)


Noah-Kennedy (approved):

Noah-Kennedy:
> Shouldn't we use `getrandom = "*"`? That way it will always be the same version of `getrandom` that `prio` picked? Instead of ending up with two different versions of getrandom being pulled.
> 
> I'm not sure if this is a good idea, I've never done it this way

No, that can lead to us getting bumped to newer. There is never a case to use `= "*"`.

PR #616, opened 2024-06-11
Refactor GetResult in kv/cache.rs to CacheResult

mendess:


crates/daphne-server/src/storage_proxy_connection/kv/cache.rs
```diff
@@ -30,12 +30,9 @@ pub struct Cache {
...
 
 pub enum GetResult<T: 'static> {
-    NotFound {
-        /// Whether the caller should check KV for the value.
-        read_through: bool,
-    },
+    NotFound,
     MismatchedType,
-    Found(Marc<T>),
+    Found(Option<Marc<T>>),
```
cjpatton, line 35: Is this what you intend?
```suggestion
    /// None indicates no value was found, but that the caller should cache the negative response.
    Found(Option<Marc<T>>),
```

crates/daphne-server/src/storage_proxy_connection/kv/cache.rs
```diff
@@ -30,12 +30,9 @@ pub struct Cache {
 }
 
 pub enum GetResult<T: 'static> {
-    NotFound {
-        /// Whether the caller should check KV for the value.
-        read_through: bool,
-    },
+    NotFound,
```
cjpatton, line 33: Is this what you intend?
```suggestion
    /// The caller should read through to KV.
    NotFound,
```

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -212,12 +212,8 @@ impl<'h> Kv<'h> {
...
         tracing::debug!(key, "GET");
         match self.cache.read().await.get::<P>(&key) {
-            cache::GetResult::NotFound { read_through } => {
-                if !read_through {
-                    return Ok(None);
-                }
-            }
-            cache::GetResult::Found(t) => return Ok(Some(mapper(t))),
+            cache::GetResult::NotFound => {}
+            cache::GetResult::Found(t) => return Ok(t.map(mapper)),
```
cjpatton, line 216: We want to return `Ok(None)` if `!read_through`, correct? Can you add a comment to clarify?

cjpatton (changes requested):
Please add some documentation. Also the commit message needs more context than "Simplify GetResult".

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
mendess: Can you review the code now that I changed the name of the variants? Maybe it's more obvious now and doesn't need a comment

crates/daphne-server/src/storage_proxy_connection/kv/cache.rs
mendess: Don't think about it in terms of the cache telling the caller whether it should read through KV and more in terms of "there was a cache hit or a cache miss" and the caller then decides what it wants to do on either circumstance

crates/daphne-server/src/storage_proxy_connection/kv/cache.rs
```diff
@@ -29,44 +29,32 @@ pub struct Cache {
...
 }
 
-pub enum GetResult<T: 'static> {
-    NotFound {
-        /// Whether the caller should check KV for the value.
-        read_through: bool,
-    },
+pub enum CacheResult<T: 'static> {
+    Hit(Option<Marc<T>>),
+    Miss,
```
cjpatton, line 34: `Hit` versus `Miss` is slightly confusing because `Hit(None)` is actually a miss as well. I would at least clarify in a comment that the caller is meant to skip reading through to KV if `None`.

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -212,13 +212,9 @@ impl<'h> Kv<'h> {
...
         match self.cache.read().await.get::<P>(&key) {
-            cache::GetResult::NotFound { read_through } => {
-                if !read_through {
-                    return Ok(None);
-                }
-            }
-            cache::GetResult::Found(t) => return Ok(Some(mapper(t))),
-            cache::GetResult::MismatchedType => {
+            cache::CacheResult::Miss => {}
+            cache::CacheResult::Hit(t) => return Ok(t.map(mapper)),
```
cjpatton, line 216: I still think a code comment is needed to parse this, but if you disagree, disregard.

crates/daphne-server/src/storage_proxy_connection/kv/cache.rs
mendess: It's not a miss, the cache correctly cached that the value is not present, not present being a valid value for the given key

cjpatton: Ah I see. I have a different understanding of what "Hit" is supposed to mean, so I do think a comment to clarify would help.

mendess: Yes, this PR effectively wants to change the meaning of `Hit` to not have any special meaning, it means the wikipedia definition of a cache hit

> A cache hit occurs when the requested data can be found in a cache

hence why I believe it doesn't require a comment, my intent is that you can tell from the type that a `None` is a valid `Hit`

cjpatton, line 34:
```suggestion
    /// Cache hit. `None` indicates that the value is known to not exist and that the caller should not read through to KV. 
    Hit(Option<Marc<T>>),
    /// Cache miss. The caller should read through to KV.
    Miss,
```

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -207,18 +207,14 @@ impl<'h> Kv<'h> {
...
```
cjpatton, line 216:
```suggestion
            /// Return the cached response immediately. If the value is known to not exist, then skip reading through to KV.
            cache::CacheResult::Hit(t) => return Ok(t.map(mapper)),
```

cjpatton (approved):

PR #619, opened 2024-06-14
Expire task info stored in KV

lbaquerofierro:


crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -256,20 +256,54 @@ impl<'h> Kv<'h> {
...
+
+        // Create the body with optional expiration
+        let mut body = serde_json::json!({
+            "value": value,
+        });
+
+        if let Some(exp) = expiration {
+            body["expiration"] = serde_json::json!(exp);
+        }
+
```
mendess, line 298: This has to be done in the storage_proxy, I think to keep things backwards compatible I'd pass the expiration through a header like `X-expiration` or something close to that

mendess (changes requested):

Noah-Kennedy (approved):

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -252,28 +252,74 @@ impl<'h> Kv<'h> {
     }
 
     #[tracing::instrument(
-        name = "kv_put",
+        name = "kv_delete",
```
cjpatton, line 255: Why is this changing? I'm sure there's a good reason, but I don't understand it :)

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -252,28 +252,74 @@ impl<'h> Kv<'h> {
...
         skip_all,
         fields(key, prefix = std::any::type_name::<P>()),
     )]
-    pub async fn put<P>(&self, key: &P::Key, value: P::Value) -> Result<(), Error>
+    async fn put_internal<P>(
+        &self,
+        key: &P::Key,
+        value: P::Value,
+        expiration: Option<u64>,
```
cjpatton, line 263: For clarity (`task_config.not_after` has type `Time`)
```suggestion
        expiration: Option<Time>,
```

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -252,28 +252,74 @@ impl<'h> Kv<'h> {
...
+            .http
             .post(self.config.url.join(&key).unwrap())
             .bearer_auth(&self.config.auth_token)
-            .body(serde_json::to_vec(&value).unwrap())
-            .send()
-            .await?
-            .error_for_status()?;
+            .body(serde_json::to_vec(&value).unwrap());
+
+        // Add expiration if provided
```
cjpatton, line 279:
```suggestion
        // Add expiration if provided.
```

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -252,28 +252,74 @@ impl<'h> Kv<'h> {
...
+        if let Some(exp) = expiration {
+            put_request = put_request.json(&serde_json::json!({ "expiration": exp }));
```
cjpatton, line 281: Always useful to add references to Workers APis where relevant:
https://developers.cloudflare.com/kv/api/write-key-value-pairs/#create-expiring-keys

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -252,28 +252,74 @@ impl<'h> Kv<'h> {
...
+        }
+
+        put_request.send().await?.error_for_status()?;
+
         self.cache.write().await.put::<P>(key, Some(value.into()));
         Ok(())
     }
 
+    #[tracing::instrument(
+        name = "kv_delete",
```
cjpatton, line 291: Is this correct? cc/ @mendess

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -252,28 +252,74 @@ impl<'h> Kv<'h> {
...
+        skip_all,
+        fields(key, prefix = std::any::type_name::<P>()),
+    )]
+    pub async fn put_with_expiration<P>(
+        &self,
+        key: &P::Key,
+        value: P::Value,
+        expiration: u64,
```
cjpatton, line 299:
```suggestion
        expiration: Time,
```

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
cjpatton, line 281: If I'm not mistaken, this actually replaces the payload of the request. I.e., it overwrites the `.body(serde_json::to_vec(&value).unwrap())` on line 277 above. Am I reading this right?

cjpatton: Ah, you're right!

mendess (changes requested):

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -252,28 +253,64 @@ impl<'h> Kv<'h> {
...
+        value: P::Value,
+        expiration: Option<Time>,
+    ) -> Result<(), Error>
     where
         P: KvPrefix,
         P::Key: std::fmt::Debug,
+        P::Value: Serialize,
     {
         let key = Self::to_key::<P>(key);
+        // let expiration_str = expiration.map_or_else(String::new, |time| time.to_string());
```
mendess, line 272: letftover comment?

crates/daphne-worker/src/storage_proxy/mod.rs
```diff
@@ -223,12 +223,24 @@ async fn handle_kv_request(ctx: &mut RequestContext<'_>, key: &str) -> worker::R
...
         }
         worker::Method::Post => {
+            let expiration_header = ctx.req.headers().get("X-Expiration")?;
+            let expiration_unix_timestamp = if let Some(expiration) = expiration_header {
+                expiration.parse::<u64>().map_err(|e| {
+                    worker::Error::RustError(format!("Failed to parse expiration header: {e:?}"))
+                })
+            } else {
+                Ok(0)
+            }?;
```
mendess, line 233: instead of using a magic 0 value, you can use option
```suggestion
            let expiration_unix_timestamp = expiration_header.map(|ex| ex.parse()).transpose().map_err(|e| {
                worker::Error::RustError(format!("Failed to parse expiration header: {e:?}"))
            })?;
```

if transpose is too cryptic (I sometimes find it a bit cryptic) you can do this instead.
```suggestion
            let expiration_unix_timestamp = if let Some(expiration) = expiration_header {
                Some(expiration.parse::<u64>().map_err(|e| {
                    worker::Error::RustError(format!("Failed to parse expiration header: {e:?}"))
                })?)
            } else {
                None
            }?;
```

crates/daphne-worker/src/storage_proxy/mod.rs
```diff
@@ -246,7 +258,17 @@ async fn handle_kv_request(ctx: &mut RequestContext<'_>, key: &str) -> worker::R
...
+
         worker::Method::Put => {
+            let expiration_header = ctx.req.headers().get("X-Expiration")?;
+            let expiration_unix_timestamp = if let Some(expiration) = expiration_header {
+                expiration.parse::<u64>().map_err(|e| {
+                    worker::Error::RustError(format!("Failed to parse expiration header: {e:?}"))
+                })
+            } else {
+                Ok(0)
+            }?;
```
mendess, line 270: same here

mendess (changes requested):

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
mendess: Why was this resolved? It should be `kv_put`

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -252,28 +253,64 @@ impl<'h> Kv<'h> {
...
+
         self.cache.write().await.put::<P>(key, Some(value.into()));
         Ok(())
     }
 
+    pub async fn put_with_expiration<P>(
+        &self,
+        key: &P::Key,
+        value: P::Value,
+        expiration: Option<Time>,
```
cjpatton, line 295: nit: since the caller is already opting into using expiration, consider making it not optional and construction `Some(expiration)` within the function.
```suggestion:
        expiration: Time,
```

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -252,28 +253,64 @@ impl<'h> Kv<'h> {
...
             .post(self.config.url.join(&key).unwrap())
             .bearer_auth(&self.config.auth_token)
-            .body(serde_json::to_vec(&value).unwrap())
-            .send()
-            .await?
-            .error_for_status()?;
+            .body(serde_json::to_vec(&value).unwrap());
+
+        if let Some(expiration) = expiration {
+            request = request.header("X-Expiration", expiration.to_string());
```
cjpatton, line 282: This is a header intended for a public API. To minimize the chance of someone accidentally overwriting this in a request, I think we should make it more specific. Perhaps something like
```suggestion
            request = request.header("x-daphne-storage-proxy-kv-put-expiration", expiration.to_string());
```
Also, I believe lower case is preferred. (Compliant HTTP libraries will always convert to lower case.)

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -252,28 +253,64 @@ impl<'h> Kv<'h> {
...
+        key: &P::Key,
+        value: P::Value,
+        expiration: Option<Time>,
+    ) -> Result<(), Error>
+    where
+        P: KvPrefix,
+        P::Key: std::fmt::Debug,
+        P::Value: Serialize,
+    {
+        self.put_internal::<P>(key, value, expiration).await
```
cjpatton, line 302: I haven't checked, but the compiler may be able to infer this type. Here and below.
```suggestion
        self.put_internal(key, value, expiration).await
```

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
```diff
@@ -294,13 +332,18 @@ impl<'h> Kv<'h> {
...
             .http
             .put(self.config.url.join(&key).unwrap())
             .bearer_auth(&self.config.auth_token)
-            .body(serde_json::to_vec(&value).unwrap())
-            .send()
-            .await?;
+            .body(serde_json::to_vec(&value).unwrap());
+
+        if let Some(expiration) = expiration {
+            request = request.header("X-Expiration", expiration.to_string());
```
cjpatton, line 343: Let's make the header name a constant that we reuse rather than re-typing the raw string everywhere.

```rust
pub const STORAGE_PROXY_PUT_KV: &str = "...";
```

mendess: there's a module for this as well `crates/daphne-service-utils/src/http_headers.rs` so you can share the constant between services

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
lbaquerofierro: Sorry about that, changed only one instance of this error and resolved. Corrected

crates/daphne-worker/src/storage_proxy/mod.rs
lbaquerofierro: Thanks I didn't know about transpose. Pretty useful

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
lbaquerofierro: Got an error when tried it
` ^^^^^^^^^^^^ cannot infer type of the type parameter `P` declared on the method `put_internal``

crates/daphne-server/src/storage_proxy_connection/kv/mod.rs
lbaquerofierro: Changed it for `put_with_expiration`

crates/daphne-worker/src/storage_proxy/mod.rs
```diff
@@ -209,12 +210,26 @@ async fn handle_kv_request(ctx: &mut RequestContext<'_>, key: &str) -> worker::R
             }
         }
         worker::Method::Post => {
+            let expiration_header = ctx.req.headers().get(STORAGE_PROXY_PUT_KV_EXPIRATION)?;
+            let expiration_unix_timestamp = expiration_header
+                .map(|expiration| {
+                    expiration.parse::<u64>().map_err(|e| {
```
cjpatton, line 216:
```suggestion
                    expiration.parse::<Time>().map_err(|e| {
```

crates/daphne-worker/src/storage_proxy/mod.rs
```diff
@@ -232,7 +247,19 @@ async fn handle_kv_request(ctx: &mut RequestContext<'_>, key: &str) -> worker::R
...
         worker::Method::Put => {
+            let expiration_header = ctx.req.headers().get(STORAGE_PROXY_PUT_KV_EXPIRATION)?;
+            let expiration_unix_timestamp = expiration_header
+                .map(|expiration| {
+                    expiration.parse::<u64>().map_err(|e| {
+                        worker::Error::RustError(format!(
+                            "Failed to parse expiration header: {e:?}"
+                        ))
+                    })
+                })
```
cjpatton, line 260: Consider reactoring this code into a function that is used here and above.

cjpatton (changes requested):
Implementation looks good! There's one more place where we should add an expiration date. In fact, this one is probably the most important because it has the highest write rate: https://github.com/cloudflare/daphne/blob/main/crates/daphne-server/src/roles/aggregator.rs#L291

cjpatton (approved):

mendess (approved):

PR #621, opened 2024-06-18
Allow disabling replay protection in daphne-server

mendess:


crates/daphne-server/src/roles/aggregator.rs
```diff
@@ -41,21 +44,10 @@ impl DapAggregator<DaphneAuth> for crate::App {
...
-                },
-            )
-            .await
-            .inspect_err(
-                |e| tracing::error!(error = ?e, "failed to fetch skip_replay_protection from kv"),
-            )
-            .ok() // treat error as false
-            .flatten()
-            .unwrap_or_default(); // treat missing as false
+        let skip_replay_protection = matches!(
```
fisherdarling, line 47: I see there's a helper for `enabled()`: https://github.com/cloudflare/daphne/pull/621/files#diff-7c6740c5b3ceb5af4a8f7daa7f7e1adec53ccf3dda8eb714c918fcd860697dbfR385

Should we make one for disabled too?

crates/daphne/src/protocol/aggregator.rs
```diff
@@ -398,7 +411,16 @@ impl DapTaskConfig {
             let mut processed = HashSet::with_capacity(report_count_hint);
             let mut reports = pin!(reports);
             while let Some(report) = reports.next().await {
-                if processed.contains(&report.report_metadata.id) {
+                let enforce_replay_check = {
+                    #[cfg(feature = "report-generator")]
+                    {
+                        use std::sync::atomic::Ordering;
```
fisherdarling, line 417: Since we're using a setter for the atomic: https://github.com/cloudflare/daphne/pull/621/files#diff-5bbd4da852a4404371d706296eb9269892f0d85ded2d6570bcc7de42d6c82c35R60
we could consider adding a getter to clean up usage here.

crates/daphne-service-utils/src/durable_requests/bindings/aggregate_store.rs
```diff
@@ -128,6 +133,13 @@ impl DurableRequestPayload for AggregateStoreMergeReq {
                 };
             }
         }
+        {
+            let AggregateStoreMergeOptions {
```
fisherdarling, line 137: Is this destructuring required? Can we do `*options.skip_replay_protection`?

mendess: Destructuring here guarantees that if this struct changes this code stops compiling, which is a good thing because we probably need to change it

Noah-Kennedy (approved):

crates/daphne-service-utils/src/durable_requests/bindings/aggregate_store.rs
fisherdarling: Nice! I like it

PR #626, opened 2024-06-21
Move metrics registry instantiation responsability to caller of daphne-worker

mendess:


cjpatton:
I'll review once CI passes.

Issue #618, opened 2024-06-12
Add support for draft-chen-cfrg-vdaf-pine


cjpatton:
- [X] Upstream implementation of draft 00.
- [x] Replace VDAF test driver with the new one in the `prio` crate (with feature `"test-utils"`).
- [x] Plumb into daphne configuration.
- [x] Update to draft 01 (work-in-progress)
- [x] Plumb into taskprov.

PR #629, opened 2024-06-26
pine: Use VDAF test driver from the `prio` crate

cjpatton:
Based on #632 (merge that first).
Partially addresses #618.


cjpatton:
Rebased.

mendess (approved):

PR #641, opened 2024-07-29
Fix clippy for 1.80

cjpatton:
Also, make sure we're building the benches.

### DIFF
--- a/.github/workflows/daphneci.yml
+++ b/.github/workflows/daphneci.yml
@@ -32,6 +32,8 @@ jobs:
         run: cargo fmt --all --check
       - name: Documentation
         run: RUSTDOCFLAGS='-D warnings' cargo doc --locked --no-deps --all-features --workspace
+      - name: Build
+        run: cargo build --all-targets
       - name: Linting
         run: cargo hack clippy --locked --each-feature -- -D warnings
       - name: Testing
--- a/crates/daphne/benches/pine.rs
+++ b/crates/daphne/benches/pine.rs
@@ -17,7 +17,7 @@ fn pine(c: &mut Criterion) {
         (100_000, 320 * 6),
         (1_000_000, 1_000 * 8),
     ] {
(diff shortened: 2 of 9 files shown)


RageKnify (approved):

crates/daphne/benches/pine.rs
```diff
@@ -17,7 +17,7 @@ fn pine(c: &mut Criterion) {
         (100_000, 320 * 6),
         (1_000_000, 1_000 * 8),
     ] {
-        let pine = Pine::new_64(1.0, dimension, 15, chunk_len).unwrap();
+        let pine = Pine::new_64(1 << 15, dimension, 15, chunk_len).unwrap();
```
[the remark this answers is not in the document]
cjpatton: Actually, this has to do with the change from `f64` to `u64` in https://github.com/cloudflare/daphne/pull/636. In more detail, the L2 norm bound is a real number that we encode in a particular way into an integer by multiplying by the 2^frac_bits, which here is 15.

.github/workflows/daphneci.yml
```diff
@@ -32,6 +32,8 @@ jobs:
         run: cargo fmt --all --check
       - name: Documentation
         run: RUSTDOCFLAGS='-D warnings' cargo doc --locked --no-deps --all-features --workspace
+      - name: Build
+        run: cargo build --all-targets --all-features
```
mendess, line 36: why add this?

cjpatton: CI was not building the benchmarks and did not detect a build error in daphne/benches/pine.rs.

mendess: can you do only `--all-targets`? `--all-features` can cause weird issues where features that shouldn't be enabled together are enabled together

cjpatton: Done.

mendess (approved):

PR #642, opened 2024-07-29
protocol: Handle invalid measurement for PINE more gracefully

cjpatton:
Partially addresses #618.

It is fatal for the client to handle the wrong measurement type. When this happens for PINE, we currently output a useless error message. Replace this with a message that's more useful.


mendess (approved):
:rofl:

PR #645, opened 2024-08-06
storage: add metrics for KV and DO request timings

Noah-Kennedy:


lbaquerofierro (approved):

cjpatton (changes requested):
Please fix the clippy warnings before merging.

Noah-Kennedy:
> Please fix the clippy warnings before merging.

One step ahead of you!

cjpatton (approved):

crates/daphne-worker/src/storage_proxy/mod.rs
```diff
@@ -210,13 +210,33 @@ fn parse_expiration_header(ctx: &RequestContext) -> Result<Option<Time>, worker:
 
 /// Handle a kv request.
 async fn handle_kv_request(ctx: &mut RequestContext<'_>, key: &str) -> worker::Result<Response> {
+    let start = std::time::Instant::now();
```
RageKnify, line 213: Shouldn't this be `Worker::Date::now()`? I believe WASM doesn't implement the `std::time` module. For plexi I used `Worker::Date::now()` to measure the time to process requests.

mendess: I believe it should be. Some docs on the subject https://developers.cloudflare.com/workers/runtime-apis/performance/

mendess: You were right, check #647 on how to forbid that error from ever compiling

PR #646, opened 2024-08-12
Update expirtaion time in taksprov opt in config function 

lbaquerofierro:
Make `task_config` mutable to avoid value move issues in `taskprov_opt_infunction`, ensuring correct expiration time from `DapTaskConfig`

### DIFF
--- a/crates/daphne-server/src/roles/aggregator.rs
+++ b/crates/daphne-server/src/roles/aggregator.rs
@@ -240,36 +240,36 @@ impl DapAggregator<DaphneAuth> for crate::App {
         task_config: taskprov::DapTaskConfigNeedsOptIn,
         global_config: &DapGlobalConfig,
     ) -> Result<DapTaskConfig, DapError> {
-        // For now we always opt-in.
-        //
-        // Opt-in parameters are backed by KV. If not found, then use the default number of shards.
-        let param = if let Some(param) = self
+        if let Some(param) = self
             .kv()
             .get_cloned::<kv::prefix::TaskprovOptInParam>(task_id, &KvGetOptions::default())
             .await
             .map_err(|e| fatal_error!(err = ?e))?
         {
-            param
+            Ok(task_config.into_opted_in(&param))
         } else {
             let param = taskprov::OptInParam {
                 not_before: self.get_current_time(),
                 num_agg_span_shards: global_config.default_num_agg_span_shards,
             };
+
+            let task_config = task_config.into_opted_in(&param);
+            let expiration_time = task_config.not_after;
+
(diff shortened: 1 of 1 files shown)


crates/daphne-server/src/roles/aggregator.rs
```diff
@@ -237,7 +237,7 @@ impl DapAggregator<DaphneAuth> for crate::App {
     async fn taskprov_opt_in(
         &self,
         task_id: &TaskId,
-        task_config: taskprov::DapTaskConfigNeedsOptIn,
+        mut task_config: taskprov::DapTaskConfigNeedsOptIn,
```
mendess, line 240: this doesn't need to be mutable

mendess (changes requested):
I would use the entire if as an expression to avoid having to look up and down to follow the code
```rs
    async fn taskprov_opt_in(
        &self,
        task_id: &TaskId,
        mut task_config: taskprov::DapTaskConfigNeedsOptIn,
        global_config: &DapGlobalConfig,
    ) -> Result<DapTaskConfig, DapError> {
        if let Some(param) = self
            .kv()
            .get_cloned::<kv::prefix::TaskprovOptInParam>(task_id, &KvGetOptions::default())
            .await
            .map_err(|e| fatal_error!(err = ?e))?
        {
            Ok(task_config.into_opted_in(&param))
        } else {
            let param = taskprov::OptInParam {
                not_before: self.get_current_time(),
                num_agg_span_shards: global_config.default_num_agg_span_shards,
            };

            let task_config = task_config.into_opted_in(&param);
            let expiration_time = task_config.not_after;

            if let Err(e) = self
                .kv()
                .put_with_expiration::<kv::prefix::TaskprovOptInParam>(
                    task_id,
                    param.clone(),
                    expiration_time,
                )
                .await
            {
                tracing::warn!(error = ?e, "failed to store taskprov opt in param");
            }

            Ok(task_config)
        }
    }
```

crates/daphne-server/src/roles/aggregator.rs
```diff
@@ -255,18 +255,23 @@ impl DapAggregator<DaphneAuth> for crate::App {
...
             };
+
+            let task_config = task_config.into_opted_in(&param);
+            let expiration_time = task_config.not_after;
+
             if let Err(e) = self
                 .kv()
                 .put_with_expiration::<kv::prefix::TaskprovOptInParam>(
                     task_id,
                     param.clone(),
```
mendess, line 266: this clone can probably be removed too, right?

mendess (approved):

cjpatton (approved):

PR #649, opened 2024-08-13
pine: Compute the squared norm just once

cjpatton:
Based on #648 (merge that first).
Partially addresses #618.

Split the existing FLP circuit into two:

1. One that computes the squared norm bound and checks that it is equal to the claimed value

2. Another for everything else, including checking that the claimed norm bound is in range and that the wraparound tests succeeded

The first circuit does not require joint randomness to compute, which means we can safely run it once without opening ourselves to offline attacks. It is also the most expensive part of the computation.

This change aligns our implementation with a planned change for the next version of PINE:
https://github.com/junyechen1996/draft-chen-cfrg-vdaf-pine/pull/92
https://github.com/junyechen1996/draft-chen-cfrg-vdaf-pine/pull/94


crates/daphne/src/pine/flp.rs
```diff
@@ -360,10 +354,8 @@ impl<F: FftFriendlyFieldElement> Type for PineType<F> {
     type Field = F;
 
     fn encode_measurement(&self, _measurement: &()) -> Result<Vec<F>, FlpError> {
-        // Pine can't use this feature of `Type` because the encoding takes an input called the
-        // "wraparound joint randomness" that is derived from secret shares of an intermediate
-        // representation of the input.
-        unreachable!()
+        // PINE uses a randomized encoding, where the randomness is derived by the protocol.
+        unimplemented!("use encode_with_wr_joint_rand() instead")
     }
```
mendess, line 359: if it can't implement this method than it shouldn't implement this trait :thinking: is there a way to rethink this?

crates/daphne/src/pine/flp.rs
```diff
@@ -486,6 +471,106 @@ impl<F: FftFriendlyFieldElement> Type for PineType<F> {
...
+        unimplemented!("use PineType to decode the aggregate result")
+    }
+
+    fn truncate(&self, _input: Vec<F>) -> Result<Vec<F>, FlpError> {
+        unimplemented!("use PineType to truncate the input")
+    }
+
+    fn output_len(&self) -> usize {
+        unimplemented!("use PineType to truncate the input")
+    }
```
mendess, line 502: Same thing here

crates/daphne/src/pine/flp.rs
cjpatton: This unfortunately is required by the API upstream in the `prio` crate: https://github.com/divviup/libprio-rs/issues/1108

mendess (approved):
I'm going to approve to unblock, but maybe prio should change to have more smaller traits to avoid situations like this? I'm not super familiar with prio but methods like
```rs
    fn encode_measurement(&self, _measurement: &()) -> Result<Vec<F>, FlpError>
    fn decode_result(&self, _data: &[F], _num_measurements: usize) -> Result<(), FlpError> 
    fn truncate(&self, _input: Vec<F>) -> Result<Vec<F>, FlpError>
    fn output_len(&self) -> usize
```
seem to be concerned with the domain of serialization(?) whereas the other methods of the trait seem concerned cryptography itself. Maybe these could be split into two traits?

crates/daphne/src/pine/flp.rs
```diff
@@ -486,6 +471,106 @@ impl<F: FftFriendlyFieldElement> Type for PineType<F> {
...
+
+        // Compute the squared norm of the gradient.
+        //
+        // This corresponds to the third call to `parallel_sum()` in the validity circuit of the
+        // reference implementation.
+        let sq_norm = encoded_gradient
+            .chunks(self.cfg.chunk_len_sq_norm_equal)
+            .map(|chunk| {
+                for x in chunk {
+                    buf.push(*x);
```
armfazh, line 530: double-check: push twice?

crates/daphne/src/pine/vdaf.rs
```diff
@@ -495,9 +517,16 @@ impl<F: FftFriendlyFieldElement> Aggregator<16, 16> for Pine<F> {
...
+            if !self.flp_sq_norm_equal.decide(v_sq_norm_equal)? {
+                return Err(VdafError::Uncategorized(
+                    "squared norm equality proof check failed".into(),
+                ));
+            }
+
+            for verifier in v.chunks(self.flp.verifier_len()) {
                 if !self.flp.decide(verifier)? {
-                    return Err(VdafError::Uncategorized("proof check failed".into()));
+                    return Err(VdafError::Uncategorized("main proof check failed".into()));
```
armfazh, line 529: it maybe good to refactor the 'main' proof and give it a more specific name.

armfazh (approved):

crates/daphne/src/pine/flp.rs
cjpatton: Yup, this is correct! `buf` is a sequence of inputs to the `Mul` gadget, which has two input wires and outputs the multiplication of the inputs. Here we're just using the `Mul` gadget to square each of the gradient entries.

crates/daphne/src/pine/vdaf.rs
cjpatton: Unfortunately the functional split between the two FLPs is a little awkward. The "main" circuit is everything but the squared norm equality check. In the future we might try to go for a split for which the two circuits would have more meaningful names: https://github.com/junyechen1996/draft-chen-cfrg-vdaf-pine/issues/95

PR #651, opened 2024-08-15
pine: Optimizations

cjpatton:
Partially addresses #618.
Based on #649 (merge that first).

This PR implements the following optimizations for PINE:
https://github.com/junyechen1996/draft-chen-cfrg-vdaf-pine/pull/96
https://github.com/junyechen1996/draft-chen-cfrg-vdaf-pine/pull/97


crates/daphne/src/pine/flp.rs
```diff
@@ -174,27 +174,54 @@ impl<F: FftFriendlyFieldElement> PineType<F> {
...
+        for (wr_test_result, rand_per_test) in wr_test_results
+            .iter_mut()
+            .zip(rand.chunks(rand_len_per_test))
+        {
+            for (gradient_chunk, r) in gradient.chunks(4).zip(rand_per_test) {
+                match r & 3 /* 0b11 */ {
+                    0 => *wr_test_result -= gradient_chunk[0],
+                    3 => *wr_test_result += gradient_chunk[0],
+                    _ => (),
+                };
```
armfazh, line 197: you can try this formula instead:

```rs
let n = (r & 0b11) //  n \in {0,1,2,3}
let a = gradient_chunk[0]
wr_test_result += a * ((n-1)>>1);
or
wr_test_result += a & ((n-1)>>1);

```

The match sentence can do a lot of comparisons. Verify whether this has some impact.

crates/daphne/src/pine/flp.rs
```diff
@@ -174,27 +174,54 @@ impl<F: FftFriendlyFieldElement> PineType<F> {
...
+
+                if gradient_chunk.len() < 2 {
+                    continue;
+                }
```
armfazh, line 201: if- sentences is what reduce preformance on programs, so what if you use [exact_chunks](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks_exact), so the loop has no branches, and only process the remainder bytes one at a time.

armfazh (changes requested):

crates/daphne/src/pine/flp.rs
cjpatton: I tacked on a commit with this change. Interestingly there is no change in performance. I suspect this is because the CPU is predicting that the `if` branches aren't taken?

crates/daphne/src/pine/flp.rs
cjpatton: I see what you're getting at, but I don't see how it works:
1. For the math two work, you need to convert `n` into the field, so that `-1` becomes `prime_modulus - 1`.
2. We can't do bit twiddling on the montgomery representation as far as I know

So what we want is something like (sorry about how verbose this is, we need to improve the field API in libprio):
```rust
a * (F::from(F::Integer::try_from(n as usize).unwrap()) - F::one()) * F::from(F::Integer::try_from(2).unwrap()).inv()
```

This costs two field multiplications per test and gradient element. Wouldn't this be more expensive than missing a branch prediction? 

Perhaps we can do the arithmetic in the integer representation of the field elements?

armfazh: Since the values are field elements, the bit tricks have to be expanded to field operations, which may result in overheads. 
Let's use continue using the current approach.

crates/daphne/src/pine/flp.rs
armfazh: Likely, however with this change we've removed the branch prediction logic.

crates/daphne/src/pine/flp.rs
```diff
@@ -174,28 +174,76 @@ impl<F: FftFriendlyFieldElement> PineType<F> {
...
+                };
+
+            if gradient_chunk.len() < 3 {
+                continue;
+            }
+            match r & 48 /* 0b110000 */ {
+                    0 => *wr_test_result -= gradient_chunk[2],
+                    48 => *wr_test_result += gradient_chunk[2],
+                    _ => (),
+                };
```
armfazh, line 246:
```suggestion
          for i,x in gradient_chunk.iter().enumerate() {
            match r & 3 /* 0b11 */ {
                    0 => *wr_test_result -= gradient_chunk[i],
                    3 => *wr_test_result += gradient_chunk[i],
                    _ => (),
                };
            r = r >> 2;
          }
```

armfazh (approved):
one minor nit, but other than that it looks good.

crates/daphne/src/pine/flp.rs
cjpatton: Done. I also added this loop to the `chunks_exact()` loop above: this doesn't seem to impact performance.

PR #655, opened 2024-08-22
Populate ProblemDetails::instance

mendess:
- Add task id to DapAbort::UnrecognizedTask
- Add report id to DapAbort::ReportToLate
- Make task id mandatory in DapAbort::InvalidMessage
- Make ProblemDetails::instance mandatory per RFC9457
- Merge DapError impl blocks

### DIFF
--- a/crates/daphne-server/src/roles/aggregator.rs
+++ b/crates/daphne-server/src/roles/aggregator.rs
@@ -88,7 +88,9 @@ impl DapAggregator<DaphneAuth> for crate::App {
         let task_config = self
             .get_task_config_for(task_id)
             .await?
-            .ok_or(DapError::Abort(DapAbort::UnrecognizedTask))?;
+            .ok_or(DapError::Abort(DapAbort::UnrecognizedTask {
+                task_id: *task_id,
+            }))?;
 
         let durable = self.durable();
         let mut requests = Vec::new();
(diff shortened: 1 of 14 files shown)


cjpatton (approved):
🔥 🚒

mendess:
> 🔥 🚒 

Do you think this is a good implementation of the instance field? I opted for what felt easier + useful. At the same time I feel like no will ever make use of this

cjpatton:
> > 🔥 🚒
> 
> Do you think this is a good implementation of the instance field? I opted for what felt easier + useful. At the same time I feel like no will ever make use of this

Hmm, actually the instance should be a URI: https://datatracker.ietf.org/doc/html/rfc7807#section-3.1

mendess:
> > > 🔥 🚒
> > 
> > 
> > Do you think this is a good implementation of the instance field? I opted for what felt easier + useful. At the same time I feel like no will ever make use of this
> 
> Hmm, actually the instance should be a URI: https://datatracker.ietf.org/doc/html/rfc7807#section-3.1

that RFC has been obsoleted by 9457 and in that one it allows for a opaque string with meaning for the server: https://datatracker.ietf.org/doc/html/rfc9457#section-3.1.5-3

Maybe I misunderstood and it should always be a URI. If it must be a URI what URI would I even use?

cjpatton:
> that RFC has been obsoleted by 9457 and in that one it allows for a opaque string with meaning for the server: https://datatracker.ietf.org/doc/html/rfc9457#section-3.1.5-3
> 
> Maybe I misunderstood and it should always be a URI. If it must be a URI what URI would I even use?

Nice catch,, that's the correct reference. My read of this text is that it wants a proper URI that an HTTP client can get in order to get more info about the error. But you're right that it doesn't have to be "dereferencable", which I take to mean if I try "GET /some-instance" I may not get a response.

I'd suggest reformatting the strings so that they could be interpreted as relative URIs.

PR #657, opened 2024-08-23
vdaf: Replace Pine128 with Pine64HmacSha256Aes128

cjpatton:
Partially addresses #618.
Based on #656.

The latter variant will be deployed first.

### DIFF
--- a/crates/daphne/src/vdaf/mod.rs
+++ b/crates/daphne/src/vdaf/mod.rs
@@ -207,7 +207,7 @@ pub enum VdafPrepState {
         out_share: Vec<Field64>,
     },
     #[cfg(feature = "experimental")]
-    Pine128(PinePrepState<Field128, 16>),
+    Pine64HmacSha256Aes128(PinePrepState<Field64, 32>),
 }
 
 #[cfg(any(test, feature = "test-utils"))]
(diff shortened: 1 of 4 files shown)


crates/daphne/src/vdaf/mod.rs
```diff
@@ -277,7 +277,7 @@ pub enum VdafPrepMessage {
     #[cfg(feature = "experimental")]
     MasticShare(Field64),
     #[cfg(feature = "experimental")]
-    Pine128Share(crate::pine::msg::PrepShare<Field128, 16>),
+    Pined64HmacSha256Aes128(crate::pine::msg::PrepShare<Field64, 32>),
```
armfazh, line 280: 32 is the seed size? out of curiosity, why a smaller field uses a larger seed size?

armfazh (approved):

crates/daphne/src/vdaf/mod.rs
cjpatton: The two are unrelated:

* Pine128 uses XofTurboShaek128, which at the moment has a 16-byte seed
* Pine64HmacSha256Aes128 uses XofHmacSha256Aes128 (a non-standard XOF implemented in this repo), which has a 32 bytes eed.

crates/daphne/src/vdaf/pine.rs
```diff
@@ -164,23 +172,25 @@ impl PineConfig {
 
         match (var, host_state, host_share) {
             (
-                PineVariant::Field128,
-                VdafPrepState::Pine128(state),
-                VdafPrepMessage::Pine128Share(share),
+                PineVariant::Field64HmacSha256Aes128,
+                VdafPrepState::Pine64HmacSha256Aes128(state),
+                VdafPrepMessage::Pined64HmacSha256Aes128(share),
```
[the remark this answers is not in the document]
cjpatton: Nice catch!

PR #659, opened 2024-08-26
Rewrite workers to use the axum router

mendess:
the `workers` library supports using types from the `http` library instead of its own reimplementation of these. This means we can use axum to do the routing for us.

This comes with a few advantages, one of which is proper use of middleware to implement authorization and timings.

### DIFF
--- a/crates/daphne-worker/Cargo.toml
+++ b/crates/daphne-worker/Cargo.toml
@@ -18,12 +18,11 @@ crate-type = ["cdylib", "rlib"]
 [dependencies]
 axum.workspace = true
 axum-extra = { workspace = true, features = ["typed-header"] }
-axum-macros = "0.4.1"
 # TODO(mendess): delete
 bincode.workspace = true
 bytes.workspace = true
 chrono = { workspace = true, default-features = false, features = ["clock", "wasmbind"] }
-constcat = "0.5.0"
+constcat.workspace = true
 daphne = { path = "../daphne", features = ["prometheus"] }
 futures = { workspace = true, optional = true }
 getrandom = { workspace = true, features = ["js"] }
@@ -43,6 +42,7 @@ tracing-core.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter", "json"]}
 tracing.workspace = true
 url.workspace = true
+wasm-streams.workspace = true
 worker.workspace = true
 tower-service.workspace = true
 
--- a/crates/daphne-worker/Cargo.toml
+++ b/crates/daphne-worker/Cargo.toml
@@ -16,13 +16,21 @@ description = "Workers backend for Daphne"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
+axum.workspace = true
+axum-extra = { workspace = true, features = ["typed-header"] }
+axum-macros = "0.4.1"
 # TODO(mendess): delete
 bincode.workspace = true
+bytes.workspace = true
 chrono = { workspace = true, default-features = false, features = ["clock", "wasmbind"] }
+constcat = "0.5.0"
 daphne = { path = "../daphne", features = ["prometheus"] }
 futures = { workspace = true, optional = true }
 getrandom = { workspace = true, features = ["js"] }
+headers.workspace = true
 hex.workspace = true
+http-body-util.workspace = true
+http.workspace = true
 prio.workspace = true
 prometheus.workspace = true
 rand.workspace = true
@@ -36,6 +44,7 @@ tracing-subscriber = { workspace = true, features = ["env-filter", "json"]}
 tracing.workspace = true
 url.workspace = true
 worker.workspace = true
+tower-service.workspace = true
 
 [dependencies.daphne-service-utils]
 path = "../daphne-service-utils"
--- a/crates/daphne-worker/src/storage_proxy/middleware.rs
+++ b/crates/daphne-worker/src/storage_proxy/middleware.rs
@@ -21,8 +21,8 @@ use tower_service::Service;
 
 use super::RequestContext;
 
-/// Check if the request's authorization. If unauthorized, return the reason why.
-pub async fn unauthorized_reason(
+/// Performs bearer token auth of a request.
+pub async fn bearer_auth(
     ctx: State<Arc<RequestContext>>,
     bearer: TypedHeader<Authorization<Bearer>>,
     request: axum::extract::Request,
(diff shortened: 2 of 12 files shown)


crates/daphne-worker/Cargo.toml
```diff
@@ -16,13 +16,21 @@ description = "Workers backend for Daphne"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
+axum.workspace = true
+axum-extra = { workspace = true, features = ["typed-header"] }
+axum-macros = "0.4.1"
```
cjpatton, line 21: Move this to the the `Cargo.toml` for the workspace? Here and below.

crates/daphne-worker/src/storage_proxy/middleware.rs
```diff
@@ -0,0 +1,116 @@
...
+            "Authorization token for storage proxy is not configured",
+        )
+            .into_response();
+    };
+
+    if !constant_time_eq(bearer.token().as_bytes(), trusted_token.as_bytes()) {
+        return (StatusCode::UNAUTHORIZED, "Incorrect authorization token").into_response();
+    }
+
+    next.call(request.map(axum::body::Body::new)).await.unwrap()
```
cjpatton, line 51: Why is panicking safe here?

crates/daphne-worker/src/storage_proxy/middleware.rs
```diff
@@ -0,0 +1,116 @@
...
+    }) else {
+        tracing::warn!("trusted bearer token not configured");
+        return (
+            StatusCode::INTERNAL_SERVER_ERROR,
+            "Authorization token for storage proxy is not configured",
+        )
+            .into_response();
+    };
+
+    if !constant_time_eq(bearer.token().as_bytes(), trusted_token.as_bytes()) {
```
cjpatton, line 47: Potential attack: If the request has no bearer token, i.e., `bearer.token().as_bytes().is_some()`, will this equality hold?

crates/daphne-worker/src/storage_proxy/middleware.rs
```diff
@@ -0,0 +1,116 @@
...
+    headers::{authorization::Bearer, Authorization},
+    TypedHeader,
+};
+use daphne::messages::constant_time_eq;
+use http::{Method, StatusCode};
+use tower_service::Service;
+
+use super::RequestContext;
+
+/// Check if the request's authorization. If unauthorized, return the reason why.
```
cjpatton, line 24: I don't think this comment is accurate anymore.

crates/daphne-worker/src/storage_proxy/mod.rs
```diff
@@ -193,204 +202,208 @@ async fn storage_purge(ctx: &RequestContext<'_>) -> worker::Result<Response> {
...
+            .map(Self)
+            .ok_or_else(headers::Error::invalid)
+    }
+}
 
-            let elapsed = elapsed(&start);
+impl ExpirationHeader {
+    fn at_least_60s_in_the_future(self) -> Self {
+        // KV wont let you request an expiration that isn't at least 60 seconds into the
+        // future. If you try to do so, it will return a 400. Problem is, the only error
```
cjpatton, line 242:
```suggestion
        // future. If you try to do so, it will return a 400. The problem is, the only error
```

crates/daphne-worker/src/storage_proxy/mod.rs
```diff
@@ -193,204 +202,208 @@ async fn storage_purge(ctx: &RequestContext<'_>) -> worker::Result<Response> {
...
+            .and_then(|v| v.to_str().ok())
+            .and_then(|v| v.parse().ok())
+            .map(Self)
+            .ok_or_else(headers::Error::invalid)
+    }
+}
 
-            let elapsed = elapsed(&start);
+impl ExpirationHeader {
+    fn at_least_60s_in_the_future(self) -> Self {
```
cjpatton, line 240: A little more rusty, plus more to the point
```suggestion
    fn with_kv_minimum_time_out(self) -> Self {
```

crates/daphne-worker/src/storage_proxy/mod.rs
```diff
@@ -193,204 +202,208 @@ async fn storage_purge(ctx: &RequestContext<'_>) -> worker::Result<Response> {
...
+        // KV wont let you request an expiration that isn't at least 60 seconds into the
+        // future. If you try to do so, it will return a 400. Problem is, the only error
+        // the worker API returns is a JsValue that might contain a string that might
+        // explain that.
+        //
+        // In order to avoid parsing the error message we opt to just "hardcode" the
+        // expiration to be, at least 65 seconds from now. Effectively expiring the value as soon
+        // as possible. The extra 5 seconds are just in case we take a really long time from here to
+        // the request.
+        let now_plus_65_seconds = (Date::now().as_millis() / 1000) + 65;
```
cjpatton, line 250: Can you get `as_secs()`?

crates/daphne-worker/src/storage_proxy/middleware.rs
mendess: it is safe, but I made it more explicit. This will look better by the end of the year

crates/daphne-worker/src/storage_proxy/middleware.rs
mendess: If the bearer token doesn't exist, this function isn't called at all

crates/daphne-worker/src/storage_proxy/mod.rs
mendess: `with_` feels like you're adding something to it, whereas my name I think conveys the idea that it's making a guarantee, almost like an assertion

crates/daphne-worker/Cargo.toml
mendess: this can actually be deleted, it's a tool to help debug stuff

cjpatton (commented):
One comment left, otherwise LGTM

crates/daphne-worker/src/storage_proxy/middleware.rs
mendess: This is not assumption, it's the contract of `TypedHeader` that it can only be instantiated if the header passed to it exists, there is no `Option` here.

This makes use of [extractors](https://docs.rs/axum/latest/axum/#extractors) feature of axum, and makes use of the [TypedHeader](https://docs.rs/axum-extra/latest/axum_extra/struct.TypedHeader.html) type.

cjpatton (approved):

PR #662, opened 2024-08-27
Add support for SSLKEYLOGFILE in dapf

mendess:


crates/dapf/src/acceptance/mod.rs
```diff
@@ -166,6 +166,19 @@ pub enum HttpClient {
 
 impl HttpClient {
     fn reused(leader_tls_identity: Option<Identity>) -> Self {
```
cjpatton, line 168: SSLKEYLOGFILE should not be enabled by default, as it is not secure. The user should have to set it with a flag.

cjpatton (changes requested):

crates/dapf/src/acceptance/mod.rs
mendess: even if it's only used from a dapf? I guess we run dapf on the edge. I'll add that flag

crates/dapf/src/http_client.rs
```diff
@@ -0,0 +1,178 @@
...
+    inner: HttpClientInner,
+}
+
+#[allow(clippy::large_enum_variant)]
+enum HttpClientInner {
+    NoReuse {
+        identity: Option<Identity>,
+        ssl_key_log_file_enabled_tls: Option<rustls::ClientConfig>,
+    },
+    Reuse(Client),
```
cjpatton, line 28: A comment here explaining what we mean by "reuse" would be helpful.

crates/dapf/src/http_client.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+fn load_identity() -> anyhow::Result<Option<Identity>> {
+    // LEADER_ variables left here for backwards compatibility
+    const TLS_CLIENT_CERT_VAR: [&str; 2] = ["LEADER_TLS_CLIENT_CERT", "TLS_CLIENT_CERT"];
+    const TLS_CLIENT_KEY_VAR: [&str; 2] = ["LEADER_TLS_CLIENT_KEY", "TLS_CLIENT_KEY"];
+
+    let leader_tls_client_cert = TLS_CLIENT_CERT_VAR
+        .iter()
+        .find_map(|key| env::var(key).ok());
```
cjpatton, line 38: nit: Probably better to check for the non-legacy variable first, i.e., "TLS_CLIENT_CERT" should be preferred over "LEADER_TLS_CLIENT_CERT" if both exist

crates/dapf/src/http_client.rs
```diff
@@ -0,0 +1,178 @@
...
+        http_client_builder = http_client_builder.identity(identity);
+    }
+
+    http_client_builder
+        .build()
+        .expect("failed to build http client")
+}
+
+impl HttpClient {
+    pub fn reused(enable_ssl_key_log_file: bool) -> anyhow::Result<Self> {
```
cjpatton, line 99: ?
```suggestion
    pub fn reuse(enable_ssl_key_log_file: bool) -> anyhow::Result<Self> {
```

cjpatton (approved):

cjpatton (approved):

PR #663, opened 2024-08-27
Reorganize dapf

mendess:
A bit of house keeping, I think dapf has grown enough that separating its
different actions into categories makes sense.


cjpatton (approved):
SGTM.

PR #668, opened 2024-08-29
taskprov: Opt out of some PINE parameter ranges

cjpatton:
Partially addresses #618.
Based on #667 (merge that first).

Security considerations:

* For robustness, the number of successful wraparound tests shouldn't be too small: see [1].

* For robustness, the number of proofs for the main circuit should not be lower than 2 for Field64 and 5 for Field32. Again, this may end up being too low [1].

Also, operationally we want to prevent ourselves from doing too much work, so pick reasonable upper bounds for the dimension and number of wraparound tests.

[1] https://github.com/junyechen1996/draft-chen-cfrg-vdaf-pine/issues/39

### DIFF
--- a/crates/daphne/src/taskprov.rs
+++ b/crates/daphne/src/taskprov.rs
@@ -7,8 +7,6 @@
 
 use std::num::NonZeroUsize;
 
-#[cfg(feature = "experimental")]
-use crate::vdaf::pine::{pine32_hmac_sha256_aes128, pine64_hmac_sha256_aes128, PineConfig};
 use crate::{
     fatal_error,
     hpke::HpkeConfig,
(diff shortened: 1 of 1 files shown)


crates/daphne/src/taskprov.rs
```diff
@@ -206,9 +239,13 @@ impl VdafConfig {
                     num_proofs,
                 },
             ) => {
-                const MAX_PROOFS: u8 = 3;
-                if num_proofs == 0 || num_proofs > MAX_PROOFS {
-                    return Err(DapAbort::InvalidTask { detail: format!("invalid number of proofs for Prio3SumVecField64MultiproofHmacSha256Aes128: got {num_proofs}; expected a value in range [0, {MAX_PROOFS}]"), task_id: *task_id });
+                if !(1..PRIO3_MAX_PROOFS + 1).contains(&num_proofs) {
```
cjpatton, line 242: Reviewer note: previously we accepted 0 proofs here, but this doesn't make much sense.

mendess: you can use `1..=PRIO3_MAX_PROOFS` instead of `+1` to have an inclusive range

mendess (approved):

crates/daphne/src/taskprov.rs
cjpatton: Done.

armfazh (approved):

PR #674, opened 2024-09-09
pine: Add support for multiple squared norm equality proofs

cjpatton:
Partially addresses #618.

**WIP: Need to regenerate some test vectors. Plus, I'm going to refactor to avoid duplicating `cfg`. **

As specified in:
https://github.com/junyechen1996/draft-chen-cfrg-vdaf-pine/pull/100

Update the taskprov opt-in logic to accept as few as `1` and as many as `MAX_PROOFS`.


crates/daphne/src/pine/flp.rs
```diff
@@ -56,7 +58,22 @@ use super::{chunk_count, f64_to_field, field_to_f64, PineConfig, USAGE_WR_JOINT_
...
-    pub(crate) cfg: PineConfig<F>,
+    // PINE parameters
+    pub(crate) dimension: usize,
+    pub(crate) frac_bits: usize,
+    pub(crate) chunk_len: usize,
+    pub(crate) num_wr_tests: usize,
+    pub(crate) num_wr_successes: usize,
+
+    // FLP parameters
+    pub(crate) sq_norm_bound: F,
```
armfazh, line 69: is it possible to store the bounds as integers and internally convert them to field elements?
Thus, the 'PineType' struct is not attached to any field in particular.

crates/daphne/src/pine/vdaf.rs
```diff
@@ -452,38 +463,43 @@ impl<F: FftFriendlyFieldElement, X: Xof<SEED_SIZE>, const SEED_SIZE: usize>
...
 
-            verifiers_share.append(&mut self.flp_sq_norm_equal.query(
-                &meas_share,
-                ps_sq_norm_equal,
-                qr_sq_norm_equal,
-                &[],
-                2,
-            )?);
+            for (proof_share, query_rand) in ps_sq_norm_equal
+                .chunks(self.flp_sq_norm_equal.proof_len())
```
armfazh, line 491: if the vector splits evenly, it's better to use `chunks_exact`

armfazh (approved):
a couple of nits only

crates/daphne/src/pine/flp.rs
cjpatton: Well, the type of interger would be `F::Integer`: for Field64 this is u64; for Field128 this would be u128. So we'd still have a generic parameter no matter what unfortunately.

Another consideration here: we use this in multiple parts of the code. If we stored it as an integer, we'd have to convert to a field element on the fly, which has some cost.

crates/daphne/src/pine/vdaf.rs
cjpatton: Nice, that's possible basically for every call to `chunks()` in this file. I've tacked on a commit.

PR #675, opened 2024-09-10
Rewrite benchmarks and other small refactors

mendess:
- Rewrite the aggregation benchmark
- Stop using a global variable to control replay protection in tests and dapf
- Make more fields in InMemoryAggregator private
- Fix dapf http client
- Add dhat profiling to helper
- Add benchmark results

### DIFF
--- a/crates/daphne/src/protocol/aggregator.rs
+++ b/crates/daphne/src/protocol/aggregator.rs
@@ -417,28 +417,52 @@ impl DapTaskConfig {
     where
         S: Stream<Item = Report>,
     {
+        self.produce_agg_job_req_impl(
+            decrypter,
+            initializer,
+            task_id,
+            part_batch_sel,
+            agg_param,
+            reports,
+            metrics,
+            ReplayProtection::Enabled,
+        )
+        .await
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    async fn produce_agg_job_req_impl<S>(
+        &self,
+        decrypter: &impl HpkeDecrypter,
+        initializer: &impl DapReportInitializer,
+        task_id: &TaskId,
+        part_batch_sel: &PartialBatchSelector,
+        agg_param: &DapAggregationParam,
+        reports: S,
+        metrics: &dyn DaphneMetrics,
+        replay_protection: ReplayProtection,
+    ) -> Result<(DapAggregationJobState, AggregationJobInitReq), DapError>
+    where
+        S: Stream<Item = Report>,
+    {
         let (report_count_hint, _upper_bound) = reports.size_hint();
         let mut consumed_reports = Vec::with_capacity(report_count_hint);
         let mut helper_shares = Vec::with_capacity(report_count_hint);
         {
-            let mut processed = HashSet::with_capacity(report_count_hint);
+            let mut processed = replay_protection
+                .enabled()
+                .then(|| HashSet::with_capacity(report_count_hint));
             let mut reports = pin!(reports);
             while let Some(report) = reports.next().await {
-                let enforce_replay_check = {
-                    #[cfg(feature = "report-generator")]
-                    {
-                        !crate::testing::report_generator::generator_replaying_reports()
+                if let Some(processed) = &mut processed {
+                    if processed.contains(&report.report_metadata.id) {
+                        return Err(fatal_error!(
+                            err = "tried to process report sequence with non-unique report IDs",
+                            non_unique_id = %report.report_metadata.id,
+                        ));
                     }
-                    #[cfg(not(feature = "report-generator"))]
-                    true
-                };
-                if enforce_replay_check && processed.contains(&report.report_metadata.id) {
-                    return Err(fatal_error!(
-                        err = "tried to process report sequence with non-unique report IDs",
-                        non_unique_id = %report.report_metadata.id,
-                    ));
+                    processed.insert(report.report_metadata.id);
                 }
-                processed.insert(report.report_metadata.id);
 
                 let [leader_share, helper_share] = report.encrypted_input_shares;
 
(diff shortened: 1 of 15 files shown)


crates/daphne/src/protocol/aggregator.rs
```diff
@@ -421,24 +422,20 @@ impl DapTaskConfig {
...
+                .then(|| HashSet::with_capacity(report_count_hint));
             let mut reports = pin!(reports);
             while let Some(report) = reports.next().await {
-                let enforce_replay_check = {
-                    #[cfg(feature = "report-generator")]
-                    {
-                        !crate::testing::report_generator::generator_replaying_reports()
+                if let Some(processed) = &mut processed {
+                    if processed.contains(&report.report_metadata.id) {
+                        return Err(fatal_error!(
```
cjpatton, line 432: I'm not sure when this happened, but this shouldn't be a fatal error. Rather, it should be a DAP protocol abort triggered by a non-compliant leader. Would you mind fixing this here?

cjpatton (changes requested):
Hmm, I'm not sure if I reviewed the change that introduced the replay check exception logic, but I think there is a problem with it. See inline comments.

crates/daphne/src/protocol/aggregator.rs
```diff
@@ -412,6 +412,7 @@ impl DapTaskConfig {
         part_batch_sel: &PartialBatchSelector,
         agg_param: &DapAggregationParam,
         reports: S,
+        replay_protection: ReplayProtection,
```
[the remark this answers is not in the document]
mendess: I thought about adding it to task config but I decide not to because I I'm wary of overloading the role of the task config. This is supposed to be more of an ad-hoc parameter passed in tests.

On the point of production, this is used by the leader only, which is not in production, but yes, it could be used in a production leader but I expect the code will change again by then. Either way changing the parameter name will not change anything in the call site so I don't think it improves readability. We could make this parameter available in tests only though :thinking:

crates/daphne/src/protocol/aggregator.rs
cjpatton: Ohhh, yes you're right, if this is called by the Leader than it can only be a fatal error. (Arguably this is an issue with the protocol.)

crates/daphne/src/protocol/aggregator.rs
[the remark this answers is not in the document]
mendess: fenced behind the feature flag. Maybe the feature flag itself should be "insecure-test-utils" :thinking:

crates/daphne/src/protocol/aggregator.rs
```diff
@@ -414,31 +414,55 @@ impl DapTaskConfig {
...
+    async fn produce_agg_job_req_impl<S>(
+        &self,
+        decrypter: &impl HpkeDecrypter,
+        initializer: &impl DapReportInitializer,
+        task_id: &TaskId,
+        part_batch_sel: &PartialBatchSelector,
+        agg_param: &DapAggregationParam,
+        reports: S,
+        metrics: &dyn DaphneMetrics,
+        replay_protection: ReplayProtection,
```
cjpatton, line 443: I still don't think this is a good idea. Adding the word "insecure" into the code makes it clear to someone reading this function what secure behavior looks like. I would go with `inesecure_disable_replay_protection: bool`: if you don't like this, then see my suggestion below.

crates/daphne/src/protocol/aggregator.rs
```diff
@@ -529,6 +553,35 @@ impl DapTaskConfig {
         ))
     }
 
+    #[allow(clippy::too_many_arguments)]
+    #[cfg(any(test, feature = "test-utils"))]
+    pub async fn produce_agg_job_req_allowing_replayed_reports<S>(
```
cjpatton, line 558:
```suggestion
    pub async fn produce_agg_job_req_insecure_disable_replay_portection<S>(
```

crates/daphne/src/protocol/aggregator.rs
```diff
@@ -383,7 +383,7 @@ pub(crate) enum ReportProcessedStatus {
     Rejected(TransitionFailure),
 }
 
-#[derive(Default, Debug)]
+#[derive(Default, Debug, Clone, Copy)]
 pub enum ReplayProtection {
     #[default]
     Enabled,
```
cjpatton, line 389: Change the variant "Disabled" to "InsecureDisabled". TBH it would be nice if we could put this whole enum behind "test-utils".

cjpatton (changes requested):

crates/daphne/src/protocol/aggregator.rs
cjpatton: EDIT: my suggestion spelled "protection" wrong 😮

mendess: what about `test_produce_agg_job_req`? The function itself is not insecure but passing `ReplayProtection::InsecureDisabled` is

crates/daphne/src/protocol/aggregator.rs
mendess: We can't put this enum behind test utils for as long as we have an override to disabled replay protection in production. After that's gone we can do this

crates/daphne/src/protocol/aggregator.rs
mendess: I don't like to use `bool` parameters, they are impossible to read from calling code.
```rs
produce_agg_job_req(dec, init, param, reports, metrics, false); // what does false mean? Is this the correct way?
```

instead I'd rather keep the parameter name as is and take the suggestion to rename Disabled to InsecureDisabled. Using that variant is insecure, calling functions that take in this enum is not

crates/daphne/src/protocol/aggregator.rs
cjpatton: How about removing the replay_protection parameter from this function? Then in the test code you'd pick which one to call based on whether you want to disable.

mendess: I tried that, it made the test a lot more verbose (having to repeat all 8 parameters) for no real gain imo

cjpatton: Alright, "test_produce_agg_job_req" is fine. I don't think we need to bikeshed this more.

crates/daphne/src/testing/report_generator.rs
```diff
@@ -75,6 +67,7 @@ impl ReportGenerator {
         version: DapVersion,
         now: Time,
         extensions: Vec<messages::Extension>,
+        replay_reports: bool,
```
cjpatton, line 70: Does this mean "replay reports" or "disable replay protection"? If the latter, then change the type to `ReplayProtection`.

cjpatton (approved):
Just one more nit.

crates/daphne/src/testing/report_generator.rs
mendess: it means "replay reports"

PR #682, opened 2024-09-18
Move authentication logic out of the daphne crate

mendess:


crates/daphne-server/examples/configuration-helper.toml
```diff
@@ -21,6 +21,9 @@ base_url = "http://127.0.0.1:8788"
 allow_taskprov = true
 default_num_agg_span_shards = 4
 
+[service.peer_auth.helper]
+leader_token = "I-am-the-leader" # SECRET
```
cjpatton, line 25: This doesn't have any risk of being deployed.
Likewise below.

```suggestion
leader_token = "I-am-the-leader"
```

crates/daphne-server/src/lib.rs
```diff
@@ -84,6 +94,8 @@ pub struct App {
     /// colleciton requests. Note that in a production Leader, it is necessary to store this state
     /// across requsets.
     test_leader_state: Arc<Mutex<InMemoryLeaderState>>,
+    #[cfg(feature = "test-utils")]
+    interop_bearer_tokens: crate::roles::InteropBearerTokens,
```
cjpatton, line 98: What do you mean by "interop" bearer token? Note that the type has no documentation and I can't tell what this is from context.

crates/daphne-server/src/lib.rs
```diff
@@ -100,6 +112,38 @@ impl router::DaphneService for App {
     fn signing_key(&self) -> Option<&p256::ecdsa::SigningKey> {
         self.service_config.signing_key.as_ref()
     }
+
+    fn check_bearer_token(
+        &self,
+        token: &BearerToken,
+        sender: DapSender,
+        task_id: Option<TaskId>,
+    ) -> Result<(), String> {
```
cjpatton, line 121: This return type looks strange to me. I guess`Ok(())` means the token is good and `Err(s)` means it's bad? I'd suggest documenting the behavior of this function.

crates/daphne-server/src/lib.rs
```diff
@@ -100,6 +112,38 @@ impl router::DaphneService for App {
...
+            (_, sender, Some(task_id)) => {
+                if self.interop_bearer_tokens.matches(sender, task_id, token) {
+                    Ok(())
+                } else {
+                    Err(format!(
+                        "the indicated bearer token is incorrect for the {sender:?} with task_id {}",
+                        daphne::messages::Base64Encode::to_base64url(&task_id),
+                    ))
+                }
+            }
```
cjpatton, line 141: So the fact this code is behind "test-utils" gives me a hint: by "interop bearer token" you perhaps mean the set of bearer tokens that would have been configured via https://divergentdave.github.io/draft-dcook-ppm-dap-interop-test-design/draft-dcook-ppm-dap-interop-test-design.html. Is that right?

If so, I think you're missing an important detail. The original vision was that each task would be associated with a method or methods for authorization, e.g., the bearer token for the leader. I don't want us to lose cite of this vision. If we're changing this part of the code, I think it would be a good time to make sure what we build can be made multitenant.

cjpatton (commented):
Decoupling authorization logic from the daphne crate is a good idea. But if we're taking the time to rework this code, I think we should try to think ahead to how we'd build a multitenant service.

crates/daphne-server/examples/configuration-helper.toml
mendess: I thought this comment served as a kind of documentation to show that this value is supposed to be a secret in a production environment

mendess:
> Decoupling authorization logic from the daphne crate is a good idea. But if we're taking the time to rework this code, I think we should try to think ahead to how we'd build a multitenant service.

designing for multi-tenancy is not an easy task and depends a lot on how the system is deployed. I believe the best way to design for multi-tenancy in this case is to not design at all. Our internal implementation already has some semblance of multi tenancy going on and it doesn't depend on this implementation to implement it

crates/daphne-server/src/lib.rs
mendess: it is indeed to support the interop draft (https://divergentdave.github.io/draft-dcook-ppm-dap-interop-test-design/draft-dcook-ppm-dap-interop-test-design.html), will document

crates/daphne-server/src/lib.rs
mendess: what do you mean by "each task would be associated with a method or methods for authorization, e.g., the bearer token for the leader"? The way I wrote the code each task has an associated bearer token, that's why `matches` takes a `task_id`

crates/daphne-server/examples/configuration-helper.toml
cjpatton: That sounds like a reasonable interpretation. In that case, this comment should also be added below:
```
[service.leader_to_helper_tokens]
helper = "I-am-the-leader"
```

crates/daphne-server/src/lib.rs
```diff
@@ -100,6 +112,45 @@ impl router::DaphneService for App {
...
         self.service_config.signing_key.as_ref()
     }
+
+    async fn check_bearer_token(
+        &self,
+        token: &BearerToken,
+        sender: DapSender,
+        task_id: Option<TaskId>,
+    ) -> Result<(), String> {
+        match (&self.service_config.peer_auth, sender, task_id) {
```
cjpatton, line 122: This `service_config.peer_auth` is specific to taskprov, correct? If so, it should be named accordingly.

crates/daphne-server/src/roles/leader.rs
```diff
@@ -165,16 +140,30 @@ impl crate::App {
...
+            .get(&url.host().unwrap().to_owned())
+        {
+            headers.insert(
+                HeaderName::from_static(http_headers::DAP_AUTH_TOKEN),
+                HeaderValue::from_str(auth.as_str()).map_err(
+                    |e| fatal_error!(err = ?e, "failed to construct authentication header"),
+                )?,
+            );
+        }
+        #[cfg(feature = "test-utils")]
```
cjpatton, line 155: Is this necessary?
```suggestion
```

crates/daphne-server/src/roles/leader.rs
```diff
@@ -165,16 +140,30 @@ impl crate::App {
                 .map_err(|e| fatal_error!(err = ?e, "failed to construct content-type header"))?,
         );
 
-        if let Some(bearer_token) = req.sender_auth.and_then(|auth| auth.bearer_token) {
+        if let Some(auth) = self
+            .service_config
+            .leader_to_helper_tokens
+            .get(&url.host().unwrap().to_owned())
```
cjpatton, line 146: I don't think this unwrap is safe. From the documentation:
```
    /// Return the parsed representation of the host for this URL.
    /// Non-ASCII domain labels are punycode-encoded per IDNA if this is the host
    /// of a special URL, or percent encoded for non-special URLs.
    ///
    /// Cannot-be-a-base URLs (typical of `data:` and `mailto:`) and some `file:` URLs
    /// don’t have a host.
```

cjpatton: We're unlikely to panic here, but it seems like a good idea to handle as a fatal error.

crates/daphne-server/src/roles/leader.rs
```diff
@@ -165,16 +140,30 @@ impl crate::App {
                 .map_err(|e| fatal_error!(err = ?e, "failed to construct content-type header"))?,
         );
 
-        if let Some(bearer_token) = req.sender_auth.and_then(|auth| auth.bearer_token) {
+        if let Some(auth) = self
+            .service_config
+            .leader_to_helper_tokens
```
cjpatton, line 145: `leader_to_helper_tokens`: Again, this is specific to taskprov, correct?

crates/daphne-server/src/roles/leader.rs
```diff
@@ -165,16 +140,30 @@ impl crate::App {
...
+        }
+        #[cfg(feature = "test-utils")]
+        if let Some(auth) = self
+            .bearer_tokens()
+            .get(daphne::DapSender::Leader, task_id)
+            .await
+            .map_err(|e| fatal_error!(err = ?e, "failed to get leader bearer token"))?
+        {
             headers.insert(
                 HeaderName::from_static(http_headers::DAP_AUTH_TOKEN),
```
cjpatton, line 163: Hmm, if we set this header in the previous `if let Some(auth) = ...` then we'll overwrite it here. I don't think that's what we want to do.

crates/daphne-server/src/lib.rs
mendess: I did document, but I documented in the trait and not in the implementation

crates/daphne-server/src/lib.rs
mendess: it is not specific to taskprov no. This is just a bearer token we accept

crates/daphne-server/src/roles/leader.rs
mendess: I think it's okay to panic here, this url should originate from trusted input (configuration) and not from external input

crates/daphne-server/src/roles/leader.rs
mendess: No, this has nothing to do with taskprov, this is just for communication between the leader and the helper

crates/daphne-server/src/roles/leader.rs
mendess: No, this was wrong

crates/daphne-server/src/lib.rs
```diff
@@ -100,6 +112,45 @@ impl router::DaphneService for App {
...
+            }
+            (_, sender, Some(task_id)) => {
+                if self
+                    .bearer_tokens()
+                    .matches(sender, task_id, token)
+                    .await
+                    .map_err(|error| {
+                        tracing::error!(?error, "failed to fetch bearer token from kv");
+                        "internal error occured while running authentication".to_string()
+                    })?
```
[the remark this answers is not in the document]
mendess: You're right. I was being lazy here. I've changed it to return two different errors. I decided to include the `either` crate for this. I could abuse result but I didn't want to overload it's meaning

mendess:
One of the intentions of this PR is to distance taskprov from authentication. There are two methods by which an aggregator can determine if the request it receives is authorized:
- static: defined in configuration, this allows for just one token (see [peer_auth][peer-auth])
- dynamic: stored in kv mapping (Role, TaskId) pairs to a token (see [bearer_tokens][bearer-tokens])

[peer-auth]: https://github.com/cloudflare/daphne/blob/62036ff253606d661ad35e04c54ea28f1e50f90d/crates/daphne-service-utils/src/config.rs#L58
[bearer-tokens]: https://github.com/cloudflare/daphne/blob/62036ff253606d661ad35e04c54ea28f1e50f90d/crates/daphne-server/src/roles/mod.rs#L42

crates/daphne-server/src/roles/leader.rs
cjpatton: IMO we shouldn't treat configuration as valid input as a general matter. In this particular case, I think the `Url` parser would be happy to consume a URL that doesn't have a host.

crates/daphne-server/src/lib.rs
cjpatton: I believe this is necessary because it avoids granting the "taskprov peer" access to tasks it hasn't configured.

crates/daphne-server/src/roles/leader.rs
cjpatton: Question: are there any conditions under which we expect the peer_auth token to be the same as one of the per-task tokens? I suspect not, but perhaps you're thinking further ahead than I am.

If this is not expected, I think it would be best to treat it as a fatal error.

crates/daphne-server/src/router/mod.rs
```diff
@@ -369,7 +371,12 @@ where
                 .server_metrics()
                 .auth_method_inc(metrics::AuthMethod::BearerToken);
             state
-                .check_bearer_token(&token, sender, request.task_id)
+                .check_bearer_token(
+                    &token,
+                    sender,
+                    request.task_id.unwrap(), // task ids are mandatory
+                    request.taskprov.is_some(),
```
mendess, line 378: Is this enough to determine if the task is using taskprov? It's different from how we used to test this

cjpatton: We should also consult the `DapGlobalConfig` to see if we've enabled taskprov support (`allow_taskprov`). If we haven't, then the advertisement in the request will be ignored.

mendess: I'm doing that inside the `check_bearer_token` method

cjpatton: Sounds good. Maybe make a note here if you think it's not clear from the control flow. (I haven't looked.)

crates/daphne-server/src/router/mod.rs
```diff
@@ -246,33 +273,9 @@ where
...
 
         let extract_header_as_string = |header: &'static str| -> Option<String> {
-            parts
-                .headers
-                .get(header)?
-                .to_str()
-                .ok()
-                .map(ToString::to_string)
+            extract_header_as_string(&parts.headers, header)
         };
```
cjpatton, line 277: Let's just completely get rid of this closure and inline `extract_header_as_string`.

crates/daphne-server/src/router/mod.rs
```diff
@@ -324,18 +327,97 @@ where
...
+                    reason.either(
+                        |reason| (StatusCode::UNAUTHORIZED, reason),
+                        |_| (StatusCode::INTERNAL_SERVER_ERROR, "internal error".into()),
+                    )
+                })?;
+            true
+        } else {
+            false
+        };
+        let mtls_authed = if let Some(verification_result) = cf_tls_client_auth {
```
cjpatton, line 391: Right now mutual TLS is only supported for the taskprov owner. We should therefore only use a client certificate to authorize access to the taskprov tasks.

crates/daphne-server/src/router/mod.rs
```diff
@@ -324,18 +327,97 @@ where
...
+            state
+                .server_metrics()
+                .auth_method_inc(metrics::AuthMethod::TlsClientAuth);
+            if verification_result != "SUCCESS" {
```
cjpatton, line 395: Why is this sufficient? This check is very Cloudflare specific and it makes some assumptions about how the zone is set up. Let's document those assumptions here.

crates/daphne-server/src/router/mod.rs
```diff
@@ -390,9 +480,28 @@ mod test {
...
+                    .then_some(())
+                    .ok_or_else(|| Left("invalid token".into()))
+            }
         }
 
         async fn handler(
+            State(ch): State<Arc<Channel>>,
+            UnauthenticatedDapRequestExtractor(req): UnauthenticatedDapRequestExtractor,
+        ) -> impl IntoResponse {
+            ch.send(req).await.unwrap();
```
cjpatton, line 501: Hmm, the return type of this function is `()`. Is that intended? Is this unwrap() safe?

crates/daphne-service-utils/src/config.rs
```diff
@@ -22,17 +24,36 @@ pub struct TaskprovConfig {
     #[serde(with = "hex")]
     pub vdaf_verify_key_init: [u8; 32],
 
-    /// Leader, Helper: Method for authorizing Leader requests.
-    #[serde(with = "from_raw_string")]
-    pub leader_auth: DaphneWorkerAuthMethod,
+    /// Peer authentication tokens.
```
cjpatton, line 27:
```suggestion
    /// Peer's bearer token token.
```

crates/daphne-service-utils/src/config.rs
```diff
@@ -22,17 +24,36 @@ pub struct TaskprovConfig {
...
+    pub peer_auth: PeerAuth,
```
cjpatton, line 28:
```suggestion
    pub peer_bearer_token: PeerAuth,
```

crates/daphne-service-utils/src/config.rs
```diff
@@ -232,3 +253,48 @@ mod from_raw_string {
         serde_json::from_str(&s).map_err(<D::Error as de::Error>::custom)
     }
 }
+
```
cjpatton, line 256: In a comment, say why we need to implement deserialization ourselves.

cjpatton (commented):
Looking good overall. I believe there's one last authentication bug: if the peer presents a valid TLS certificate, it gets access to any task. Mutual TLS is only meant to be available for the taskprov owner at the moment.

.github/workflows/daphneci.yml
```diff
@@ -32,8 +32,6 @@ jobs:
         run: cargo fmt --all --check
       - name: Documentation
         run: RUSTDOCFLAGS='-D warnings' cargo doc --locked --no-deps --all-features --workspace
-      - name: Build
-        run: cargo build --all-targets
```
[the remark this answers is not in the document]
mendess: It doesn't catch anything the cargo hack or the e2e test doesn't catch. So it just adds time to our CI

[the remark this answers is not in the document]
cjpatton: Ack. Nothing builds "experimental" right now, which is a problem, but we can fix in a separate PR.

crates/daphne-server/src/roles/leader.rs
```diff
@@ -165,16 +140,31 @@ impl crate::App {
                 .map_err(|e| fatal_error!(err = ?e, "failed to construct content-type header"))?,
         );
 
-        if let Some(bearer_token) = req.sender_auth.and_then(|auth| auth.bearer_token) {
```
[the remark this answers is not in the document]
mendess: They are. I had a lot of CI failures until I made that test pass :D

crates/daphne-server/src/roles/mod.rs
```diff
@@ -32,18 +36,78 @@ pub async fn fetch_replay_protection_override(kv: Kv<'_>) -> ReplayProtection {
...
+}
+
+impl<'s> From<kv::Kv<'s>> for BearerTokens<'s> {
+    fn from(kv: kv::Kv<'s>) -> Self {
+        Self { kv }
+    }
+}
+
+impl BearerTokens<'_> {
+    #[cfg(feature = "test-utils")]
```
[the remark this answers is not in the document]
mendess: No, without this we get a dead_code warning

crates/daphne-server/src/lib.rs
```diff
@@ -100,6 +112,59 @@ impl router::DaphneService for App {
     fn signing_key(&self) -> Option<&p256::ecdsa::SigningKey> {
         self.service_config.signing_key.as_ref()
     }
+
+    async fn check_bearer_token(
+        &self,
+        presented_token: &BearerToken,
+        sender: DapSender,
+        task_id: TaskId,
+        is_taskprov: bool,
```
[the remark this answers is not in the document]
mendess: My goal here was to avoid passing in the entire DapRequest just to check some of it's fields. It's specially awkward since the `sender` may not be present in the DapRequest and then this method has to deal with that. 

I don't like this bool parameter, but it felt like the least bad option

crates/daphne-server/src/roles/mod.rs
```diff
@@ -32,18 +36,78 @@ pub async fn fetch_replay_protection_override(kv: Kv<'_>) -> ReplayProtection {
...
+
+    pub async fn matches(
+        &self,
+        role: DapSender,
+        task_id: TaskId,
+        token: &BearerToken,
+    ) -> Result<bool, storage_proxy_connection::Error> {
+        self.kv
+            .peek::<kv::prefix::KvBearerToken, _, _>(
+                &(role, task_id).into(),
```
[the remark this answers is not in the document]
mendess: I was careful to keep the same format. I'm not using a auto generated implementation 
https://github.com/cloudflare/daphne/pull/682/files#diff-e12772d8564bc4a07a8ea34a0b5de4b0af7a0ee00119877bdf8a2f2aa707e692R127-R130

crates/daphne-server/src/roles/mod.rs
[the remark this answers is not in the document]
cjpatton: Which could is dead? Don't we expect all of these methods to get used in prod?

crates/daphne-server/src/router/mod.rs
mendess: This also needs fixing, but this is how it is currently implemented in main so I'd rather postpone to another PR

crates/daphne-server/src/router/mod.rs
mendess: Yes, this is fine. We are actually just "smugling" the request out through the channel so we can observe in the test how it was parsed

crates/daphne-server/src/router/mod.rs
cjpatton: Please file an issue to track (or an internal ticket if you prefer).

crates/daphne-server/src/router/mod.rs
cjpatton: Add a comment to explain.

crates/daphne-service-utils/src/config.rs
mendess: this change makes configuration go from
```
peer_auth.leader.expected_token = "I-am-the-leader" # SECRET
```
to
```
peer_bearer_token.leader.expected_token = "I-am-the-leader" # SECRET
```

which is a bit redundant. I can instead change the `PeerAuth` type to `PeerBearerToken`

cjpatton: changing the name of the type SGTM

crates/daphne-server/src/router/mod.rs
```diff
@@ -19,39 +19,58 @@ use axum::{
...
+    /// - right: an internal error that made checking the token impossible.
+    async fn check_bearer_token(
+        &self,
+        presented_token: &BearerToken,
+        sender: DapSender,
+        task_id: TaskId,
+        is_taskprov: bool,
+    ) -> Result<(), Either<String, DapError>>;
+
+    async fn is_taskprov(&self, req: &DapRequest) -> Result<bool, DapError>;
```
cjpatton, line 67: Add documentation

crates/daphne-server/src/router/mod.rs
```diff
@@ -324,21 +329,110 @@ where
...
+    async fn from_request(req: Request<B>, state: &S) -> Result<Self, Self::Rejection> {
+        let bearer_token = extract_header_as_string(req.headers(), http_headers::DAP_AUTH_TOKEN)
+            .map(BearerToken::from);
+        let cf_tls_client_auth = extract_header_as_string(req.headers(), "X-Client-Cert-Verified");
+
+        let request = UnauthenticatedDapRequestExtractor::from_request(req, state)
+            .await?
+            .0;
+
+        let is_taskprov = dbg!(state
```
cjpatton, line 369: Docs for `dbg!()` say this is for "quick and dirty" debugging, so maybe we should remove it.

crates/daphne-server/src/roles/leader.rs
```diff
@@ -165,18 +140,48 @@ impl crate::App {
...
-                        http_headers::DAP_AUTH_TOKEN
-                    )
-                })?,
-            );
-        }
+        // TODO: it is posssible for a taskprov request to not contain this, for example, when
+        // sending a collect request to a helper, this parameter is not necessary since the task
+        // must have already been configured in a previous request.
+        //
+        // Therefore a better way to handle taskprov auth here must be designed.
```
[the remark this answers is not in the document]
mendess: if that's the case than this comment can be removed

crates/daphne-server/src/lib.rs
```diff
@@ -100,6 +112,59 @@ impl router::DaphneService for App {
...
+            .filter(|_| self.service_config.global.allow_taskprov && is_taskprov)
+        {
+            match (&taskprov.peer_auth, sender) {
+                (PeerBearerToken::Leader { expected_token }, DapSender::Leader)
+                | (PeerBearerToken::Collector { expected_token }, DapSender::Collector)
+                    if expected_token == presented_token =>
+                {
+                    Ok(())
+                }
+                _ => reject(format_args!("using taskprov")),
```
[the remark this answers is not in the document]
mendess: that makes sense. It's likely a configuration error

PR #684, opened 2024-09-23
make task id mandatory in DapRequest

mendess:
- Move authentication logic out of the daphne crate
- use in memory store for auth tokens in interop tests
- Make task id obligatory in DapRequest
- Don't pass TaskId by reference


cjpatton (approved):
Nice, good catch on this one. Note that in draft-02 there were some DAP requests for which we didn't know the task ID, hence it being optional. This is no longer the case in 09 and up.

PR #685, opened 2024-09-24
Refactor DapRequest in order to extract parsing code out of protocol logic

mendess:
- use DapError in DapExtractor
- Move DapRequest and it's extractors to their own modules
- Split the request into payload and metadata.
- Parameterize DapRequest by its payload type


crates/daphne-server/src/router/extractor.rs
```diff
@@ -1,30 +1,104 @@
...
+use super::{AxumDapResponse, DaphneService};
+
+pub trait DecodeFromDapHttpBody: Sized {
+    fn decode_from_http_body(bytes: Bytes, meta: &DapRequestMeta) -> Result<Self, DapAbort>;
+}
+
+pub enum AggregationJobReq {
+    Init(AggregationJobInitReq),
+    // TODO: support continue requests
+}
```
cjpatton, line 35: Do you mean `AggregationJobContinueReq`?

This is required for VDAFs that require more than one round of "preparation". This is the case for Poplar1, but we don't currently implement this. For Prio2, Prio3, Pine, and Mastic (to be implemented), we only need one. 

I'm betting that we're never going to need to implement Poplar1, in which case we won't need to support variants. My suggestion would be not write code in anticipation of this feature.

Alternatively, replace the TODO with a note:

```rust
pub enum AggregationJobReq {
    Init(AggregationJobInitReq),
    // NOTE We currently don't support multi-round VDAFs, hence why we don't support `AggregationJobContinueReq`.
}
```

crates/daphne-server/src/router/extractor.rs
```diff
@@ -76,8 +150,8 @@ where
             None
         };
 
-        // TODO(mendess): this is very eager, we could redesign DapResponse later to allow for
-        // streaming of data.
+        // TODO(mendess): this allocates needlessly, if prio supported some kind of
+        // AsyncParameterizedDecode we could avoid this allocation
```
cjpatton, line 154: Consider filing an issue.

crates/daphne-server/src/router/extractor.rs
```diff
@@ -239,12 +317,27 @@ mod test {
...
     use tower::ServiceExt;
 
-    use super::DapRequestExtractor;
-
-    use super::UnauthenticatedDapRequestExtractor;
+    use super::DecodeFromDapHttpBody;
 
     const BEARER_TOKEN: &str = "test-token";
 
+    // in these tests we don't care about the body, so we can type alias it away
```
cjpatton, line 324: Grammar: "in" -> "In" and add a period "."

crates/daphne/src/error/aborts.rs
```diff
@@ -198,7 +203,7 @@ impl DapAbort {
             });
 
         Self::BadRequest(format!(
-            "unexpected content-type: got {got_content_type}; want {want_content_type}"
+            "unexpected content-type: got {got_content_type}; want any of {want_content_type:?}"
         ))
```
cjpatton, line 207: Why "any of"?

crates/daphne/src/messages/request.rs
```diff
@@ -33,7 +33,7 @@ pub enum DapResource {
     Undefined,
 }
 
-/// Fields common to all dap requests.
+/// Fields common to all dap requests
```
cjpatton, line 36: Grammar + capitalize acronyms: "DAP requests."

crates/daphne/src/roles/mod.rs
```diff
@@ -528,14 +526,13 @@ mod test {
...
-        pub fn leader_req<M: ParameterizedEncode<DapVersion>>(
+        pub fn leader_req<M>(
             task_id: &TaskId,
             task_config: &DapTaskConfig,
             agg_job_id: Option<&AggregationJobId>,
             media_type: DapMediaType,
-            msg: M,
-        ) -> DapRequest {
-            let payload = msg.get_encoded_with_param(&task_config.version).unwrap();
+            payload: M,
```
cjpatton, line 534: "M" -> "P" so that we use the same generic parameter name as the definition of `DapRequest`

crates/daphne/src/testing/mod.rs
```diff
@@ -1067,30 +1068,44 @@ impl DapLeader for InMemoryAggregator {
             .finish_collect_job(task_id, coll_job_id, collection)
     }
 
-    async fn send_http_post(&self, req: DapRequest, _url: Url) -> Result<DapResponse, DapError> {
+    async fn send_http_post<M>(
```
cjpatton, line 1071: Likewise here and below ("M" -> "P")

cjpatton (approved):
Beautiful work, just nit-picky comments.

crates/daphne/src/error/aborts.rs
mendess: This changed to an array to support DapAggregationJobContinueReq. But I'm going to revert it back to a single content type

PR #691, opened 2024-10-09
Add test-routes create-add-task-json to help with adding new tasks to test aggregators

mendess:


crates/dapf/README.md
```diff
@@ -46,6 +46,54 @@ used when issuing requests to `internal/test/add_task`:
 DAP and base64 encoded hpke config: mwAgAAEAAQAgpj3X5rv8aNVOiyXjnOuCbN9RAVkqAmWQy5NddU5PIQ0
 ```
 
+### Generating the requert payload for `internal/test/add_task`
+
```
cjpatton, line 50: document the draft that defines this endpoint

crates/dapf/src/main.rs
```diff
@@ -116,6 +136,82 @@ impl CliVdafConfig {
...
+                },
+            }))
+        } else {
+            Err(format!("{s} is an invalid query config"))
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+struct CliTaskId(TaskId);
```
cjpatton, line 185: Is Cli short for "command line interface" or is it short for "Client"? Why do we need a wrapper for task IDs here?

cjpatton (approved):

crates/dapf/src/main.rs
mendess: We need a wrapper to TaskId to impl `FromStr` for it. Previously this problem was solved by using `value_parser` from clap, but by using this new type we can also ask the user interactively. See `request_from_user`

mendess: Cli stands for command line interface. I intend to move all these `Cli*` types to their own module which will make this less confusing in the future

PR #692, opened 2024-10-09
Don't require async when decrypting reports

mendess:
Decrypting reports usually happens as they are consumed, which is a CPU bound
task. Not requiring async for this has a few advantages:
- it can be more easily offloaded to a thread outside the async runtime to
    prevent it from blocking the runtime
- it can be more easily benchmarked and flamegraphed, as flamegraphing async
    functions is almost impossible


cjpatton (approved):
Nice!

PR #693, opened 2024-10-10
Display Id types as base64 instead of hex

mendess:


RageKnify (approved):

RageKnify (approved):

cjpatton (approved):
🦆 , but just to be 100% clear: the spec requires URL-safe base64, i.e., `base64url`, not plain base64.

PR #694, opened 2024-10-11
Bring leader up to spec in regards to collection

mendess:
- Use the same url for starting and polling collection jobs
- Simplify router test code
- Hardcode sender based on route instead of deducing based on content type
- Parameterize DapRequest on the resource it holds.

### DIFF
--- a/crates/daphne-server/src/router/extractor.rs
+++ b/crates/daphne-server/src/router/extractor.rs
@@ -26,6 +26,7 @@ use crate::metrics;
 
 use super::{AxumDapResponse, DaphneService};
 
+/// Trait used to decode a DAP http body.
 pub trait DecodeFromDapHttpBody: Sized {
     fn decode_from_http_body(bytes: Bytes, meta: &DapRequestMeta) -> Result<Self, DapAbort>;
 }
@@ -39,7 +40,9 @@ macro_rules! impl_decode_from_dap_http_body {
                     meta: &DapRequestMeta,
                 ) -> Result<Self, DapAbort> {
                     let mut cursor = Cursor::new(bytes.as_ref());
+                    // Check that media type matches.
                     meta.get_checked_media_type(DapMediaType::$type)?;
+                    // Decode the body
                     $type::decode_with_param(&meta.version, &mut cursor)
                         .map_err(|e| DapAbort::from_codec_error(e, meta.task_id))
                 }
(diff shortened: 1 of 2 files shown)


crates/daphne-server/src/router/extractor.rs
```diff
@@ -159,14 +159,34 @@ where
...
+    pub const fn to_enum(id: DapSender) -> daphne::DapSender {
+        match id {
+            FROM_CLIENT => daphne::DapSender::Client,
+            FROM_COLLECTOR => daphne::DapSender::Collector,
+            FROM_HELPER => daphne::DapSender::Helper,
+            FROM_LEADER => daphne::DapSender::Leader,
+            _ => panic!("invalid dap sender. Please specify a valid dap_sender from the crate::extractor::dap_sender module"),
+        }
+    }
+}
```
cjpatton, line 223: I guess the purpose of this is to be able to express the sender as a trait bound? That's pretty cool.

Why not just replace `daphne::DapSender` with thee `DapSender` in this module? It seems to me that long term this layer of indirection will get confusing.

crates/daphne-server/src/router/extractor.rs
```diff
@@ -60,36 +59,76 @@ impl DecodeFromDapHttpBody for () {
...
         #[derive(Debug, Deserialize)]
         #[serde(deny_unknown_fields)]
-        struct PathParams {
+        struct PathParams<R: resource_parsers::Resource> {
             version: DapVersion,
             task_id: TaskId,
-            agg_job_id: Option<AggregationJobId>,
-            collect_job_id: Option<CollectionJobId>,
+            #[serde(flatten)]
+            resource_parser: R::Parser,
```
cjpatton, line 141: What happens if deserializtion fails? Will this be handled as an internal error or a DAP abort?

crates/daphne-server/src/router/extractor.rs
```diff
@@ -37,7 +38,9 @@ macro_rules! impl_decode_from_dap_http_body {
                     meta: &DapRequestMeta,
                 ) -> Result<Self, DapAbort> {
                     let mut cursor = Cursor::new(bytes.as_ref());
+                    // check that media type matches.
```
cjpatton, line 41: Grammar: Start sentence with a capital letter

crates/daphne-server/src/router/extractor.rs
```diff
@@ -37,7 +38,9 @@ macro_rules! impl_decode_from_dap_http_body {
...
                     meta.get_checked_media_type(DapMediaType::$type)?;
+                    // decode the body
```
cjpatton, line 43: Grammar: Start sentence with a capital letter and end with a period.

crates/daphne-server/src/router/extractor.rs
```diff
@@ -53,29 +56,42 @@ impl_decode_from_dap_http_body!(
...
+/// Parsers that allow us to have to deduce the resource parser to use based on the resource passed
+/// in through the `R` type parameter in [`DapRequestExtractor`] and
+/// [`UnauthenticatedDapRequestExtractor`].
 mod resource_parsers {
     use daphne::messages::request::resource;
     use serde::{de::DeserializeOwned, Deserialize};
 
     #[derive(Deserialize)]
     pub struct AggregationJobIdParser {
+        /// this field name defines the parameter name in the route matcher because we use
```
cjpatton, line 75: Grammar

crates/daphne-server/src/router/extractor.rs
```diff
@@ -53,29 +56,42 @@ impl_decode_from_dap_http_body!(
...
+        /// `#[serde(flatten)]` in the
+        /// `UnauthenticatedDapRequestExtractor::from_request::PathParams::resource_parser` field.
         agg_job_id: resource::AggregationJobId,
     }
 
     #[derive(Deserialize)]
     pub struct CollectionJobIdParser {
+        /// this field name defines the parameter name in the route matcher because we use
```
cjpatton, line 83: Grammar

cjpatton (commented):
Beautiful :) One blocking thing, which is to make sure deserialization failures are handled properly.

Also, I think we should just replace `daphne::DapSender` with the new fancy version.

crates/daphne-server/src/router/extractor.rs
cjpatton: If I'm being honest, I no longer completely follow the architecture of `daphne-server`. The ask here is to make sure that, if parsing a DAP request fails, the Helper responds as specified in the protocol: with 400 Bad Request and a Problems Detail document with the proper error.

crates/daphne-server/src/router/extractor.rs
mendess: The purpose of this is to be able to pass the sender as a const generic parameter, which currently only supports very primitive types (numbers). 

Using integers for "enums" has the problem that whenever you match on then you have to include a `_ =>` branch to handle the default value where you probably have to just panic (like I did here), so all the code that currently matches on DapSender will become worse.

This set of constants will never be used outside of these extractors. At least I don't intend it to be used outside of here. And if this is true then I can actually guarantee that this panic never happen at runtime because `to_enum` is `const` and it's [called in a const block](https://github.com/cloudflare/daphne/pull/694/files#diff-a55f2d195924b3d5fef4513679389e84a00d89a0ca8c4b13a26b714be3b9b23cR281), which forces the compiler to execute it at compile time.

Give it a go. Checkout this code and change `FROM_HELPER` [https://github.com/cloudflare/daphne/pull/694/files#diff-963e76984effba09b673f5f9da56571fdb5325e1907af4a2dfd197165d1ead8eR72](here) to `42` and you'll see a compile time error (if you do `cargo build`, sadly `cargo check` doesn't catch it)

cjpatton: I see. This makes me sad, but I think it really improves readability here.

cjpatton: I'm cool with keeping it.

Another option might be to replace the enum with "marker traits" that are supertraits of `DapSender`. Something to noodle on later.

cjpatton (approved):

PR #695, opened 2024-10-15
Reorganize dapf

mendess:
- Separate out human friendly parsers of DAP types to seperate module
- Cleanup TestAction::CreateAddTaskJson
- create dap functions module
- Move to functions::helper the code needed to make a request to the helper
    which will allow us to possibly reuse this code in a individual dapf function


crates/dapf/src/cli_parsers.rs
```diff
@@ -0,0 +1,324 @@
+//! Human friendly parsers for common types of parameters to DAP functions.
+// Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
```
cjpatton, line 3: The copyright notice should be at the top of the file. The module docstring should have newlines around it.

crates/dapf/src/cli_parsers.rs
```diff
@@ -0,0 +1,324 @@
...
+    vdaf::{Prio3Config, VdafConfig},
+    DapQueryConfig,
+};
+
+#[derive(Debug, Clone, Copy, Default)]
+pub enum DefaultVdafConfigs {
+    Prio2Dimension99k,
+    #[default]
+    Prio3NumProofs2,
+    Prio3NumProofs3,
```
cjpatton, line 26: These names are very strange and need documentation. For each variant, I would spell out what VDAF is in use and what are its parameters.

I also think it's worth saying something like "these variants are used for load testing".

crates/dapf/src/cli_parsers.rs
```diff
@@ -0,0 +1,324 @@
...
+                        } else {
+                            default_err
+                        }
+                    })
+            })
+    }
+}
+
+impl CliVdafConfig {
+    pub fn into_vdaf(self) -> VdafConfig {
```
cjpatton, line 118: "into_config()"?

crates/dapf/src/cli_parsers.rs
cjpatton: BTW I see this code is just moved, but we might as well add some documentation!

crates/daphne-service-utils/src/test_route_types.rs
```diff
@@ -23,6 +26,47 @@ pub struct InternalTestVdaf {
     pub chunk_length: Option<String>,
 }
 
+impl From<VdafConfig> for InternalTestVdaf {
```
cjpatton, line 29: What does "Internal" mean in this context?

crates/dapf/src/functions/mod.rs
```diff
@@ -0,0 +1,6 @@
+//! The various DAP functions dapf can perform.
+// Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
```
cjpatton, line 3: Copyright at the top, module docstring in its own paragraph

crates/dapf/src/functions/mod.rs
```diff
@@ -0,0 +1,6 @@
...
+
+pub mod hpke;
+pub mod test_routes;
```
cjpatton, line 7: What does "test routes" mean in this context? If you're referring to the interop API, then document it here. If these includce non-standard APIs that are specific to Daphne, then document this as well.

crates/dapf/src/acceptance/mod.rs
```diff
@@ -509,62 +504,30 @@ impl Test {
...
-                    agg_job_init_req
-                        .get_encoded_with_param(&task_config.version)
-                        .unwrap(),
-                )
-                .headers(headers),
-        )
-        .await?;
+        let agg_job_resp = self
+            .http_client
+            .submit_aggregate_init_req(
```
cjpatton, line 514: "aggregation_job_init_req"

cjpatton (approved):
Minor nits only. Note one of the commit titles is so long that GitHub splits it out into the commit body. I think it's a good idea to make the title fit on one line so that `git log` is easier to read.

crates/daphne-service-utils/src/test_route_types.rs
mendess: You came up with this name, [you tell me](https://github.com/cloudflare/daphne/pull/153) :eyes:

crates/dapf/src/functions/mod.rs
mendess: It maps to routes in the file with the same name `daphne-server/src/router/test_routes.rs`. Do you think we should rename all of these to something that references interop more clearly?

crates/dapf/src/cli_parsers.rs
mendess: I just wanted some quick default vdafs to pass through the commandline. So I wanted the names to be short which means they won't be very good either

crates/daphne-service-utils/src/test_route_types.rs
cjpatton: It wasn't clear to me that this was related to one of the /internal/... endpoints. Perhaps just add some documentation.

crates/dapf/src/functions/mod.rs
cjpatton: Thanks for clarifying. No need to rename things, just add some documentation.

crates/dapf/src/cli_parsers.rs
cjpatton: The names are find, just document :)

PR #696, opened 2024-10-15
Trim daphne-server-utils

mendess:
This crate was created during the transition from the 100% worker implementation
to the native version. It contained types that both implementations needed. Now
some of it's code can be merged into the daphne-server implementation. Reducing
it's size and making more obvious the types that are actually shared by other
members of the workspace.


cjpatton (approved):
🦆

PR #699, opened 2024-10-15
Don't request the entire Aggregate Share just to check if it exists

### LINKED ISSUE #581: Don't request the entire Aggregate Share just to check if it exists
current [`batch_exists`](https://github.com/cloudflare/daphne/blob/4c612db3d42230a97899ccfe09e0f2566af94bbf/daphne_server/src/roles/aggregator.rs#L310) implementation fetches the entire Aggregate Share just to see if it exists before fetching it again to return it. This takes 600ms on my machine

mendess:
Fixes #581

- Add method for getting just the report count of an aggregate share
- Use AggregateStore::ReportCount to check if a batch exists instead of loading the entire aggregate share

### DIFF
--- a/crates/daphne-server/src/roles/aggregator.rs
+++ b/crates/daphne-server/src/roles/aggregator.rs
@@ -19,7 +19,7 @@ use daphne::{
 use daphne_service_utils::durable_requests::bindings::{
     self, AggregateStoreMergeOptions, AggregateStoreMergeReq, AggregateStoreMergeResp,
 };
-use futures::{future::try_join_all, StreamExt, TryStreamExt};
+use futures::{future::try_join_all, StreamExt, TryFutureExt, TryStreamExt};
 use mappable_rc::Marc;
 use rayon::prelude::{IntoParallelIterator, ParallelIterator};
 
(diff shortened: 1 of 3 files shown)


cjpatton (commented):
Don't we need to make the same change for `MockAggregator` in `daphne::testing`?

crates/daphne-server/src/roles/aggregator.rs
```diff
@@ -316,15 +316,33 @@ impl DapAggregator for crate::App {
...
+                let params = (version, task_id, &bucket);
+
+                let new_get_report_count = || {
+                    durable
+                        .request(bindings::AggregateStore::ReportCount, params)
+                        .send::<u64>()
+                };
+
+                let old_get_report_count = || {
+                    durable
```
cjpatton, line 329: What do you mean by "old" and "new" here?

mendess: it's the old way to do it and the new way to do it. Just because I want the daphne-server to be compatible with the previous version of the worker while things transition. Do you have a better name of these variables? I honestly don't like what I came up with

cjpatton: Names are fine, I was not clear what the behavior was. Please add some documentation here to note that the old method is for backwards compatibility and will be removed after some future deploymenet chnage.

mendess:
> Don't we need to make the same change for `MockAggregator` in `daphne::testing`?

There is no MockAggregator in `daphne::testing` :thinking:

cjpatton:
> > Don't we need to make the same change for `MockAggregator` in `daphne::testing`?
> 
> There is no MockAggregator in `daphne::testing` 🤔

Disregard, I was thinking of `InMemoryAggregator`, which used to be called `MockAggregator`. But this PR isn't changing the role traits, so it's not relevant to `InMemoryAggregator` at al.

cjpatton (approved):

PR #705, opened 2024-10-21
Remove unused dependencies

mendess:


crates/daphne-worker-test/Cargo.toml
```diff
@@ -20,14 +20,8 @@ default = ["console_error_panic_hook"]
 [dependencies]
 cap.workspace = true
 cfg-if.workspace = true
-# The `console_error_panic_hook` crate provides better debugging of panics by
-# logging them with `console.error`. This is great for development, but requires
-# all the `std::fmt` and `std::panicking` infrastructure, so isn't great for
-# code size when deploying.
```
RageKnify, line 26: Comment could stay

RageKnify (approved):

crates/daphne-worker-test/Cargo.toml
mendess: I think it's misleading. It's not like we are avoiding `std::fmt` or `std::panicking` in the rest of the codebase

PR #706, opened 2024-10-22
helper: add metrics for agg job timings

Noah-Kennedy:


crates/daphne-server/src/metrics.rs
```diff
@@ -76,6 +78,10 @@ mod prometheus {
         fn daphne(&self) -> &dyn DaphneMetrics {
             self
         }
+
+        fn aggregate_job_latency(&self, _time: Duration) {
+            // unimplemented by default
+        }
```
mendess, line 84: Why is this unimplemented by default?

Noah-Kennedy: Because it's higher cardinality and I'm not sure who all wants to use it.

mendess: Alright, leave a comment in the code with this justification and we can merge

mendess (approved):

PR #707, opened 2024-10-29
Change reports from PUT to POST

jhoyla:
In draft-10, based on the report from the HTTP Directorate, clients should now upload reports with `POST`s rather than `PUT`s. This PR implements this change.

### DIFF
--- a/crates/daphne-server/src/router/leader.rs
+++ b/crates/daphne-server/src/router/leader.rs
@@ -8,14 +8,14 @@ use axum::{
     extract::State,
     http::StatusCode,
     response::{IntoResponse, Response},
-    routing::put,
+    routing::{post, put},
 };
 use daphne::{
     constants::DapMediaType,
     error::DapAbort,
     messages::{self, request::resource},
     roles::leader::{self, DapLeader},
-    DapError,
+    DapError, DapVersion,
 };
 use prio::codec::ParameterizedEncode;
 
@@ -32,7 +32,8 @@ where
     B::Error: Send + Sync,
 {
     router
-        .route("/:version/tasks/:task_id/reports", put(upload))
+        .route("/:version/tasks/:task_id/reports", put(upload_draft09))
+        .route("/:version/tasks/:task_id/reports", post(upload_draft13))
         .route(
             "/:version/tasks/:task_id/collection_jobs/:collect_job_id",
             put(start_collection_job).post(collect),
@@ -46,7 +47,7 @@ where
         version = ?req.version,
     )
 )]
-async fn upload<A>(
+async fn upload_draft13<A>(
     State(app): State<Arc<A>>,
     UnauthenticatedDapRequestExtractor(req): UnauthenticatedDapRequestExtractor<
         messages::Report,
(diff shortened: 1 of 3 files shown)


crates/daphne-server/src/router/leader.rs
```diff
@@ -46,7 +47,7 @@ where
         version = ?req.version,
     )
 )]
-async fn upload<A>(
+async fn upload_13<A>(
```
cjpatton, line 50: Align with notation elsewhere (we write "draft09" instead of "09")
```suggestion
async fn upload_draft13<A>(
```

crates/daphne-server/src/router/leader.rs
```diff
@@ -56,12 +57,45 @@ async fn upload<A>(
 where
     A: DapLeader + DaphneService + Send + Sync,
 {
+    if req.version == DapVersion::Draft09 {
```
cjpatton, line 60: Defensive programming
```suggestion
    if req.version != DapVersion::DraftLatest {
```

crates/daphne-server/tests/e2e/e2e.rs
```diff
@@ -157,6 +158,11 @@ async fn leader_upload(version: DapVersion) {
     let client = t.http_client();
     let hpke_config_list = t.get_hpke_configs(version, client).await.unwrap();
     let path = t.upload_path();
+    let method = if version == DapVersion::Draft09 {
+        &http::Method::PUT
+    } else {
+        &http::Method::POST
+    };
```
cjpatton, line 165: Defensive programming: Use `match` so that if add variants in the future, the compiler tells you if you failed to handle a variant
```suggestion
    let method = match version {
        DapVersion::Draft09 =>  &http::Method::PUT,
        DavVersion::DraftLatest => &http::Method::POST,
    };
```

crates/daphne-server/tests/e2e/test_runner.rs
```diff
@@ -437,114 +438,7 @@ impl TestRunner {
         }
 
         let resp = client
```
cjpatton, line 440: Why are we deleting all of this?

cjpatton (commented):
Nice, just some minor suggestions.

crates/daphne-server/tests/e2e/test_runner.rs
jhoyla: I refactored to make `leader_put_expect_ok` and `leader_post_expect_ok` one function that takes a http::Method parameter, to simplify testing the backwards compatibility of the upload report functions.

crates/daphne-server/src/router/leader.rs
jhoyla: So I was trying to be defensive, by saying that Draft09 is frozen, whereas DraftLatest will change.

cjpatton: Excellent point, I think you're right.

crates/daphne-server/tests/e2e/test_runner.rs
```diff
@@ -359,11 +359,11 @@ impl TestRunner {
         get_raw_hpke_config(client, self.task_id.as_ref(), &self.helper_url, "helper").await
     }
 
-    #[expect(dead_code)]
-    pub async fn leader_post_expect_ok(
+    pub async fn leader_request_expect_ok(
         &self,
         client: &reqwest::Client,
         path: &str,
+        method: &http::Method,
```
mendess, line 366: Don't use a reference here. I don't think there's much to gain from it, and you end up cloning anyway. So just take by value

jhoyla: The borrow checker didn't let me. This was just my hack to make the borrow checker happy :sweat_smile:

PR #712, opened 2024-11-08
Coalesce requests to KV

mendess:
This PR is best viewed commit by commit.

- Add get or insert method to kv (this commit is setup, this method will later being coalesced)
- Abstract away the internals of kv connection
- rename get_impl to get_internal
- Add request coalescer to kv (this is the important commit to review)

### DIFF
--- a/crates/daphne-server/src/storage_proxy_connection/kv/request_coalescer.rs
+++ b/crates/daphne-server/src/storage_proxy_connection/kv/request_coalescer.rs
@@ -1,0 +1,494 @@
+// Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+//! This module provides a means to coalesce arbitrary requets to KV.
+//!
+//! When issuing multiple identical requests, these can be coalesced into one single network
+//! request.
+//!
+//! The way this is done is by assigning one of the callers as the `executor` of the request and
+//! the other callers as `waiters` of the request. The first caller is given the `executor` role.
+//!
+//! In case the `executor` is canceled for any reason, one of the `waiters` must be promoted to
+//! `executor`, this is done by instructing the callers to retry the coalesced request. **This
+//! means that the request must be idempotent or retrying might be unsafe.**
+
+use mappable_rc::Marc;
+use operation_id::OpId;
+use std::{
+    any::Any,
+    collections::{hash_map::Entry, HashMap},
+    future::Future,
+    marker::PhantomData,
+    sync::Mutex,
+};
+use tokio::sync::watch::{self, Receiver, Sender};
+
+/// For coalescing to work correctly the key we use for lookup has to be simultaneously:
+/// - very unique: distinct code can't have the same key
+/// - very duplicated: the same code must use the same key
+///
+/// To achieve this we use a key pair:
+/// - `key` is the KV key, obviously distinct keys must originate distinct requests.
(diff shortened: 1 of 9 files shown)


crates/daphne-server/src/storage_proxy_connection/kv/request_coalescer.rs
```diff
@@ -0,0 +1,494 @@
...
+use std::{
+    any::Any,
+    collections::{hash_map::Entry, HashMap},
+    future::Future,
+    marker::PhantomData,
+    sync::Mutex,
+};
+use tokio::sync::watch::{self, Receiver, Sender};
+
+/// For coalescing to work correctly the key the key we use for lookup has to be simultaneously:
```
armfazh, line 27:
```suggestion
/// For coalescing to work correctly the key we use for lookup has to be simultaneously:
```

crates/daphne-server/src/storage_proxy_connection/kv/request_coalescer.rs
```diff
@@ -0,0 +1,494 @@
...
+    #[cfg(feature = "test-utils")]
+    shutting_down: std::sync::atomic::AtomicBool,
+}
+
+impl RequestCoalescer {
+    #[cfg(feature = "test-utils")]
+    pub async fn reset(&self) {
+        self.shutting_down
+            .store(true, std::sync::atomic::Ordering::SeqCst);
+        let recievers = self
```
armfazh, line 60:
```suggestion
        let receivers = self
```

crates/daphne-server/src/storage_proxy_connection/kv/request_coalescer.rs
```diff
@@ -0,0 +1,494 @@
...
+            #[cfg(feature = "test-utils")]
+            assert!(
+                !self.shutting_down.load(std::sync::atomic::Ordering::SeqCst),
+                "RequestCoalescer is shutting down"
+            );
+
+            let op_key = OperationKey {
+                op_id: operation_id::of::<F, Fut, R, E>(),
+                key: key.clone(),
+            };
```
armfazh, line 98: can the `op_key` variable be defined before the loop?

crates/daphne-server/src/storage_proxy_connection/kv/request_coalescer.rs
```diff
@@ -0,0 +1,494 @@
...
+            Self::Canceled => Self::Canceled,
+            Self::Error(e) => Self::Error(e.clone()),
+        }
+    }
+}
+
+type CoalescedValue =
+    Result<Option<Marc<dyn Any + Send + Sync>>, CoalesceError<dyn Any + Send + Sync>>;
+
+/// The send half to be used by the `executor` to send the result of it's computation to the
```
armfazh, line 151:
```suggestion
/// The send have to be used by the `executor` to send the result of it's computation to the
```

armfazh (approved):
Nice functionality, changes are ok.

crates/daphne-server/src/storage_proxy_connection/kv/request_coalescer.rs
mendess: This is actually correct. A channel has two parts, the sending part and the receiving part, commonly referred to as the sending and receiving halves

crates/daphne-server/src/storage_proxy_connection/kv/request_coalescer.rs
mendess: That results in an additional clone when passing it to the ResponseSender

PR #717, opened 2024-11-15
Begin renumbering errors

jhoyla:
Draft 09 and Draft 13 use different numbers to indicate the same errors.
To make this code backwards compatible we need to support both sets at the same time.

### DIFF
--- a/crates/daphne/src/messages/mod.rs
+++ b/crates/daphne/src/messages/mod.rs
@@ -21,6 +21,7 @@ use std::{
     convert::{TryFrom, TryInto},
     fmt,
     io::{Cursor, Read},
+    ops::{Deref, DerefMut},
 };
 
 // Query types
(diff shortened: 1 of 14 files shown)


crates/daphne/src/messages/mod.rs
```diff
@@ -594,7 +611,86 @@ pub enum TransitionFailure {
...
+    ) -> Result<Self, CodecError> {
+        match *version {
+            DapVersion::Draft09 => match TransitionFailureDraft09::decode(bytes) {
+                Ok(x) => Ok(Self::Draft09(x)),
+                Err(x) => Err(x),
+            },
+            DapVersion::Latest => match TransitionFailureLatest::decode(bytes) {
+                Ok(x) => Ok(Self::DraftLatest(x)),
+                Err(x) => Err(x),
+            },
```
mendess, line 654:
```suggestion
            DapVersion::Draft09 => match TransitionFailureDraft09::decode(bytes).map(Self::Draft09),
            DapVersion::Latest => TransitionFailureLatest::decode(bytes).map(Self::DraftLatest),
```

This works because when you define a tuple struct (`struct Foo(i32, f32)`) rust also defines for you a function of the same name as the type (`fn Foo(a: i32, b: f32) -> Foo`), the same is true for tuple enum variants.

crates/daphne/src/messages/mod.rs
```diff
@@ -594,7 +611,86 @@ pub enum TransitionFailure {
...
+                }
+                Err(x) => Err(x),
+            },
+            DapVersion::Latest => match u8::decode(&mut Cursor::new(bytes.as_slice())) {
+                Ok(x) => {
+                    self::TransitionFailure::DraftLatest(x.try_into()?);
+                    Ok(())
+                }
+                Err(x) => Err(x),
+            },
```
mendess, line 679: This should encode, not decode

crates/daphne/src/roles/helper.rs
```diff
@@ -250,23 +253,60 @@ async fn finish_agg_job_and_aggregate(
...
+                                    report_id,
+                                    ReportProcessedStatus::Rejected(
+                                        TransitionFailure::DraftLatest(
+                                            TransitionFailureLatest::ReportReplayed,
+                                        ),
+                                    ),
+                                )
+                            }))
+                        }
+                    };
```
mendess, line 280:
```suggestion
                   let rejection = match task_config.version {
                       DapVersion::Draft09 => ReportProcessedStatus::Rejected(TransitionFailure::Draft09(
                           TransitionFailureDraft09::ReportReplayed,
                        ))
                        DapVersion::Latest => ReportProcessedStatus::Rejected(TransitionFailure::DraftLatest(
                            TransitionFailureLatest::ReportReplayed,
                        ),
                    };
                    report_status.extend(replays.into_iter().map(|report_id| (report_id, rejection)));
```

crates/daphne/src/roles/helper.rs
```diff
@@ -250,23 +253,60 @@ async fn finish_agg_job_and_aggregate(
...
+                                    ReportProcessedStatus::Rejected(
+                                        TransitionFailure::DraftLatest(
+                                            TransitionFailureLatest::BatchCollected,
+                                        ),
+                                    ),
+                                )
+                            }));
+                        }
+                    }
+
```
mendess, line 309: same transformation here

mendess (changes requested):

PR #719, opened 2024-11-19
Upgrade daphne-server to http 1

mendess:
Upgrades to http 1 and axum 0.7, which is a breaking change but for us means a simplification of the code as we no longer have to care about the Body of a request, it is now `axum::body::Body` instead of being generic.


crates/daphne-server/examples/service.rs
```diff
@@ -120,11 +121,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Sync + Send>> {
...
 
     // hand the router to axum for it to run
-    let serve = axum::Server::bind(&std::net::SocketAddr::new(
-        "0.0.0.0".parse().unwrap(),
-        config.port,
-    ))
-    .serve(router.into_make_service());
+    let serve = axum::serve(
+        TcpListener::bind(std::net::SocketAddr::new(
+            "0.0.0.0".parse().unwrap(),
```
armfazh, line 126: could we remove the parsing as not really needed
```suggestion
            IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)),
```

crates/daphne-server/src/router/extractor.rs
```diff
@@ -177,7 +175,7 @@ where
 
         // TODO(mendess): this allocates needlessly, if prio supported some kind of
         // AsyncParameterizedDecode we could avoid this allocation
-        let payload = hyper::body::to_bytes(body).await;
+        let payload = axum::body::to_bytes(body, usize::MAX).await;
```
armfazh, line 178: we may want to only accept a certain amount of bytes based on the payload we are expecting. currently this is not supported, right?

armfazh (approved):
two nits only

crates/daphne-server/src/router/extractor.rs
mendess: Yes, I'm gonna open an issue. The previous implementation didn't allow limiting this so I kept it "the same" but we should investigate if we can deduce a sane limit we can put here

crates/daphne-server/examples/service.rs
mendess: I prefer the parsing one because it's less verbose. I think it's more fitting for a code example

armfazh: parsing has been shown to be a source of [bugs](https://blog.cloudflare.com/incident-report-on-memory-leak-caused-by-cloudflare-parser-bug/). 
I would do parsing, unless there is no more secure option.

PR #720, opened 2024-11-21
Remove `by_batch_id` from fixed-size queries

jhoyla:
As of draft-12 collectors can no longer request batches by batch id, but must use leader selected batches.
These changes also change the wire format for Query, BatchSelector and PartialBatchSelector.
These PRs update the relevant structs.

### DIFF
--- a/crates/daphne/src/messages/mod.rs
+++ b/crates/daphne/src/messages/mod.rs
@@ -365,27 +365,71 @@ impl From<BatchSelector> for PartialBatchSelector {
     }
 }
 
-impl Encode for PartialBatchSelector {
-    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
-        match self {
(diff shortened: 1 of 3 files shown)


mendess (approved):

crates/daphne/src/messages/mod.rs
```diff
@@ -365,27 +365,71 @@ impl From<BatchSelector> for PartialBatchSelector {
...
+                    Self::FixedSizeByBatchId { batch_id } => {
+                        QUERY_TYPE_FIXED_SIZE.encode(bytes)?;
+                        batch_id.encode(bytes)?;
+                    }
+                };
+                Ok(())
             }
-        };
-        Ok(())
+            DapVersion::Latest => {
```
armfazh, line 385: my only concern about using 'latest' is because until the RFC is approved, there could be several 'latest' drafts.

PR #722, opened 2024-11-25
Remove max_batch_query_count from latest

jhoyla:
Draft 12 drops the `max_batch_query_count` field from `QueryConfig`

### DIFF
--- a/crates/daphne-server/tests/e2e/e2e.rs
+++ b/crates/daphne-server/tests/e2e/e2e.rs
@@ -471,6 +471,69 @@ async fn leader_upload_taskprov() {
     .unwrap();
 }
 
+async fn leader_upload_taskprov_wrong_version(version: DapVersion) {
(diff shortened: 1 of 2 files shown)


mendess:
Wouldn't it make sense to make the field `Option` and private and then provide a getter that takes in the version and asserts that nothing gets out of sync?

I don't like how there's a magic number `1` that could be used in code that isn't aware of the current version it's operating under

jhoyla:
I'll give it a go. My goal was trying to make the code easy to rip out later, but I think a private `Option` should be just as easy.

mendess (approved):

jhoyla:
So I had a go with `Option`, but it makes the code more fragile because the `QueryConfig` struct is now dependent on the wire format, rather than version agnostic. The magic value `1` is always correct in version 13, because there is no support for the multi-collection of batches, so you can only query a batch once by definition.

PR #723, opened 2024-11-25
Remove `max_batch_size` from Latest

jhoyla:
Draft-12 removes `max_batch_size` from "fixed-size" batch mode. This PR removes `max_batch_size` from the wire format, and sets the value to 0 for `Latest`, which, per [taskprov-00](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-taskprov-00), the last version to feature `max_batch_size`, signals that there is no `max_batch_size`. 

This PR also makes `max_batch_size` the same type, `u32`, everywhere in the code. The DAP spec doesn't specify a size, so I've take the size from `taskprov`.

### DIFF
--- a/crates/daphne/src/lib.rs
+++ b/crates/daphne/src/lib.rs
@@ -83,7 +83,7 @@ use std::{
     cmp::{max, min},
     collections::{HashMap, HashSet},
     fmt::Debug,
-    num::NonZeroUsize,
+    num::{NonZeroU32, NonZeroUsize},
     str::FromStr,
 };
 use url::Url;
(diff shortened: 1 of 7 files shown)


armfazh (approved):

crates/daphne/src/lib.rs
```diff
@@ -732,7 +732,7 @@ impl DapTaskConfig {
             DapQueryConfig::FixedSize {
                 max_batch_size: Some(max_batch_size),
             } => {
-                if report_count > max_batch_size {
+                if (report_count > u64::from(max_batch_size)) && max_batch_size != 0 {
```
mendess, line 735: why are you checking that max_batch_size != 0 here?

jhoyla: Per the old draft a `max_batch_size` of 0 means there is no limit.
```
The maximum batch size for fixed_size query is optional. If query_type is fixed_size and max_batch_size is 0, then the task does not have maximum batch size limit.
```
https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-taskprov-00#section-3.1-7

mendess (approved):
Small nit picks around shorter ways to convert between numeric types

PR #728, opened 2024-11-27
Make more use of DapRole enums

mendess:
The DapRole enum is underused and under featured, leading to code like `is_leader: bool` where an instance of the enum would make calling code more readable (there are places where `false /* helper */` can be see where `DapRole::Helper` would be much better)

This PR also factors out the calculation of the `info` and `aad` parameters needed for hpke encryption and decryption. Making this part both more type safe and taking it out of higher level code

This PR should be reviewed one commit at a time

### DIFF
--- a/crates/daphne/src/hpke.rs
+++ b/crates/daphne/src/hpke.rs
@@ -17,6 +17,7 @@ use crate::{
     DapError, DapVersion,
 };
 use async_trait::async_trait;
+use info_and_aad::InfoAndAad;
 use serde::{Deserialize, Serialize};
 use std::{borrow::Borrow, ops::Deref};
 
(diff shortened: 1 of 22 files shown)


crates/daphne/src/hpke.rs
```diff
@@ -437,6 +438,135 @@ impl From<HpkePublicKeySerde> for HpkePublicKey {
...
+
+    fn into_bytes(constant: &'static [u8], sender: DapRole, receiver: DapRole) -> Vec<u8> {
+        static ROLE_2_ID: [u8; 4] = {
+            let mut ids = [0; 4];
+            ids[DapRole::Collector as usize] = 0;
+            ids[DapRole::Client as usize] = 1;
+            ids[DapRole::Leader as usize] = 2;
+            ids[DapRole::Helper as usize] = 3;
+            ids
+        };
```
armfazh, line 560: it would be better if `DapRole` has a method `getID` that returns its id.

mendess: replaced the table with this
```rs
        fn role_to_id(role: DapRole) -> u8 {
            match role {
                DapRole::Collector => 0,
                DapRole::Client => 1,
                DapRole::Leader => 2,
                DapRole::Helper => 3,
            }
        }
```

armfazh: This looks better and simpler

armfazh (approved):

PR #729, opened 2024-11-28
Improve type safeness of InitializedReport

mendess:
InitializedReport has a `Option<Vec<u8>>` field which is only set by the helper, but this fact is known at compile time, it doesn't need to be encoded in a runtime option.

By replacing `Option<Vec<u8>>` with a generic that can either be `()` or `WithPeerPrepShare` we make it explicit which parts of the code are dealing with reports from a client and which parts are dealing with reports from a peer (leader).

In other words, we lift an invariant that was held by a comment
```rs
        // Set by the Helper.
        peer_prep_share: Option<Vec<u8>>
```
into one upheld by the compiler.

And finally instead of having on `new` constructor we have `from_client` and `from_leader` to make it obvious what the intent is when constructing, which in turn results in different sets of arguments being required (the former doesn't require a peer_prep_share where the latter does)

Other advantages of this change are:
- we no longer make clippy [angry](https://github.com/cloudflare/daphne/pull/729/files#diff-7c6740c5b3ceb5af4a8f7daa7f7e1adec53ccf3dda8eb714c918fcd860697dbfL90) that our function has too many arguments
- we no longer need to [ignore](https://github.com/cloudflare/daphne/pull/729/files#diff-7c6740c5b3ceb5af4a8f7daa7f7e1adec53ccf3dda8eb714c918fcd860697dbfL364) the `Some` case in leader code
- we no longer need to [handle](https://github.com/cloudflare/daphne/pull/729/files#diff-7c6740c5b3ceb5af4a8f7daa7f7e1adec53ccf3dda8eb714c918fcd860697dbfL584-L588) the `None` case in helper code
- we can no longer write tests that [give the wrong idea of what the correct set of arguments are](https://github.com/cloudflare/daphne/pull/729/files#diff-878e5eb58e035454910e0cbc294154bc57b38ccd7b9d784e375f7e03c913fe22R136-R151) since it was incorrect to call `new` with the `helper` role and pass `None` to the peer_prep_share argument, but this test did it anyway

### DIFF
--- a/crates/daphne/src/protocol/mod.rs
+++ b/crates/daphne/src/protocol/mod.rs
@@ -1,15 +1,69 @@
 // Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
 // SPDX-License-Identifier: BSD-3-Clause
 
+use prio::codec::{CodecError, Decode as _};
+use std::{collections::HashSet, io::Cursor};
+
 pub(crate) mod aggregator;
 mod client;
 mod collector;
+pub(crate) mod report_init;
+
+/// checks if an iterator has no duplicate items, returns the ok if there are no dups or an error
+/// with the first offending item.
+fn no_duplicates<I>(iterator: I) -> Result<(), I::Item>
+where
+    I: Iterator,
+    I::Item: Eq + std::hash::Hash,
+{
+    let (lower, upper) = iterator.size_hint();
+    let mut seen = HashSet::with_capacity(upper.unwrap_or(lower));
+
+    for item in iterator {
+        if let Some(repeat) = seen.replace(item) {
+            return Err(repeat);
+        }
+    }
+    Ok(())
+}
+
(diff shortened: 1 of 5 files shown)


crates/daphne/src/protocol/mod.rs
```diff
@@ -1,23 +1,76 @@
...
 mod client;
 mod collector;
+pub(crate) mod report_init;
+
+/// checks if an iterator has no duplicate items, returns the ok if there are no dups or an error
+/// with the first offending item.
+fn no_duplicates<I>(iterator: I) -> Result<(), I::Item>
+where
+    I: Iterator,
+    I::Item: Eq + std::hash::Hash + Copy,
```
armfazh, line 17: is it possible to remove the `Copy` constraint?

armfazh (approved):

crates/daphne/src/protocol/mod.rs
mendess: Removed

PR #732, opened 2024-12-02
Remove per-task HPKE config

jhoyla:
In draft-12 per-task HPKE configs were removed. The aggregator can simply return a list of configs in preference order.


mendess (changes requested):
LGTM. Please leave a TODO(draft09) in the `DapAggregator::get_hpke_config_for` to remove the `task_id` parameter

mendess (approved):

PR #734, opened 2024-12-09
Bump the version of VDAF used to v13

jhoyla:


crates/daphne/src/vdaf/prio3_latest.rs
```diff
@@ -0,0 +1,816 @@
...
+        VDAF_TYPE_PRIO3_SUM_VEC_FIELD64_MULTIPROOF_HMAC_SHA256_AES128,
+        SumVec::new(bits, length, chunk_length).map_err(|e| {
+            VdafError::Dap(fatal_error!(
+                err = ?e,
+                "failed to create sum vec from bits({bits}), length({length}), chunk_length({chunk_length})"
+            ))
+        })?,
+    )
+    .map_err(|e| VdafError::Dap(fatal_error!(err = ?e, "failed to create prio3")))
+}
```
cjpatton, line 54: Let's remove `Prio3LatestSumVecField64MultiproofHmacSha256Aes128`, since we don't need it. Note that we should expect the test vectors to fail.

crates/daphne/src/vdaf/prio2.rs
```diff
@@ -8,7 +8,7 @@ use crate::{
     fatal_error, vdaf::VdafError, DapAggregateResult, DapMeasurement, VdafAggregateShare,
     VdafPrepShare, VdafPrepState, VdafVerifyKey,
 };
-use prio::{
+use prio_09::{
```
cjpatton, line 11: I believe we could use prio_latest here.

crates/daphne/src/vdaf/mod.rs
```diff
@@ -36,9 +43,13 @@ pub use self::mastic::MasticWeightConfig;
 #[derive(Debug, thiserror::Error)]
 pub(crate) enum VdafError {
     #[error("{0}")]
-    Codec(#[from] CodecError),
+    Codec(#[from] prio_09::codec::CodecError),
+    #[error("{0}")]
+    Vdaf(#[from] prio_09::vdaf::VdafError),
+    #[error("{0}")]
+    CodecLatest(#[from] prio_latest::codec::CodecError),
```
cjpatton, line 50: I see the logic in this, but adding these variants makes the code harder to read. Two suggestions:
* Add comments explaining the purpose of the variants
* Our naming convention should be that the legacy thing gets an odd name, and the new thing gets the normal name. For example:
```suggestion
    CodecDraft09(#[from] prio_09::codec::CodecError),
    #[error("{0}")]
    VdafDraft09(#[from] prio_09::vdaf::VdafError),
    #[error("{0}")]
    Codec(#[from] prio_latest::codec::CodecError),
```
Optionally, you might add "Latest" the latest variants.

crates/daphne/src/vdaf/mod.rs
```diff
@@ -195,6 +220,15 @@ pub enum VdafPrepState {
     Prio3Field64(Prio3PrepareState<Field64, 16>),
     Prio3Field64HmacSha256Aes128(Prio3PrepareState<Field64, 32>),
     Prio3Field128(Prio3PrepareState<Field128, 16>),
+    Prio3LatestField64(
```
cjpatton, line 223: Likewise here: the old thing should get the more specific name.

```suggestion
    Prio3Field64(
```

Cargo.toml
```diff
@@ -60,7 +60,8 @@ mappable-rc = "0.1.1"
 matchit = "0.7.3"
 p256 = { version = "0.13.2", features = ["ecdsa-core", "ecdsa", "pem"] }
 paste = "1.0.15"
-prio = "0.16.7"
+prio_09 = { package = "prio", version = "0.16.7" }
+prio_latest = { package = "prio", git = "https://github.com/divviup/libprio-rs.git", branch = "main" }
```
cjpatton, line 64: Naming convention: Old thing gets the weird name; new thing gets the normal name.

```suggestion
prio_draft09 = { package = "prio", version = "0.16.7" }
prio = { package = "prio", git = "https://github.com/divviup/libprio-rs.git", branch = "main" }
```

crates/daphne/src/hpke.rs
```diff
@@ -444,7 +444,7 @@ pub mod info_and_aad {
         messages::{encode_u32_bytes, encode_u32_prefixed, BatchSelector, ReportMetadata, TaskId},
         DapAggregationParam, DapVersion,
     };
-    use prio::codec::{CodecError, Encode, ParameterizedEncode};
+    use prio_09::codec::{CodecError, Encode, ParameterizedEncode};
```
cjpatton, line 447: You should be able to use the latest draft here.

crates/daphne/src/lib.rs
cjpatton, line 1: Can you use the latest draft here?

crates/daphne/src/messages/mod.rs
cjpatton, line 1: Can you use the latest draft here?

crates/daphne/src/messages/taskprov.rs
cjpatton, line 1: You should be able to use the latest draft here.

crates/daphne/src/roles/aggregator.rs
cjpatton, line 1: You should be able to use the latest draft here.

cjpatton (commented):
This actually isn't too bad. There are some places where macros might help with de-duplication. When @mendess has a look, he might have some ideas.

Once you make the following changes, please mark as ready for review:

* Please rebase the PR.
* I commented on a few modules where I think you can probably use the latest version of `prio` rather than the old one. In fact, I'm fairly certain the only place you can't is `pine`. Please use the latest version wherever possible.
* Naming convention: When there are two versions of a type, we generally follow the convention that the latest version gets the "normal" name and the old version gets a distinguished name. E.g., `thing` and `thing_draft09` rather than `thing_latest` and `thing`. This makes it easier to eventually remove the old version.

Cargo.toml
```diff
@@ -56,11 +56,13 @@ hpke-rs = "0.2.0"
 hpke-rs-crypto = "0.2.0"
 hpke-rs-rust-crypto = "0.2.0"
 http = "1"
+js-sys = "=0.3.66"
```
cjpatton, line 59: Why add this dependency? And why pin the version?

Cargo.toml
```diff
@@ -85,7 +87,8 @@ tracing-core = "0.1.32"
 tracing-subscriber = "0.3.18"
 url = { version = "2.5.2", features = ["serde"] }
 webpki = "0.22.4"
-worker = { version = "0.4", features = ["http"] }
+worker = { version = "=0.3.3", features = ["http"] }
+worker-sys = "=0.3.3"
```
cjpatton, line 91: I think Machete is complaining in CI that this dependency is set but not used.

Cargo.toml
cjpatton: FWIW, pinning versions can downgrade other dependencies, which could explain the bug you're seeing when you run the tests locally.

jhoyla: This was a (vain) attempt to get the tests working again, as the bug appears more often in the newer version of workers.

crates/daphne/src/vdaf/prio3_latest.rs
jhoyla: IIUC we do in fact still need this.

crates/daphne-worker-test/Cargo.toml
```diff
@@ -25,6 +25,5 @@ daphne-worker = { path = "../daphne-worker", features = ["test-utils"] }
 prometheus.workspace = true
 tracing.workspace = true
 worker.workspace = true
-
```
cjpatton, line 28: nit: Revert this change

crates/daphne-service-utils/src/durable_requests/durable_request.capnp
```diff
@@ -35,10 +35,13 @@ struct DapAggregateShare @0xb34ce529a4a66aed {
...
-        field64 @4 :Data;
-        field128 @5 :Data;
-        fieldPrio2 @6 :Data;
+        field64Draft09@4 :Data;
+        field128Draft09@5 :Data;
+        fieldPrio2Draft09@6 :Data;
         none @7 :Void;
+        field64 @8 :Data;
+        field128 @9 :Data;
+        fieldPrio2 @10 :Data;
```
cjpatton, line 44: @mendess can you double check this renaming is safe? Not sure if the field can be renamed w/o side-effects to the serialization format.

crates/daphne/src/lib.rs
```diff
@@ -814,7 +819,25 @@ impl deepsize::DeepSizeOf for DapAggregationParam {
...
+        match self {
+            Self::Empty => Some(0),
+            #[cfg(feature = "experimental")]
+            Self::Mastic(agg_param) => agg_param.encoded_len(),
+        }
+    }
+}
+
+impl EncodeDraft09 for DapAggregationParam {
+    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecErrorDraft09> {
```
cjpatton, line 840: This is an ideal use case for macros:

```rust
macro_rules! impl_encode_for_agg_param {
    ($encode_trait:ident, $codec_error:ident) => {
        impl $encode_trait for DapAggregationParam {
            fn encode(&self, _bytes: &mut Vec<u8>) -> Result<(), $codec_error> {
                match self {
                    Self::Empty => Ok(()),
                    #[cfg(feature = "experimental")]
                    Self::Mastic(_) => Err($codec_error::UnexpectedValue),
                }
            }

            fn encoded_len(&self) -> Option<usize> {
                match self {
                    Self::Empty => Some(0),
                    #[cfg(feature = "experimental")]
                    Self::Mastic(agg_param) => agg_param.encoded_len(),
                }
            }
        }
    };
}
impl_encode_for_agg_param!(Encode, CodecError);
impl_encode_for_agg_param!(EncodeDraft09, CodecErrorDraft09);
```

Likewise for `ParameterizedDecode` below.

crates/daphne/src/protocol/mod.rs
```diff
@@ -81,18 +81,27 @@ mod test {
...
         },
     };
     use rand::prelude::*;
     use std::iter::zip;
 
     const TEST_VDAF: &VdafConfig = &VdafConfig::Prio3(Prio3Config::Count);
+    const TEST_VDAF_DRAFT09: &VdafConfig = &VdafConfig::Prio3Draft09(Prio3Config::Count);
 
     fn roundtrip_report(version: DapVersion) {
+        const CTX_STRING: &[u8] = "dap-13".as_bytes();
```
cjpatton, line 104: nit: The "context string" is "dap-13" followed by the task ID. I'd refer to this as the context string prefix. However since this is test code, I would probably wait to construct the context string until the task ID is known. That is, on line 181 below:
```rust
let ctx = ["dap-13", &.task_id.0].concat();
```

crates/daphne/src/protocol/mod.rs
```diff
@@ -595,6 +713,51 @@ mod test {
...
+
+        let vdaf = Prio3Draft09::new_count(2).unwrap();
+        assert_eq!(
+            vdaf.unshard(&(), [leader_agg_share, helper_agg_share], num_measurements,)
+                .unwrap(),
+            3,
+        );
+    }
+
+    test_versions! { finish_agg_job_vdaf_draft09 }
```
cjpatton, line 759: This macro creates two tests, one for draft09 and one for the latest draft. Let's just make `finish_agg_job_vdaf_draft09` a test.

crates/daphne/src/vdaf/mastic.rs
cjpatton, line 1: Let's bump Mastic to draft 13.

cjpatton, line 1: Let's use `prio` for this module instead of `prio_draft09`.

crates/daphne/src/vdaf/mod.rs
```diff
@@ -33,8 +52,14 @@ use std::io::Read;
 #[cfg(feature = "experimental")]
 pub use self::mastic::MasticWeightConfig;
 
+const CTX_STRING: &[u8] = "dap-13".as_bytes();
```
cjpatton, line 55:
```suggestion
const CTX_STRING_PREFIX: &[u8] = "dap-13".as_bytes();
```

crates/daphne/src/vdaf/mod.rs
```diff
@@ -411,10 +512,13 @@ pub(crate) fn decode_field_vec<F: FieldElement>(
...
 fn shard_then_encode<V: Vdaf + Client<16>>(
     vdaf: &V,
+    task_id: TaskId,
     measurement: &V::Measurement,
     nonce: &[u8; 16],
 ) -> Result<(Vec<u8>, [Vec<u8>; 2]), VdafError> {
-    let (public_share, input_shares) = vdaf.shard(measurement, nonce)?;
+    let binding = [CTX_STRING, &task_id.0].concat();
+    let ctx = binding.as_slice();
+    let (public_share, input_shares) = vdaf.shard(ctx, measurement, nonce)?;
```
cjpatton, line 521: Save an allocation

```suggestion
    let mut ctx = [0; CTX_STRING.len() + 32];
    ctx[..CTX_STRING.len()].copy_from_slice(CTX_STRING);
    ctx[CTX_STRING.len()..].copy_from_slice(&task_id.0);
    let (public_share, input_shares) = vdaf.shard(&ctx, measurement, nonce)?;

```

crates/daphne/src/vdaf/mod.rs
```diff
@@ -444,8 +549,12 @@ where
     // Decode the Helper's inbound message.
     let peer_share = V::PrepareShare::get_decoded_with_param(&host_state, peer_share_data)?;
```
cjpatton, line 551: Likewise here: might as well save an allocation

crates/daphne/src/vdaf/mod.rs
```diff
@@ -475,8 +585,10 @@ where
     // Decode the inbound message from the peer, which contains the preprocessed prepare message.
     let peer_message = V::PrepareMessage::get_decoded_with_param(&host_state, peer_message_data)?;
 
+    let binding = [CTX_STRING, &task_id.0].concat();
+    let ctx = binding.as_slice();
```
cjpatton, line 589: And here.

crates/daphne/src/vdaf/prio3_latest.rs
```diff
@@ -0,0 +1,500 @@
+// Copyright (c) 2022 Cloudflare, Inc. All rights reserved.
```
cjpatton, line 1: It is the year of our lord 2024

```suggestion
// Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
```

crates/daphne/src/vdaf/prio3_latest.rs
cjpatton, line 1: Let's just call this file `prio3.rs`?

crates/daphne/src/vdaf/prio3_latest.rs
```diff
@@ -0,0 +1,500 @@
...
+// SPDX-License-Identifier: BSD-3-Clause
+
+//! Parameters for the [Prio3 VDAF](https://datatracker.ietf.org/doc/draft-patton-cfrg-vdaf/).
```
cjpatton, line 4: This link needs to be updated, here and in `prio3_draft09.rs`

crates/daphne/src/vdaf/prio3_latest.rs
```diff
@@ -0,0 +1,500 @@
...
+    codec::ParameterizedDecode,
+    flp::Type,
+    vdaf::{
+        prio3::{Prio3, Prio3InputShare, Prio3PrepareShare, Prio3PrepareState, Prio3PublicShare},
+        xof::Xof,
+        Aggregator,
+    },
+};
+
+const ERR_FIELD_TYPE: &str = "unexpected field type for step or message";
```
cjpatton, line 26: Let's just inline this since we didn't end up using string constants for errors in this file.

crates/daphne/src/vdaf/prio3_latest.rs
```diff
@@ -0,0 +1,500 @@
...
+const CTX_STRING: &[u8] = "dap-13".as_bytes();
```
cjpatton, line 27:
```suggestion
const CTX_STRING_PREFIX: &[u8] = "dap-13".as_bytes();
```

crates/daphne/src/vdaf/prio3_latest.rs
```diff
@@ -0,0 +1,500 @@
...
+    {
+        // Parse the public share.
+        let public_share = Prio3PublicShare::get_decoded_with_param(&vdaf, public_share_data)?;
+
+        // Parse the input share.
+        let input_share =
+            Prio3InputShare::get_decoded_with_param(&(&vdaf, agg_id), input_share_data)?;
+
+        let binding = [CTX_STRING, &task_id.0].concat();
+        let ctx = binding.as_slice();
```
cjpatton, line 218: Save an allocation as suggested previously

crates/daphne/src/vdaf/prio3_latest.rs
```diff
@@ -0,0 +1,500 @@
...
+            let (out_share, outbound) =
+                prep_finish_from_shares(&vdaf, task_id, agg_id, state, share, peer_share_data)?;
+            let agg_share = VdafAggregateShare::Field128(vdaf.aggregate(&(), [out_share])?);
+            (agg_share, outbound)
+        }
+        (
+            Prio3Config::Sum { .. },
+            VdafPrepState::Prio3Field128(_),
+            VdafPrepShare::Prio3Field128(_),
+        ) => Err(VdafError::Dap(fatal_error!(err = "sum unimplemented")))?,
```
cjpatton, line 277: For consistency with error handling elsewhere.
```suggestion
        ) => Err(VdafError::Dap(fatal_error!(err = "Prio3Sum is not supported in VDAF-13")))?,
```

crates/daphne-service-utils/src/durable_requests/durable_request.capnp
jhoyla: From the Cap'n Proto docs[1]
```
Any symbolic name can be changed, as long as the type ID / ordinal numbers stay the same. Note that type declarations have an implicit ID generated based on their name and parent’s ID, but you can use capnp compile -ocapnp myschema.capnp to find out what that number is, and then declare it explicitly after your rename.
```
[1] https://capnproto.org/language.html#evolving-your-protocol

crates/daphne/src/vdaf/mastic.rs
jhoyla: I thought Mastic wasn't yet defined for draft-13? Also, do you want backwards compatibility with draft-09?

cjpatton: Sorry, it sounds like there was a bit of a misscomunication. Mastic is not implemented, but when it is, it will be compatible with VDAF-13. Pine on the other hand is fully implemented but not compatible with VDAF-13.

crates/daphne/src/lib.rs
cjpatton: Ohhhh I see the output for the Mastic variant is different in each case. What I suggest we do is just return `Ok(Self::Mastic(Poplar1AggregationParam::decode(bytes)?))` in both instances.

jhoyla: Not the experimental branch?

cjpatton: You still need `#[cfg(feature = "experimental")]` because Mastic is only compiled when the "experimental" feature is enabled.

crates/daphne/src/protocol/mod.rs
```diff
@@ -199,6 +209,113 @@ mod test {
 
     test_versions! { roundtrip_report }
 
+    fn roundtrip_report_vdaf_draft09(version: DapVersion) {
```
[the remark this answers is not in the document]
jhoyla: It's the version of the VDAF, not the version of DAP.

crates/daphne/src/vdaf/mod.rs
```diff
@@ -48,6 +73,7 @@ pub(crate) enum VdafError {
 #[serde(rename_all = "snake_case")]
 #[cfg_attr(any(test, feature = "test-utils"), derive(deepsize::DeepSizeOf))]
 pub enum VdafConfig {
+    Prio3Draft09(Prio3Config),
     Prio3(Prio3Config),
     Prio2 {
```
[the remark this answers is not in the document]
jhoyla: For flow control e.g.
```rust
let res = match &task_config.vdaf {
            VdafConfig::Prio3Draft09(ref prio3_config) => prio3_draft09_prep_init(
                prio3_config,
                &task_config.vdaf_verify_key,
                agg_id,
                &report_share.report_metadata.id.0,
                &report_share.public_share,
                &input_share,
            ),
            VdafConfig::Prio3(ref prio3_config) => prio3_prep_init(
                prio3_config,
                &task_config.vdaf_verify_key,
                *task_id,
                agg_id,
                &report_share.report_metadata.id.0,
                &report_share.public_share,
                &input_share,
            ),
            [...]

File: crates/daphne/src/vdaf/prio3.rs [hunk 26]
```diff
@@ -220,55 +175,36 @@ pub(crate) fn prio3_prep_init(
...
-    type Prio3Prepared<T, const SEED_SIZE: usize> = (
-        Prio3PrepareState<<T as prio::flp::Type>::Field, SEED_SIZE>,
-        Prio3PrepareShare<<T as prio::flp::Type>::Field, SEED_SIZE>,
+    type Prio3LatestPrepared<T, const SEED_SIZE: usize> = (
+        Prio3PrepareState<<T as Type>::Field, SEED_SIZE>,
+        Prio3PrepareShare<<T as Type>::Field, SEED_SIZE>,
     );
 
-    fn prep_init<T, P, const SEED_SIZE: usize>(
+    fn prep_init_latest<T, P, const SEED_SIZE: usize>(
```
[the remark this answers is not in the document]
jhoyla: They don't actually clash. They can both be `prep_init` given that they're local only.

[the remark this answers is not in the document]
jhoyla: Actually, that gives annoyingly inconsistent naming. I'm going with your suggestion.

crates/daphne/src/lib.rs
```diff
@@ -815,7 +820,7 @@ impl deepsize::DeepSizeOf for DapAggregationParam {
 
 impl Encode for DapAggregationParam {
     fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
-        let _ = bytes;
+        _ = bytes;
```
cjpatton, line 823: nit: Revert this change if it's not related to this PR and the compiler doesn't require it.

crates/daphne/src/protocol/aggregator.rs
```diff
@@ -352,11 +365,16 @@ impl DapTaskConfig {
...
-                            Err(e @ (VdafError::Codec(..) | VdafError::Vdaf(..))) => {
+                            Err(e @ (VdafError::CodecDraft09(..) | VdafError::VdafDraft09(..))) => {
                                 tracing::warn!(error = ?e, "rejecting report");
                                 TransitionVar::Failed(TransitionFailure::VdafPrepError)
                             }
 
+                            Err(e @ (VdafError::Codec(..) | VdafError::Vdaf(..))) => {
+                                tracing::warn!(error = ?e, "rejecting report - latest");
+                                TransitionVar::Failed(TransitionFailure::VdafPrepError)
+                            }
```
cjpatton, line 376: I don't think we need to distinguish between latest and draft 09 here.

Likewise below.

crates/daphne/src/protocol/mod.rs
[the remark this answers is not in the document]
cjpatton: If that's the case, then this test only applies to DAP-09 as well, correct?

We coudl remove the test_versions! macro and just make this function a test.

crates/daphne/src/vdaf/mastic.rs
```diff
@@ -86,13 +86,13 @@ pub(crate) fn mastic_prep_init(
             // Simulate Mastic, insecurely. The public share encodes the plaintext input; the input
             // share encodes the plaintext weight.
             if public_share_bytes.len() != input_size {
-                return Err(VdafError::Codec(prio::codec::CodecError::Other(
+                return Err(VdafError::Codec(CodecError::Other(
```
cjpatton, line 89: Revert this change, since it's not necessary. Likewise below.

crates/daphne/src/vdaf/mastic.rs
```diff
@@ -149,13 +149,13 @@ pub(crate) fn mastic_prep_finish_from_shares(
             // valid. This is not secure because the weight is revealed to the caller.
             let peer_weight = Field64::get_decoded(peer_share_bytes)?;
             if peer_weight != host_weight {
-                return Err(VdafError::Vdaf(prio::vdaf::VdafError::Uncategorized(
+                return Err(VdafError::Vdaf(PrioVdafError::Uncategorized(
```
cjpatton, line 152: Revert the change. By the way, we included the `prio::vdaf::` prefix here so that we didn't have to rename this type. I generally try to avoid `use foo::Bar as FooBar` pattern because it creates a degree of indirection that makes the code harder to read.

cjpatton (commented):
Just some nits

crates/daphne/src/protocol/aggregator.rs
```diff
@@ -352,6 +365,11 @@ impl DapTaskConfig {
...
                             }
 
+                            Err(e @ (VdafError::CodecDraft09(..) | VdafError::VdafDraft09(..))) => {
+                                tracing::warn!(error = ?e, "rejecting report");
+                                TransitionVar::Failed(TransitionFailure::VdafPrepError)
+                            }
+
                             Err(e @ (VdafError::Codec(..) | VdafError::Vdaf(..))) => {
                                 tracing::warn!(error = ?e, "rejecting report");
                                 TransitionVar::Failed(TransitionFailure::VdafPrepError)
```
cjpatton, line 375: Likewise below.

```suggestion
                            Err(
                                e @ (VdafError::CodecDraft09(..)
                                | VdafError::VdafDraft09(..)
                                | VdafError::Codec(..)
                                | VdafError::Vdaf(..)),
                            ) => {
                                tracing::warn!(error = ?e, "rejecting report");
                                TransitionVar::Failed(TransitionFailure::VdafPrepError)
                            }
```

cjpatton (approved):
One last minor comment. Note that I've tacked on a couple of commits using macros to de-duplicate some code.

Before merging:
* Looks like you'll need to rebase.
* Please squash all commits into one.
* Make sure @mendess approves. This is a big enough change that we should have a couple of eyes on it.

crates/daphne/src/vdaf/mod.rs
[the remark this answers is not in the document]
mendess: you have access to the dap version here, can't you use that to decide which function to call?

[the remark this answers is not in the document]
jhoyla: So I don't want to tightly couple the DAP version to the Prio verison unnecessarily. I did have a brief go at it, but you end up having to pass the `DapVersion` around all over the place. Given that the tests pass with all four combinations this seems (marginally) cleaner.

mendess (approved):
It's gonna be a pain to work with this now, but such is life. LGTM

PR #736, opened 2024-12-16
Bump VDAF to draft 13 for Prio2

jhoyla:
Prio2 can use draft-13 or draft-09. This PR adds backwards compatible support for the latest draft.
It also contains a few missing tests for Prio3's backwards compatible support.

### DIFF
--- a/crates/daphne/src/roles/mod.rs
+++ b/crates/daphne/src/roles/mod.rs
@@ -198,7 +198,7 @@ mod test {
             };
 
             // Task Parameters that the Leader and Helper must agree on.
-            let vdaf_config = VdafConfig::Prio3Draft09(Prio3Config::Count);
+            let vdaf_config = VdafConfig::Prio3(Prio3Config::Count);
             let leader_url = Url::parse("https://leader.com/v02/").unwrap();
             let helper_url = Url::parse("http://helper.org:8788/v02/").unwrap();
             let collector_hpke_receiver_config =
@@ -486,10 +486,7 @@ mod test {
             // Construct report. We expect the VDAF to be Prio3Count so that we know what type of
             // measurement to generate. However, we could extend the code to support more VDAFs.
             let task_config = self.leader.unchecked_get_task_config(task_id).await;
-            assert_matches!(
-                task_config.vdaf,
-                VdafConfig::Prio3Draft09(Prio3Config::Count)
-            );
+            assert_matches!(task_config.vdaf, VdafConfig::Prio3(Prio3Config::Count));
 
             self.gen_test_report_for_measurement(task_id, DapMeasurement::U64(1))
                 .await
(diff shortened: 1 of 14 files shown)


crates/daphne/src/roles/mod.rs
```diff
@@ -1516,7 +1513,18 @@ mod test {
...
+        )
+        .await;
+    }
+
+    async_test_versions! { e2e_taskprov_prio2_draft09 }
+
+    async fn e2e_taskprov_prio2_latest(version: DapVersion) {
         e2e_taskprov(
             version,
             VdafConfig::Prio2 { dimension: 10 },
```
mendess, line 1530: remerge these two tests and use the passed in version to determine the prio version to use

crates/daphne/src/roles/mod.rs
```diff
@@ -1541,6 +1551,22 @@ mod test {
...
+                bits: 1,
+                length: 10,
+                chunk_length: 2,
+                num_proofs: 3,
+            }),
+            DapMeasurement::U64Vec(vec![1; 10]),
+        )
+        .await;
+    }
+
```
mendess, line 1569: Same here

mendess (changes requested):

cjpatton:
I'm waiting to review until the base PR has been merged and this PR has been rebased.

crates/daphne/src/vdaf/prio2_latest.rs
```diff
@@ -0,0 +1,200 @@
+// Copyright (c) 2022 Cloudflare, Inc. All rights reserved.
```
mendess, line 1: rename file to just prio2.rs

mendess (approved):

crates/daphne/src/roles/mod.rs
jhoyla: I don't want to tie Prio too tightly to DAP, especially in the tests.

crates/daphne/src/vdaf/prio2.rs
```diff
@@ -71,7 +77,10 @@ pub(crate) fn prio2_prep_init(
     <()>::get_decoded_with_param(&vdaf, public_share_data)?;
     let input_share: Share<FieldPrio2, 32> =
         Share::get_decoded_with_param(&(&vdaf, agg_id), input_share_data)?;
-    let (state, share) = vdaf.prepare_init(verify_key, agg_id, &(), nonce, &(), &input_share)?;
+    let binding = [CTX_STRING, &task_id.0].concat();
+    let ctx = binding.as_slice();
+    let (state, share) =
+        vdaf.prepare_init(verify_key, ctx, agg_id, &(), nonce, &(), &input_share)?;
```
cjpatton, line 83: Hmmm ... there are no functional changes to Prio2 between `prio_draft09` and `prio`. The implementation will simply ignore the context string, so there's not much point in constructing it here.

Stepping back, I don't think we need two version so the `prio2` module, as there is only one version of this algorithm we need to support (either in DAP-09 or DAP-13).

FYI, in case this wasn't clear: Prio2 is not standardized, but is provided by libprio. It hasn't changed at all since DAP-02.

cjpatton (changes requested):

PR #737, opened 2024-12-16
Rename transition failure to report error

jhoyla:
In draft 13 of the DAP spec `PrepareError` (which we called `TransitionFailure`) was renamed to `ReportError`. This PR brings us in line with the spec.

### DIFF
--- a/crates/daphne-server/src/router/mod.rs
+++ b/crates/daphne-server/src/router/mod.rs
@@ -160,9 +160,9 @@ impl AxumDapResponse {
     }
 
     pub fn new_error<E: Into<DapError>>(error: E, metrics: &dyn DaphneServiceMetrics) -> Self {
-        // trigger abort if transition failures reach this point.
+        // Trigger abort if report errors reach this point.
         let error = match error.into() {
-            DapError::Transition(failure) => DapAbort::report_rejected(failure),
+            DapError::ReportError(err) => DapAbort::report_rejected(err),
             DapError::Fatal(e) => Err(e),
             DapError::Abort(abort) => Ok(abort),
         };
--- a/crates/daphne/src/error/mod.rs
+++ b/crates/daphne/src/error/mod.rs
@@ -5,7 +5,7 @@ pub mod aborts;
 
 use std::fmt::{Debug, Display};
 
-use crate::{messages::TransitionFailure, vdaf::VdafError};
+use crate::{messages::ReportError, vdaf::VdafError};
 pub use aborts::DapAbort;
 use prio::codec::CodecError;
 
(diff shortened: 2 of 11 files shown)


crates/daphne/src/error/mod.rs
```diff
@@ -25,10 +25,10 @@ pub enum DapError {
     #[error("abort: {0}")]
     Abort(#[from] DapAbort),
 
-    /// Transition failure. This error blocks processing of a paritcular report and may, under
+    /// Report Error. This error blocks processing of a paritcular report and may, under
     /// certain conditions, trigger an abort.
-    #[error("transition error: {0}")]
-    Transition(#[from] TransitionFailure),
+    #[error("report error: {0}")]
+    Transition(#[from] ReportError),
```
mendess, line 31: Should this variant also no be renamed?

mendess (approved):

crates/daphne-server/src/router/mod.rs
```diff
@@ -160,9 +160,9 @@ impl AxumDapResponse {
     }
 
     pub fn new_error<E: Into<DapError>>(error: E, metrics: &dyn DaphneServiceMetrics) -> Self {
-        // trigger abort if transition failures reach this point.
+        // trigger abort if report errors reach this point.
```
cjpatton, line 163: Might as well fix grammar here.
```suggestion
        // Trigger abort if report errors reach this point.
```

crates/daphne/src/messages/mod.rs
```diff
@@ -760,7 +760,7 @@ impl TryFrom<u8> for TransitionFailureDraft09 {
     }
 }
 
-impl TryFrom<u8> for TransitionFailureLatest {
+impl TryFrom<u8> for ReportErrorLatest {
```
cjpatton, line 763:
```suggestion
impl TryFrom<u8> for ReportError {
```

cjpatton (changes requested):
Just a couple minor things.

crates/daphne/src/messages/mod.rs
jhoyla: So the unified enum is called `ReportError` which I think makes the most sense. We could rename the unified enum to `UnifiedReportError` but I think that makes the code less readable. Alternatively I could rename `ReportErrorLatest` to `ReportErrorDraft13` (or maybe `ReportErrorDraft12` as that was when the renumbering happened) just so that it will stay aligned with the drafts.

cjpatton: Nope, better as is!

cjpatton (approved):

PR #739, opened 2024-12-17
pine: Add encoding for `PinePrepState`

cjpatton:
Decoding requires the instance of `Pine` and an indication of whether the aggregator is the leader or helper. Note that the encoding is not standardized anywhere.


mendess (approved):
You accidentally created a vdaf.rs file at the root of the repo

PR #742, opened 2024-12-18
taskprov: Ensure the task ID matches the advertised config

### LINKED ISSUE #710: Properly opt out of taskprov tasks
Discussion: https://github.com/cloudflare/daphne/pull/709#discussion_r1823230576

Relevant spec: https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap-taskprov/draft-ietf-ppm-dap-taskprov.html#section-4.4-3.1.1

We don't reject when a task id does not match the task config

cjpatton:
Closes #710.

Add a unit test that ensures
`TaskprovAdvertisement::parse_taskprov_advertisement()` fails if the task ID doesn't match the advertised task config. Along the way, adjust some variable names to make the code easier to read.

### DIFF
--- a/crates/daphne/src/taskprov.rs
+++ b/crates/daphne/src/taskprov.rs
@@ -120,7 +120,7 @@ fn url_from_bytes(task_id: &TaskId, url_bytes: &[u8]) -> Result<Url, DapAbort> {
 }
 
 impl DapBatchMode {
-    fn try_from_taskprov(task_id: &TaskId, var: BatchMode) -> Result<Self, DapAbort> {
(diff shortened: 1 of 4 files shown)


crates/daphne/src/taskprov.rs
```diff
@@ -290,21 +290,25 @@ impl DapTaskConfigNeedsOptIn {
     pub(crate) fn try_from_taskprov(
```
mendess, line 290: maybe this function should be named `try_from_taskprov_advertisement`

mendess (approved):

cjpatton:
Took Mendes' suggestion, rebased, and squashed.

PR #745, opened 2024-12-19
Bump Prio2 to VDAF-13 Take #2

jhoyla:


mendess (approved):

crates/daphne/src/vdaf/prio2.rs
```diff
@@ -71,7 +71,8 @@ pub(crate) fn prio2_prep_init(
     <()>::get_decoded_with_param(&vdaf, public_share_data)?;
     let input_share: Share<FieldPrio2, 32> =
         Share::get_decoded_with_param(&(&vdaf, agg_id), input_share_data)?;
```
cjpatton, line 73: I'd add a comment here nothing that the `ctx` string is ignored by Prio2.

crates/daphne/src/vdaf/prio2.rs
```diff
@@ -133,7 +134,7 @@ pub(crate) fn prio2_prep_finish(
             )))
         }
     };
-    let agg_share = VdafAggregateShare::Field32Draft09(vdaf.aggregate(&(), [out_share])?);
```
cjpatton, line 136: We should be able to remove the `Field32Draft09` variant now.

cjpatton (changes requested):
Whoops, I meant to request changes, not just comment.

crates/daphne/src/vdaf/prio2.rs
jhoyla: Doesn't `Pine32HmacSha256Aes128` depend on it?

PR #747, opened 2024-12-19
Refactors to prepare for moving the aggregation logic back into the workers

mendess:


crates/daphne/src/vdaf/mod.rs
```diff
@@ -228,6 +229,28 @@ impl AsMut<[u8]> for VdafVerifyKey {
     }
 }
 
+fn upgrade_codec_error(error: CodecErrorDraft09) -> CodecError {
```
jhoyla, line 232: There is already a `from_codec_error` on line 104 which implements exactly this function. These should be merged, and I wonder if `upgrade_codec_error` and `downgrade_codec_error` could be macros?

mendess: I don't know how I could make them with macros

jhoyla: So I just had a go, and you can get fairly close, but I couldn't get all the way there. 
Something like
```rust
macro_rules! make_from_codec_error {
    ($e:ident, $f: ident, $t: ident, $($variants:expr),+) => {
        fn $e(input: $f) -> $t {
            match input {
                $(
                    $f::$variants => $t::$variants,
                )*
            }
        }
    }
}
```

mendess: yeah but you end up having to write more code than if done manually, no? Either way I deleted `downgrade_*` because I realized encoding using the prio_d9 wasn't worth it anyway

crates/daphne/src/constants.rs
```diff
@@ -120,6 +120,13 @@ pub enum DapAggregatorRole {
...
+    pub fn as_aggregator_id(self) -> usize {
+        match self {
+            DapAggregatorRole::Leader => 0,
+            DapAggregatorRole::Helper => 1,
+        }
+    }
+
     pub fn is_leader(self) -> bool {
         self == Self::Leader
     }
```
cjpatton, line 132: * Do we still need is_leader() if we have as_aggregator_id()?
* Also, the new function can use a docucomment.

crates/daphne/src/vdaf/prio3.rs
```diff
@@ -4,27 +4,57 @@
...
+        VDAF_TYPE_PRIO3_SUM_VEC_FIELD64_MULTIPROOF_HMAC_SHA256_AES128,
+        SumVec::new(bits, length, chunk_length).map_err(|e| {
+            VdafError::Dap(fatal_error!(
+                err = ?e,
+                "failed to create sum vec from bits({bits}), length({length}), chunk_length({chunk_length})"
+            ))
+        })?,
+    )
+    .map_err(|e| VdafError::Dap(fatal_error!(err = ?e, "failed to create prio3")))
+}
```
cjpatton, line 56: Why add this back? This variant of Prio3 is not yet specified for VDAF-13.

crates/daphne/src/pine/vdaf.rs
```diff
@@ -981,14 +988,14 @@ mod tests {
         };
 
         let got = PinePrepState::get_decoded_with_param(
-            &(&pine, true),
+            &(&pine, DapAggregatorRole::Leader),
```
cjpatton, line 991: I'd prefer this type not bleed into the `pine` module. Ideally this module could be used externally without having to use the rest of the crate.

crates/daphne/src/protocol/report_init.rs
```diff
@@ -67,11 +76,11 @@ impl InitializedReport<()> {
 }
 
 impl InitializedReport<WithPeerPrepShare> {
-    pub fn from_leader(
+    pub fn from_leader<'s>(
         decrypter: &impl HpkeDecrypter,
         valid_report_range: Range<messages::Time>,
         task_id: &TaskId,
-        task_config: &DapTaskConfig,
+        task_config: impl Into<PartialDapTaskConfig<'s>>,
```
cjpatton, line 83: Can you remind me what "partial" here means?

crates/daphne/src/vdaf/draft09.rs
```diff
@@ -10,42 +10,62 @@ use crate::{
...
 }
 
+gen_prio3_sum_vec!(
+    prio_draft09,
+    Prio3SumVecField64MultiproofHmacSha256Aes128,
+    new_prio3_sum_vec_field64_multiproof_hmac_sha256_aes128
+);
+gen_prio3_sum_vec!(
+    prio,
+    LatestBridgePrio3SumVecField64MultiproofHmacSha256Aes128,
```
cjpatton, line 65: Why is this type needed?

Note that Prio3SumVecField64MultiproofHmacSha256Aes128 is not defined for the latest draft. It should only be available in DAP-09.

crates/daphne/src/protocol/report_init.rs
mendess: Just enough to initialize reports

crates/daphne/src/vdaf/draft09.rs
mendess: I had to introduce this because of this variant here https://github.com/cloudflare/daphne/blob/f76357e1cdab3312aadd9f3bc4bba47c43cc56e9/crates/daphne/src/vdaf/mod.rs#L444

crates/daphne/src/protocol/report_init.rs
cjpatton: Why not use the entire `DapTaskConfig`?

mendess: I will have to serialize it later and it's very big and cumbersome to serialize (manually because of capnproto)

crates/daphne/src/vdaf/draft09.rs
cjpatton: That variant of the state can be deleted: acb75781ec4f848fa7c6fe86880e3d4a49cd4c10

cjpatton (commented):
Looks good except for the two remaining comments.

crates/daphne/src/protocol/report_init.rs
mendess: How about `PartialDapTaskConfigForReportInit`?

cjpatton: Works for me!

mendess:
> Looks good except for the two remaining comments.

done

cjpatton (approved):

PR #748, opened 2024-12-19
Remove draft09 version of vdaf config

mendess:


mendess:
Closing in favor of #749

Issue #698, opened 2024-10-15
Support the next interop target


cjpatton:
As described [here](https://mailarchive.ietf.org/arch/msg/ppm/Q-uN4IAfekbciRDgyHt3tlwYSAA/), the next set of drafts to implement are (DAP-12):
* [draft-ietf-ppm-dap-12](https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/)
* [draft-irtf-ppm-dap-taskprov-01](https://datatracker.ietf.org/doc/draft-ietf-ppm-dap-taskprov/) (WIP as of 2024/10/16)
* [draft-irtf-cfrg-vdaf-12](https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/) for the Prio3
* We also might need to bump [draft-dcook-ppm-dap-interop-test-design](https://datatracker.ietf.org/doc/draft-dcook-ppm-dap-interop-test-design/), but this needs discussion.

The draft set we currently support is (DAP-09):
* draft-ietf-ppm-dap-09
* draft-irtf-cfrg-vdaf-08 for Prio3
* draft-wang-ppm-dap-taskprov-06
* draft-dcook-ppm-dap-interop-test-design-07

We will need to add support for DAP-12 alongside DAP-09.

# Requirements

1. The version of a VDAF, of taskprov, or of the interop API that is used is determined by the version of DAP.  In particular, we won't try to support multiple versions of a VDAF or of the taskprov extension with the same DAP version.
1. Our implementations of Prio2 and Prio3 come from the [prio](https://github.com/divviup/libprio-rs) crate. Version 0.16 is compatible with DAP-09; a future version of the crate, perhaps 0.17, [will be compatible with DAP-12](https://github.com/divviup/libprio-rs/issues/1122). We will need to import both versions of the crate at the same time.
1. DAP-09 MUST support all VDAFs currently supported by the taskprov extension. This includes Prio2, PINE (see below), and some non-standard Prio3SumVec variants.
1. DAP-12 MUST support all currently supported VDAFs, excluding PINE: [draft-chen-cfrg-vdaf-pine-01](https://datatracker.ietf.org/doc/draft-chen-cfrg-vdaf-pine/) is not [API compatible with DAP-12](https://github.com/junyechen1996/draft-chen-cfrg-vdaf-pine/issues/106), so for now we'll only maintain support for DAP-09.

# Notes

[...]

PR #749, opened 2024-12-19
Refactor Prio3 dispatch logic, remove unneeded variants

cjpatton:
Partially addresses #698.
Alternative to #748.

Move sharding, preparation, and unsharding dispatchers to `Prio3Config`.
Not all variants are supported by all DAP versions, so pass the version
to each dispatcher in order to resolve DAP and Prio3 version
compatibility. (Note Prio3 is not API-compatible across versions.)

The following variants are supported in DAP-13:

* Prio3Count
* Prio3Sum
* Prio3SumVec
* Prio3Histogram

The following variants are supported in DAP-09:

* Prio3SumVecField64MultiproofHmacSha256Aes128

Modify the taskprov provisioning logic accordingly, by removing support
for the following VDAFs in DAP-13:

* Prio3SumVecField64MultiproofHmacSha256Aes128
* Pine32HmacSha256Aes128
* Pine64HmacSha256Aes128

These VDAFs need to be upgraded for VDAF-13 compatibility.

### DIFF
--- a/crates/daphne-service-utils/src/test_route_types.rs
+++ b/crates/daphne-service-utils/src/test_route_types.rs
@@ -30,34 +30,44 @@ pub struct InternalTestVdaf {
     pub length: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub chunk_length: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub dimension: Option<String>,
 }
 
 impl From<VdafConfig> for InternalTestVdaf {
     fn from(vdaf: VdafConfig) -> Self {
-        let (typ, bits, length, chunk_length) = match vdaf {
-            VdafConfig::Prio3Draft09(prio3) => match prio3 {
-                Prio3Config::Count => ("Prio3Draft09Count", None, None, None),
-                Prio3Config::Sum { bits } => ("Prio3Draft09Sum", Some(bits), None, None),
+        let (typ, bits, length, chunk_length, dimension) = match vdaf {
+            VdafConfig::Prio3(prio3) => match prio3 {
+                Prio3Config::Count => ("Prio3Count", None, None, None, None),
+                Prio3Config::Sum { max_measurement } => (
+                    "Prio3Sum",
+                    Some(usize::try_from(max_measurement).unwrap()),
+                    None,
+                    None,
+                    None,
+                ),
                 Prio3Config::Histogram {
                     length,
                     chunk_length,
                 } => (
-                    "Prio3Draft09Histogram",
+                    "Prio3Histogram",
                     None,
                     Some(length),
                     Some(chunk_length),
+                    None,
                 ),
                 Prio3Config::SumVec {
                     bits,
                     length,
                     chunk_length,
                 } => (
-                    "Prio3Draft09SumVec",
+                    "Prio3SumVec",
                     Some(bits),
                     Some(length),
                     Some(chunk_length),
+                    None,
                 ),
-                Prio3Config::SumVecField64MultiproofHmacSha256Aes128 {
+                Prio3Config::Draft09SumVecField64MultiproofHmacSha256Aes128 {
                     bits,
                     length,
                     chunk_length,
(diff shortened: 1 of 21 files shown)


crates/daphne/src/protocol/mod.rs
```diff
@@ -208,113 +198,6 @@ mod test {
 
     test_versions! { roundtrip_report }
 
-    #[test]
```
cjpatton, line 211: AFAICT the only difference between this test and the previous is that the `prio_draft09` crate is used instead of `prio`. This was necessary because we needed the right version of Prio3Sum. We avoid this by using a VDAF for testing that is supported by both versions (Prio2).

crates/daphne/src/protocol/mod.rs
```diff
@@ -711,59 +613,14 @@ mod test {
 
     test_versions! { agg_job_resp_abort_invalid_transition }
 
-    #[test]
-    fn finish_agg_job_vdaf_draft09() {
-        let version = DapVersion::Draft09;
```
cjpatton, line 716: Likewise https://github.com/cloudflare/daphne/pull/749/files#r1893170180

crates/daphne/src/protocol/mod.rs
```diff
@@ -779,36 +636,40 @@ mod test {
...
     }
 
     test_versions! { finish_agg_job }
 
-    #[tokio::test]
-    async fn agg_job_init_req_skip_vdaf_prep_error_draft09() {
-        let t =
-            AggregationJobTest::new(TEST_VDAF, HpkeKemId::X25519HkdfSha256, DapVersion::Draft09);
-        let mut reports = t.produce_reports(vec![DapMeasurement::U64(1), DapMeasurement::U64(1)]);
+    fn agg_job_init_req_skip_vdaf_prep_error(version: DapVersion) {
```
cjpatton, line 661: This test isn't specific to a DAP version.

crates/daphne/src/vdaf/mod.rs
```diff
@@ -432,10 +432,7 @@ impl Encode for VdafAggregateShare {
 impl VdafConfig {
     pub(crate) fn uninitialized_verify_key(&self) -> VdafVerifyKey {
         match self {
-            Self::Prio3Draft09(Prio3Config::SumVecField64MultiproofHmacSha256Aes128 { .. })
-            | Self::Prio2 { .. }
-            | Self::Prio3(..) => VdafVerifyKey::L32([0; 32]),
-            Self::Prio3Draft09(..) => VdafVerifyKey::L16([0; 16]),
```
cjpatton, line 438: Since we dropped the old variants of Prio3, the seed size is 32 bytes for every Prio3 variant we support :)

crates/daphne/src/vdaf/mod.rs
```diff
@@ -135,8 +133,8 @@ pub enum Prio3Config {
     Count,
 
     /// The sum of 64-bit, unsigned integers. Each measurement is an integer in range `[0,
-    /// 2^bits)`.
-    Sum { bits: usize },
+    /// max_measurement]`.
+    Sum { max_measurement: u64 },
```
cjpatton, line 137: Note that parameter change from VDAF-08 to VDAF-13.

crates/daphne-service-utils/src/test_route_types.rs
```diff
@@ -30,34 +30,44 @@ pub struct InternalTestVdaf {
...
 
 impl From<VdafConfig> for InternalTestVdaf {
     fn from(vdaf: VdafConfig) -> Self {
-        let (typ, bits, length, chunk_length) = match vdaf {
-            VdafConfig::Prio3Draft09(prio3) => match prio3 {
-                Prio3Config::Count => ("Prio3Draft09Count", None, None, None),
-                Prio3Config::Sum { bits } => ("Prio3Draft09Sum", Some(bits), None, None),
+        let (typ, bits, length, chunk_length, dimension) = match vdaf {
+            VdafConfig::Prio3(prio3) => match prio3 {
+                Prio3Config::Count => ("Prio3ount", None, None, None, None),
```
jhoyla, line 41:
```suggestion
                Prio3Config::Count => ("Prio3Count", None, None, None, None),
```

mendess (approved):
This makes sense to me, in conjunction with the stacked PRs. Closing mine

mendess:
Aren't some of these variants also unecessary?
https://github.com/cloudflare/daphne/blob/main/crates/daphne/src/vdaf/mod.rs#L233-L245
https://github.com/cloudflare/daphne/blob/main/crates/daphne/src/vdaf/mod.rs#L275-L286

cjpatton:
> Aren't some of these variants also unecessary? https://github.com/cloudflare/daphne/blob/main/crates/daphne/src/vdaf/mod.rs#L233-L245 https://github.com/cloudflare/daphne/blob/main/crates/daphne/src/vdaf/mod.rs#L275-L286

Yes indeed, nice catch :)
* Draft09Field64
* Draft09Field128

PR #750, opened 2024-12-20
Unify VDAF dispatching logic

cjpatton:
Partially addresses #698.
Stacked on #749.

Add dispatchers for sharding, preparation, and unsharding to `VdafConfig`. Pass the DAP version to the dispatcher so that we can properly resolve supported VDAFs.

Remove support for the following, unspecified combinations:

* DAP-13/Pine32HmacSha256Aes128
* DAP-13/Pine64HmacSha256Aes128
* DAP-09/Mastic

Technically, since Mastic is still a shim at the moment and not a functional VDAF, we could support it in DAP-09 as well. When the implementation is complete (https://github.com/divviup/libprio-rs/issues/947), it will only be VDAF-13-compatible and thus only supported in DAP-13.

### DIFF
--- a/crates/dapf/src/acceptance/mod.rs
+++ b/crates/dapf/src/acceptance/mod.rs
@@ -532,7 +532,6 @@ impl Test {
             agg_job_state,
             agg_job_resp,
             self.metrics(),
-            task_config.version,
         )?;
 
         let aggregated_report_count = agg_share_span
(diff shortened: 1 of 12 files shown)


mendess (approved):
This is a nice change

cjpatton:
Rebased.

PR #751, opened 2024-12-20
vdaf: Isolate `prio_draft09` usage as much as possible

cjpatton:
Partially addresses #698.
Stacked on #750.

Move some methods from the base of the module to where they're used. This reduces the number of trait collisions there.

### DIFF
--- a/crates/daphne/src/vdaf/draft09.rs
+++ b/crates/daphne/src/vdaf/draft09.rs
@@ -1,7 +1,7 @@
-// Copyright (c) 2022 Cloudflare, Inc. All rights reserved.
+// Copyright (c) 2024 Cloudflare, Inc. All rights reserved.
 // SPDX-License-Identifier: BSD-3-Clause
 
-//! Parameters for the [Prio3 VDAF](https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/09/).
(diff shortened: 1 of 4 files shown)


crates/daphne/src/vdaf/prio3_draft09.rs
```diff
@@ -51,6 +51,28 @@ type Prio3Draft09Prepared<T, const SEED_SIZE: usize> = (
     Prio3PrepareShare<<T as Type>::Field, SEED_SIZE>,
 );
 
+pub(crate) fn shard_then_encode_draft09<V: Vdaf + Client<16>>(
```
mendess, line 54: having `draft09` in this function name is a bit redundant, the module already has draft09 in it's name

mendess: This is also true for the other functions in this module

mendess (approved):
LGTM but left a nit pick

crates/daphne/src/vdaf/prio3_draft09.rs
cjpatton: Done. I also renamed the module to `draft09` since it's also used by Pine.

cjpatton:
Rebased.

PR #752, opened 2024-12-23
Factor aggregation job handling to seperate state machine

mendess:
By having the logic to handle an aggregation job live in a state machine like this instead of having just a top level function that does it all, it's easier to give users of the library ways to customize where and when the y want to execute each part necessary in the aggregation of report shares, without compromising on the correctness of these.

This PR only transforms `helper::handle_agg_job_req` into a state machine, but the other top level functions could also be transformed in similar ways. Future PRs will use this to eventually implement the async handling of aggregation jobs


crates/daphne/src/roles/helper/handle_agg_job.rs
```diff
@@ -0,0 +1,321 @@
...
+        replay_protection: crate::ReplayProtection,
+    ) -> Result<HandleAggJob<InitializedReports>, DapError> {
+        let WithTaskConfig {
+            task_config,
+            request,
+        } = self.state;
+        let task_id = request.task_id;
+        let part_batch_sel = request.payload.part_batch_sel.clone();
+        let initialized_reports = task_config.consume_agg_job_req(
+            &aggregator.get_receiver_configs(task_config.version).await?,
```
cjpatton, line 122: This would make the code a little cleaer
```suggestion
            &aggregator.get_hpke_receiver_configs(task_config.version).await?,
```

crates/daphne/src/roles/helper/handle_agg_job.rs
```diff
@@ -0,0 +1,321 @@
...
+            },
+        })
+    }
+
+    /// Splits the state machine into it's component parts, giving the caller more control over how
+    /// the initialization of reports should be done.
+    ///
+    /// To return to the [`HandleAggJob`] state machine flow, call
+    /// [`ToInitializedReportsTransition::with_initialized_reports`].
+    pub fn into_parts(
```
cjpatton, line 144: "parts" is a little vague
```suggestion
    pub fn into_initialized_reports_and_request(
```

cjpatton (approved):
Beautiful!

crates/daphne/src/roles/helper/handle_agg_job.rs
mendess: `into_initialized_reports` is misleading, this doesn't return initialized_reports, it returns the types needed for report initialization. I preferred to go with a shorter "into_parts" like the [the std lib does](https://doc.rust-lang.org/std/index.html?search=_parts) since use of this method requires reading the documentation anyway.

PR #753, opened 2024-12-31
Make CapnprotoPayload{En,De}code composable

mendess:
Stacked on #747


crates/daphne/src/pine/vdaf.rs
```diff
@@ -981,14 +988,14 @@ mod tests {
         };
 
         let got = PinePrepState::get_decoded_with_param(
-            &(&pine, true),
+            &(&pine, DapAggregatorRole::Leader),
```
cjpatton, line 991: Same comment as https://github.com/cloudflare/daphne/pull/747#discussion_r1900416785

cjpatton (approved):

PR #754, opened 2024-12-31
Add support for public extensions in Reports.

jhoyla:
Draft 13 adds mandatory support for public extensions in the ReportMetadata, but doesn't define any (Taskprov is ([for now](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov/pull/98/files)) a private extension). This PR adds support for rejecting unknown extensions, which is all of them.

### DIFF
--- a/crates/daphne-server/tests/e2e/e2e.rs
+++ b/crates/daphne-server/tests/e2e/e2e.rs
@@ -304,6 +304,10 @@ async fn leader_upload(version: DapVersion) {
                     report_metadata: ReportMetadata {
                         id: ReportId([1; 16]),
                         time: t.now,
+                        public_extensions: match version {
+                            DapVersion::Draft09 => None,
+                            DapVersion::Latest => Some(Vec::new()),
+                        },
                     },
                     public_share: b"public share".to_vec(),
                     encrypted_input_shares: [
@@ -424,6 +428,10 @@ async fn leader_upload_taskprov() {
             t.now,
             &task_id,
             DapMeasurement::U32Vec(vec![1; 10]),
+            match version {
+                DapVersion::Draft09 => None,
+                DapVersion::Latest => Some(vec![]),
+            },
             vec![Extension::Taskprov],
             version,
         )
@@ -451,6 +459,10 @@ async fn leader_upload_taskprov() {
             t.now,
             &task_id,
             DapMeasurement::U32Vec(vec![1; 10]),
+            match version {
+                DapVersion::Draft09 => None,
+                DapVersion::Latest => Some(vec![]),
+            },
             vec![Extension::Taskprov],
             version,
         )
@@ -516,6 +528,10 @@ async fn leader_upload_taskprov_wrong_version(version: DapVersion) {
             t.now,
             &task_id,
             DapMeasurement::U32Vec(vec![1; 10]),
+            match version {
+                DapVersion::Draft09 => None,
+                DapVersion::Latest => Some(vec![]),
+            },
             vec![Extension::Taskprov],
             version,
         )
@@ -540,6 +556,100 @@ async fn leader_upload_taskprov_wrong_version(version: DapVersion) {
 }
 
 async_test_versions!(leader_upload_taskprov_wrong_version);
+
+#[tokio::test]
+async fn leader_upload_taksprov_public_errors() {
+    let version = DapVersion::Latest;
+    let t = TestRunner::default_with_version(version).await;
+    let client = t.http_client();
+    let hpke_config_list = t.get_hpke_configs(version, client).await.unwrap();
+
+    let (task_config, task_id, taskprov_advertisement) = DapTaskParameters {
+        version,
+        min_batch_size: 10,
+        query: DapBatchMode::TimeInterval,
+        leader_url: t.task_config.leader_url.clone(),
+        helper_url: t.task_config.helper_url.clone(),
+        ..Default::default()
+    }
+    .to_config_with_taskprov(
+        b"cool task".to_vec(),
+        t.now,
+        daphne::roles::aggregator::TaskprovConfig {
+            hpke_collector_config: &t.taskprov_collector_hpke_receiver.config,
+            vdaf_verify_key_init: &t.taskprov_vdaf_verify_key_init,
+        },
+    )
+    .unwrap();
+
+    // Repeated public extension
+    let report = task_config
+        .vdaf
+        .produce_report_with_extensions(
+            &hpke_config_list,
+            t.now + 1,
+            &task_id,
+            DapMeasurement::U32Vec(vec![1; 10]),
+            Some(vec![Extension::Taskprov, Extension::Taskprov]),
+            vec![],
+            version,
+        )
+        .unwrap();
+    t.leader_request_expect_abort(
+        client,
+        None,
+        &format!("tasks/{}/reports", task_id.to_base64url()),
+        &http::Method::POST,
+        DapMediaType::Report,
+        Some(
+            &taskprov_advertisement
+                .serialize_to_header_value(version)
+                .unwrap(),
+        ),
+        report.get_encoded_with_param(&version).unwrap(),
+        400,
+        "invalidMessage",
+    )
+    .await
+    .unwrap();
+
+    // Unsupported public extension
+    let report = task_config
+        .vdaf
+        .produce_report_with_extensions(
+            &hpke_config_list,
+            t.now + 1,
+            &task_id,
+            DapMeasurement::U32Vec(vec![1; 10]),
+            Some(vec![
+                Extension::Taskprov,
+                Extension::NotImplemented {
+                    typ: 3,
+                    payload: b"ignore".to_vec(),
+                },
+            ]),
+            vec![],
+            version,
+        )
+        .unwrap();
+    t.leader_request_expect_abort(
+        client,
+        None,
+        &format!("tasks/{}/reports", task_id.to_base64url()),
+        &http::Method::POST,
+        DapMediaType::Report,
+        Some(
+            &taskprov_advertisement
+                .serialize_to_header_value(version)
+                .unwrap(),
+        ),
+        report.get_encoded_with_param(&version).unwrap(),
+        400,
+        "unsupportedExtension",
+    )
+    .await
+    .unwrap();
+}
 
 async fn internal_leader_process(version: DapVersion) {
     let t = TestRunner::default_with_version(version).await;
@@ -1408,6 +1518,10 @@ async fn leader_collect_taskprov_ok(version: DapVersion) {
                     now,
                     &task_id,
                     DapMeasurement::U32Vec(vec![1; 10]),
+                    match version {
+                        DapVersion::Draft09 => None,
+                        DapVersion::Latest => Some(vec![]),
+                    },
                     extensions,
                     version,
                 )
--- a/crates/daphne-service-utils/src/compute_offload/compute_offload.capnp
+++ b/crates/daphne-service-utils/src/compute_offload/compute_offload.capnp
@@ -53,10 +53,24 @@ struct PartialDapTaskConfig @0xdcc9bf18fc62d406 {
     vdafVerifyKey       @4  :VdafVerifyKey;
 }
 
+struct PublicExtensionsList @0x8b3c98c0ddd0043e {
+
+    union {
+	# Each extension is encoded according to the DAP spec in
+	# tag-length-value form.
+        list @0 :List(Data);
+
+	# draft09 compatibility: Previously DAP had no extensions in the
+	# report.
+	none @1 :Void;
+    }
+}
+
 struct ReportMetadata @0xefba178ad4584bc4 {
 
-    id   @0 :Base.ReportId;
-    time @1 :Base.Time;
+    id               @0 :Base.ReportId;
+    time             @1 :Base.Time;
+    publicExtensions @2 :PublicExtensionsList;
 }
 
 struct PrepareInit @0x8192568cb3d03f59 {
--- a/crates/daphne/src/error/aborts.rs
+++ b/crates/daphne/src/error/aborts.rs
@@ -95,6 +95,10 @@ pub enum DapAbort {
     /// Unrecognized DAP task. Sent in response to a request indicating an unrecognized task ID.
     #[error("unrecognizedTask")]
     UnrecognizedTask { task_id: TaskId },
+
+    /// Unsupported Extension. Sent in response to a report upload with an unsupported extension.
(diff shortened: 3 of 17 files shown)


cjpatton (commented):
I believe we're missing one feature: we need to resolve repeated extensions across the public and private extension fields.

jhoyla:
> I believe we're missing one feature: we need to resolve repeated extensions across the public and private extension fields.

So the spec says:
```textCheck if any two public or private report extensions have the same extension
type. (For the purposes of this check, a private report extension can conflict
with a public report extension.) If so, the Aggregator MUST mark the input share
as invalid with error invalid_message.
```
Which I take to mean a conflict between the public and private extensions is allowed.
We also don't have any public extensions defined, presumably the text in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov/pull/98 will only be included in draft-14?

mendess (commented):
not even a nit, just a comment. I don't have a preference but instead of `Vec::new()` you can write `vec![]`

cjpatton:
> > I believe we're missing one feature: we need to resolve repeated extensions across the public and private extension fields.
> 
> So the spec says:
> 
> ```
> type. (For the purposes of this check, a private report extension can conflict
> with a public report extension.) If so, the Aggregator MUST mark the input share
> as invalid with error invalid_message.
> ```
> 
> Which I take to mean a conflict between the public and private extensions is allowed. If I add support for taskprov as a public extension do I reject it if it appears in both places?

The intent of the text is: "if the same extension appears twice, then the report MUST be rejected - regardless of what fields the extensions appear in." In particular, if the taskprov extension appears in the public and private fields, then reject.

mendess (approved):

crates/daphne/src/protocol/mod.rs
```diff
@@ -784,6 +784,56 @@ mod test {
...
+        assert_eq!(initialized_report.metadata(), &report_metadata);
+        assert_matches!(
+            initialized_report,
+            InitializedReport::Rejected {
+                report_err: ReportError::InvalidMessage,
+                ..
+            }
+        );
+    }
+    test_versions! {handle_unknown_public_extensions_in_report}
```
cjpatton, line 835: Note: This creates the test for all versions of DAP, including draft 09.

crates/daphne/src/roles/leader/mod.rs
```diff
@@ -223,6 +223,47 @@ pub async fn handle_upload_req<A: DapLeader>(
         .into());
     }
 
+    match (
+        &report.report_metadata.public_extensions,
+        task_config.version,
+    ) {
+        (Some(extensions), DapVersion::Latest) => {
+            let mut unknown_extensions = Vec::<u16>::new();
+            if crate::protocol::no_duplicates(extensions.iter()).is_err() {
```
cjpatton, line 232: And `use crate::protocol;`. Also, we should rename `no_duplicates` to something more distinctive.

```suggestion
            if protocol::no_duplicates(extensions.iter()).is_err() {
```

crates/daphne/src/roles/leader/mod.rs
```diff
@@ -223,6 +223,47 @@ pub async fn handle_upload_req<A: DapLeader>(
         .into());
     }
```
cjpatton, line 225: Add a comment here telling the reader what's going on.

crates/daphne/src/roles/leader/mod.rs
```diff
@@ -223,6 +223,47 @@ pub async fn handle_upload_req<A: DapLeader>(
...
+                DapVersion::Draft09,
+                DapVersion::Latest,
+            )))
+        }
+        (None, DapVersion::Latest) => {
+            return Err(DapError::Abort(DapAbort::version_mismatch(
+                DapVersion::Latest,
+                DapVersion::Draft09,
+            )))
+        }
```
cjpatton, line 264: I'm not sure either of these arms could ever be reached, assuming the code that sets `public_extensions` is correct. If I were you, I would use the `public_extension.is_some()` to deduce which version was negotiated. Tnen we could replace this big ugly `match` with:

```rust
if let Some(public_extensions) = &report.report_metadata.public_extensions {
    ...  // check for duplicates, unorecgonized values, and so on
}
```

crates/daphne-server/tests/e2e/e2e.rs
```diff
@@ -1348,6 +1496,116 @@ async fn leader_selected() {
     .unwrap();
 }
 
+#[tokio::test]
+async fn leader_collect_taskprov_repeated_abort() {
```
cjpatton, line 1500: What does this new test do?

cjpatton (changes requested):
The code looks largely correct, though I have some suggestions for behavior changes. I think the tests need a bit of work. Also, I have some suggestions for making the code a little cleaner/easier to read.

crates/daphne/src/error/aborts.rs
```diff
@@ -259,6 +264,20 @@ impl DapAbort {
         })
     }
 
+    pub fn unsupported_extension(
+        task_id: &TaskId,
+        unknown_extensions: &[u16],
+    ) -> Result<Self, DapError> {
+        let detail = serde_json::to_string(&unknown_extensions);
```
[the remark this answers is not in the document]
mendess: Why use json here? We don't expect anyone to parse the detail, you could just use
```suggestion
        let detail = format!("{unknown_extensions:?}");
```
which I think will output exactly the same thing in this case

crates/daphne-server/tests/e2e/e2e.rs
jhoyla: It checks that the Leader aborts if the `taskprov` extension appears in both the public and private extensions.

jhoyla: I fixed a bug in it too.

crates/daphne-server/tests/e2e/e2e.rs
```diff
@@ -533,6 +537,150 @@ async fn leader_upload_taskprov_wrong_version(version: DapVersion) {
 
 async_test_versions!(leader_upload_taskprov_wrong_version);
 
+#[tokio::test]
```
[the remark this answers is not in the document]
cjpatton: Can we move these tests to the `daphne` crate?

crates/daphne-server/tests/e2e/e2e.rs
cjpatton: Can we move this to the unit tests in the `daphne` crate. or does exercise a behavior that's not implemented by `InMemoryAggregator`?

crates/daphne/src/messages/mod.rs
```diff
@@ -240,28 +240,41 @@ impl ParameterizedDecode<DapVersion> for Extension {
...
         bytes: &mut Vec<u8>,
     ) -> Result<(), CodecError> {
         self.id.encode(bytes)?;
         self.time.encode(bytes)?;
+        match (version, &self.public_extensions) {
+            (DapVersion::Draft09, None) => (),
+            (DapVersion::Latest, Some(extensions)) => {
+                encode_u16_items(bytes, version, extensions.as_slice())?;
+            }
+            _ => return Err(CodecError::UnexpectedValue),
```
cjpatton, line 259: We usually use this for encoding. How about his:
```suggestion
            _ => return Err(CodecError::Other("encountered incorrectly set public extensions".into()))s
```

crates/daphne/src/protocol/client.rs
```diff
@@ -65,17 +68,28 @@ impl VdafConfig {
...
-        extensions: Vec<Extension>,
+        public_extensions: Option<Vec<Extension>>,
+        private_extensions: Vec<Extension>,
         version: DapVersion,
     ) -> Result<Report, DapError> {
+        match (&public_extensions, version) {
+            (Some(_), DapVersion::Draft09) | (None, DapVersion::Latest) => {
+                return Err(DapError::ReportError(
+                    crate::messages::ReportError::InvalidMessage,
+                ))
```
cjpatton, line 79: I would actually go with a fatal error here
```suggestion
                return Err(fatal_error!(
                    err = format!("public extensions not set correctly for {version:?}")
                ));
```

crates/daphne/src/protocol/client.rs
```diff
@@ -65,17 +68,28 @@ impl VdafConfig {
...
+            }
+            _ => (),
```
cjpatton, line 81: nit: you can avoid having to specify this arm using `if let`:F

```rust
if let (Some(_), DapVersion::Draft09) | (None, DapVersion::Latest) =
            (&public_extensions, version)
{
    ...
}

```

crates/daphne/src/roles/leader/mod.rs
```diff
@@ -223,6 +226,34 @@ pub async fn handle_upload_req<A: DapLeader>(
         .into());
     }
 
+    if let Some(public_extensions) = &report.report_metadata.public_extensions {
+        // We can be sure at this point that the ReportMetadata is well formed (as
+        // the decoding / error checking happens in the extractor).
```
cjpatton, line 231:
```suggestion
        // the decoding / error checking happens in the extractor.
```

crates/daphne/src/roles/leader/mod.rs
```diff
@@ -223,6 +226,34 @@ pub async fn handle_upload_req<A: DapLeader>(
...
+        assert_eq!(DapVersion::Latest, task_config.version);
+        let mut unknown_extensions = Vec::<u16>::new();
+        if protocol::check_no_duplicates(public_extensions.iter()).is_err() {
+            return Err(DapError::Abort(DapAbort::InvalidMessage {
+                detail: "Repeated public extension".into(),
+                task_id,
+            }));
+        };
```
cjpatton, line 239: IF the metdata "well-formed" (see above comment), then do we need to check for repeats here?

cjpatton (commented):
Looks pretty good. Let's chat in our 1:1 about the necessity of the e2e tests.

crates/daphne-server/tests/e2e/e2e.rs
```diff
@@ -533,6 +549,150 @@ async fn leader_upload_taskprov_wrong_version(version: DapVersion) {
...
+        .vdaf
+        .produce_report(
+            &hpke_config_list,
+            t.now + 1,
+            &task_id,
+            DapMeasurement::U32Vec(vec![1; 10]),
+            version,
+        )
+        .unwrap();
+    report.report_metadata.public_extensions = Some(vec![Extension::Taskprov, Extension::Taskprov]);
```
cjpatton, line 640: Overwriting exetensions causes an HPKE decryption failure, which isn't what we intend to test here. Modify `produce_report()` to take in public extensions and set them as you've set them here.

crates/daphne-server/tests/e2e/e2e.rs
```diff
@@ -533,6 +549,150 @@ async fn leader_upload_taskprov_wrong_version(version: DapVersion) {
...
+            version,
+        )
+        .unwrap();
+    report.report_metadata.public_extensions = Some(vec![
+        Extension::Taskprov,
+        Extension::NotImplemented {
+            typ: 3,
+            payload: b"ignore".to_vec(),
+        },
+    ]);
```
cjpatton, line 676: Same thing here: don't override

crates/daphne-server/tests/e2e/e2e.rs
```diff
@@ -533,6 +549,150 @@ async fn leader_upload_taskprov_wrong_version(version: DapVersion) {
 
 async_test_versions!(leader_upload_taskprov_wrong_version);
 
+#[tokio::test]
+async fn leader_upload_taskprov_public() {
```
cjpatton, line 553: Remove this test (already covered by unit tests).

crates/daphne-server/tests/e2e/e2e.rs
```diff
@@ -1348,6 +1508,116 @@ async fn leader_selected() {
     .unwrap();
 }
 
+#[tokio::test]
+async fn leader_collect_taskprov_repeated_abort() {
```
cjpatton, line 1512: Make sure there's a unit test for this and delete.

crates/daphne/src/roles/mod.rs
```diff
@@ -773,6 +773,60 @@ mod test {
...
+    async fn handle_unknown_public_extensions() {
+        let version = DapVersion::Latest;
+        let t = Test::new(version);
+        let task_id = &t.time_interval_task_id;
+        let task_config = t.leader.unchecked_get_task_config(task_id).await;
+        let mut report = t.gen_test_report(task_id).await;
+        report.report_metadata.public_extensions = Some(vec![Extension::NotImplemented {
+            typ: 0x01,
+            payload: vec![0x01],
+        }]);
```
cjpatton, line 786: Don't override, as this causes hpke decryption to fail. Here and below.

crates/daphne/src/roles/leader/mod.rs
jhoyla: Well formed in the sense that if the version is `DapVersion::Latest` then `public_extensions.is_some()`, and that if the version is `DapVersion::Draft09` then `public_extensions.is_none()`. It doesn't say the data is good, just that it's correctly formed.

crates/daphne/src/roles/mod.rs
jhoyla: I'm leaving the one below in place. The test below checks that a malformed report is appropriately handled. Generating the malformed report with the correct HPKE data would require a lot more code, and given that in this test the HPKE data is never checked this is much less brittle.

cjpatton (approved):

cjpatton:
I've rebased this PR and updated the capnp schema to include the public extensions field. I've also added unit tests to confirm encoding/decoding works (thanks @mendess for the help there).

cjpatton:
Updated the commit to fix clippy

crates/daphne-service-utils/src/compute_offload/compute_offload.capnp
```diff
@@ -53,10 +53,24 @@ struct PartialDapTaskConfig @0xdcc9bf18fc62d406 {
     vdafVerifyKey       @4  :VdafVerifyKey;
 }
 
+struct PublicExtensionsList @0x8b3c98c0ddd0043e {
```
cjpatton, line 56: @mendess I just ran `capnp id` and pasted the output here. So this is really just a random ID that needs to be unique?

mendess (approved):
looks good to me

cjpatton:
Rebased on the PR @mendess merged, hoping that solves the `cargo doc` failure.

PR #755, opened 2025-01-01
Reject reports earlier than task_config.not_before

jhoyla:
draft-13 changes how time is described in the draft, but the only externally visible change is rejecting reports earlier than task_config.not_before. This PR implements that behaviour, but doesn't adjust our internal representation.


cjpatton (approved):

cjpatton (changes requested):
I'm start working on this and put up a new PR once I've figured out the bug.

PR #758, opened 2025-01-09
Forbid changing aggregation job parameters

mendess:
This is important for implementing the async draft later

### DIFF
--- a/crates/daphne-server/src/router/extractor.rs
+++ b/crates/daphne-server/src/router/extractor.rs
@@ -13,7 +13,7 @@ use daphne::{
     error::DapAbort,
     fatal_error,
     messages::{
-        request::{CollectionPollReq, RequestBody},
+        request::{CollectionPollReq, HashedAggregationJobReq, RequestBody},
         taskprov::TaskprovAdvertisement,
         AggregateShareReq, AggregationJobInitReq, CollectionReq, Report, TaskId,
     },
@@ -59,6 +59,17 @@ impl_decode_from_dap_http_body!(
     Report,
     CollectionReq,
 );
+
+impl DecodeFromDapHttpBody for HashedAggregationJobReq {
+    fn decode_from_http_body(bytes: Bytes, meta: &DapRequestMeta) -> Result<Self, DapAbort> {
+        let mut cursor = Cursor::new(bytes.as_ref());
+        // Check that media type matches.
+        meta.get_checked_media_type(DapMediaType::AggregationJobInitReq)?;
+        // Decode the body
+        HashedAggregationJobReq::decode_with_param(&meta.version, &mut cursor)
+            .map_err(|e| DapAbort::from_codec_error(e, meta.task_id))
+    }
+}
 
 /// Using `()` ignores the body of a request.
 impl DecodeFromDapHttpBody for CollectionPollReq {
--- a/crates/daphne-service-utils/src/durable_requests/bindings/aggregation_job_store.rs
+++ b/crates/daphne-service-utils/src/durable_requests/bindings/aggregation_job_store.rs
@@ -1,0 +1,115 @@
+// Copyright (c) 2025 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+use crate::{
+    aggregation_job_store_capnp::new_job_request,
+    capnproto::{CapnprotoPayloadDecode, CapnprotoPayloadEncode},
+    durable_requests::ObjectIdFrom,
+};
+use daphne::{
+    messages::{AggregationJobId, AggregationJobInitReq, PartialBatchSelector, TaskId},
+    DapVersion,
+};
+use serde::{Deserialize, Serialize};
+use std::{ops::Deref, slice};
+
+super::define_do_binding! {
+    const BINDING = "AGGREGATION_JOB_STORE";
+
+    enum Command {
+        NewJob = "/new-job",
+        ListJobIds = "/job-ids",
+    }
+
+    fn name((version, task_id): (DapVersion, &'n TaskId)) -> ObjectIdFrom {
+        ObjectIdFrom::Name(format!("{version}/task/{task_id}"))
+    }
+}
+
+#[derive(Debug)]
+pub struct AggregationJobReqHash(Vec<u8>);
+
+impl Deref for AggregationJobReqHash {
+    type Target = [u8];
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl From<&AggregationJobInitReq> for AggregationJobReqHash {
+    fn from(req: &AggregationJobInitReq) -> Self {
+        let AggregationJobInitReq {
+            agg_param,
+            part_batch_sel,
+            prep_inits,
+        } = req;
+
+        let mut context = ring::digest::Context::new(&ring::digest::SHA256);
+        context.update(agg_param);
+        context.update(match part_batch_sel {
+            PartialBatchSelector::TimeInterval => &[0],
+            PartialBatchSelector::LeaderSelectedByBatchId { batch_id } => batch_id.as_ref(),
+        });
+        for p in prep_inits {
+            let daphne::messages::PrepareInit {
+                report_share:
+                    daphne::messages::ReportShare {
+                        report_metadata: daphne::messages::ReportMetadata { id, time },
+                        public_share,
+                        encrypted_input_share:
+                            daphne::messages::HpkeCiphertext {
+                                config_id,
+                                enc,
+                                payload: cypher_text_payload,
+                            },
+                    },
+                payload,
+            } = p;
+
+            context.update(payload);
+            context.update(public_share);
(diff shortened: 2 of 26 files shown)


crates/daphne-service-utils/src/durable_requests/bindings/aggregation_job_store.rs
```diff
@@ -0,0 +1,117 @@
...
+            id: <_>::decode_from_reader(reader.get_id()?)?,
+            agg_job_hash: AggregationJobReqHash(reader.get_agg_job_hash()?.to_vec()),
+        })
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub enum NewJobResponse {
+    Ok,
+    IllegalJobParameters,
```
cjpatton, line 116:
```suggestion
    /// Request would change an existing aggregation job's parameters.
    IllegalJobParameters,
```

crates/daphne-worker/src/durable/aggregation_job_store.rs
```diff
@@ -0,0 +1,96 @@
...
+                "AggregationJobStore: unexpected request: method={:?}; path={:?}",
+                req.method(),
+                req.path()
+            ))),
+        }
+    }
+
+    fn should_cleanup_at(&self) -> Option<worker::ScheduledTime> {
+        todo!()
+    }
```
cjpatton, line 72: I noticed this was stubbed out and wanted to double check you didn't mean to fill it out in this PR

crates/daphne/src/roles/helper/mod.rs
```diff
@@ -11,17 +11,25 @@ use crate::{
...
     protocol::aggregator::ReplayProtection,
-    DapAggregationParam, DapError, DapRequest, DapResponse, DapTaskConfig,
+    DapAggregationParam, DapError, DapRequest, DapResponse, DapTaskConfig, DapVersion,
 };
 
 /// DAP Helper functionality.
 #[async_trait]
-pub trait DapHelper: DapAggregator {}
+pub trait DapHelper: DapAggregator {
+    async fn assert_agg_job_is_imutable(
```
cjpatton, line 25: Needs documentation.

crates/daphne/src/testing/mod.rs
```diff
@@ -516,6 +516,9 @@ pub struct InMemoryAggregator {
     // Leader: Reference to peer. Used to simulate HTTP requests from Leader to Helper, i.e.,
     // implement `DapLeader::send_http_post()` for `InMemoryAggregator`. Not set by the Helper.
     peer: Option<Arc<InMemoryAggregator>>,
+
+    // Helper: AggregationJobs
```
cjpatton, line 520:
```suggestion
    // Helper: aggregation jobs
```

cjpatton (commented):
Looks pretty good, just one blocking comment: Is it feasible to hash the AggregationJobInitReq before decoding?

crates/daphne-service-utils/src/durable_requests/bindings/aggregation_job_store.rs
```diff
@@ -0,0 +1,117 @@
...
+
+impl Deref for AggregationJobReqHash {
+    type Target = [u8];
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl From<&AggregationJobInitReq> for AggregationJobReqHash {
+    fn from(req: &AggregationJobInitReq) -> Self {
```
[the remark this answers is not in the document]
mendess: There is no decoding going on here. The goal is to send the hash and not the object itself

crates/daphne-service-utils/src/durable_requests/bindings/aggregation_job_store.rs
```diff
@@ -0,0 +1,117 @@
...
+    enum Command {
+        NewJob = "/new-job",
+        ListJobIds = "/job-ids",
+    }
+
+    fn name(
+        (version, task_id, batch_selector):
+        (DapVersion, &'n TaskId, &'n PartialBatchSelector)
+    ) -> ObjectIdFrom {
+        ObjectIdFrom::Name(format!("{version}/task/{task_id}/selector/{batch_selector}"))
```
cjpatton, line 28: Should the batch selector be part of the name? This is one of the parameters of the aggreagtion job that should be illegal to change. If the Leader changes it, then we won't notice, right?

crates/daphne-service-utils/src/durable_requests/bindings/aggregation_job_store.rs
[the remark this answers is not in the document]
mendess: if new fields are introduced the hashing will already change. Why is the hash changing when the encoding changes better for security?

crates/daphne-service-utils/src/durable_requests/bindings/aggregation_job_store.rs
mendess: You're right! Removed it

crates/daphne-service-utils/src/durable_requests/bindings/aggregation_job_store.rs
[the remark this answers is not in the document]
cjpatton: I don't think this is immediately a problem, and it may in fact never crop up of for Daphne. However in general the principle is to hash-then-decode, not decode-then-hash. There's a long history of attacks that exploit the latter.

Here's a concrete hypothetical. Suppose we have to implement a future draft of DAP that adds a field to `AggregationJobInitReq`. If we fail to update this bit of code, then we'll fail to include the new parameter, which the Leader may be able to exploit to re-run an aggregation job with a different parameter. Hashing the entire request, regardless of its format, is more future-proof.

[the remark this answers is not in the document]
mendess: I see, I was wary of that hypothetical and that's why I always destructure the values in order to get compiler errors if fields ever change, forcing whoever is changing it to update this code as well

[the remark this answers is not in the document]
cjpatton: I see. I still don't think it's the safest code. The encoding of the fields may change, even if the fields themselves don't, and that wouldn't be detected by the compiler.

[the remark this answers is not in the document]
cjpatton: I have a compromise. We can add a test vector to ensure this doesn't change on us.

```rust
impl AggregationJobInitReq {
   fn param_hash(&self) -> [u8; 32] {
       ...
   }
}

#[test]
fn agg_job_param_hash() {
   let encoded_agg_job_init_req = [ ... ];
   let expected_param_hash = [ ... ];
   let agg_job_init_req = AggregationJobInitReq::get_decoded(&encoded_agg_job_init_req()).unwrap():
   aserrt_eq!(agg_job_init_req.param_hash(), expected_param_hash);
}
```

The reason we include the encoded agg job req in the test vector is that this test will let us know if the encoding changes in a way that is relevant to the param hash.

cjpatton (approved):
Other than the decode-then-hash issue, I think this is good to go. I have one more suggestion there.

crates/daphne-server/src/router/extractor.rs
```diff
@@ -60,6 +60,17 @@ impl_decode_from_dap_http_body!(
...
+impl DecodeFromDapHttpBody for HashedAggregationJobReq {
+    fn decode_from_http_body(bytes: Bytes, meta: &DapRequestMeta) -> Result<Self, DapAbort> {
+        let mut cursor = Cursor::new(bytes.as_ref());
+        // Check that media type matches.
+        meta.get_checked_media_type(DapMediaType::AggregationJobInitReq)?;
+        // Decode the body
+        HashedAggregationJobReq::decode_with_param(&meta.version, &mut cursor)
+            .map_err(|e| DapAbort::from_codec_error(e, meta.task_id))
+    }
+}
```
cjpatton, line 72: Hashing with `HashedAggregationJobReq::decode_with_param()` is a little awkward because you have to keep track of the start and end of the message in the cursor. Also, and this is very nit picky: `HashedAggregationJobReq` is not a DAP protocol message, so arguably doesn't belong in the `messages` module.

I think I would have done this differently:

impl DecodeFromDapHttpBody for HashedAggregationJobReq {
    fn decode_from_http_body(bytes: Bytes, meta: &DapRequestMeta) -> Result<Self, DapAbort> {
        // Check that media type matches.
        meta.get_checked_media_type(DapMediaType::AggregationJobInitReq)?;
        // Decode the body
        let agg_job_req = AggregationJobReq::get_decoded_with_param, &meta.version, bytes.as_ref())
            .map_err(|e| DapAbort::from_codec_error(e, meta.task_id))?;
        Ok(Self{
             agg_job_req:
             hash: sha256(bytes.as_ref()),
        })
    }
}

Note that this also enforces that there is nothing left to ready in `bytes`.

PR #759, opened 2025-01-10
Add CPU offload server

mendess:


crates/daphne-server/src/roles/helper.rs
```diff
@@ -1,8 +1,23 @@
...
+impl DapHelper for crate::App {
+    async fn assert_agg_job_is_immutable(
+        &self,
+        _id: AggregationJobId,
+        _version: DapVersion,
+        _task_id: &TaskId,
+        _req: &AggregationJobRequestHash,
+    ) -> Result<(), DapError> {
+        // the server implementation can't check for this
+        Ok(())
```
cjpatton, line 21: Fail loudly. Also, why doesn't `App` use this function?
```suggestion
        unreachable!("the server implementation can't check for this")
```

crates/daphne-service-utils/src/compute_offload/compute_offload.capnp
cjpatton, line 1: Is this generated by hand? If so, it's a good idea to document this.

crates/daphne-service-utils/src/compute_offload/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+    messages::{self, HpkeCiphertext, PrepareInit, ReportMetadata, ReportShare, TaskId},
+    vdaf::{VdafConfig, VdafPrepShare, VdafPrepState, VdafVerifyKey},
+    InitializedReport, PartialDapTaskConfigForReportInit, WithPeerPrepShare,
+};
+use prio::codec::{Encode, ParameterizedDecode, ParameterizedEncode};
+use std::{borrow::Cow, ops::Range};
+
+pub struct InitializeReports<'s> {
+    pub hpke_keys: Cow<'s, [HpkeReceiverConfig]>,
+    pub valid_report_range: Range<messages::Time>,
```
cjpatton, line 28: WDYM by valid here?

crates/daphne-service-utils/src/compute_offload/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+    pub task_id: TaskId,
+    pub task_config: PartialDapTaskConfigForReportInit<'s>,
+    pub agg_param: Cow<'s, [u8]>,
```
cjpatton, line 31: FYI, Mastic is nearly done, at which point we'll have an actual use case for this field!

crates/daphne-service-utils/src/compute_offload/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+            vdaf_verify_key,
+        } = self;
+        builder.set_not_after(*not_after);
+        builder.set_method_is_taskprov(*method_is_taskprov);
+        builder.set_version((*version).into());
+        builder.set_vdaf(encode_vdaf_config(vdaf).as_slice().into());
+        {
+            let builder = builder.reborrow().init_vdaf_verify_key();
+            match vdaf_verify_key.as_ref() {
+                daphne::vdaf::VdafVerifyKey::L16(array) => {
```
cjpatton, line 217: FYI: I'm not realizing we can remove this variant. I'll add a PR to follow up on this.

crates/daphne-service-utils/src/lib.rs
```diff
@@ -32,3 +34,25 @@ mod durable_request_capnp {
         "/src/durable_requests/durable_request_capnp.rs"
     ));
 }
+
+#[cfg(feature = "durable_requests")]
+mod aggregation_job_store_capnp {
+    #![allow(dead_code)]
+    #![allow(clippy::pedantic)]
+    #![allow(clippy::needless_lifetimes)]
```
cjpatton, line 42: Can you explain why these warnings need to be suppressed? Here and below

crates/daphne-worker/src/durable/aggregation_job_store.rs
cjpatton, line 1: How do we shard this object?

crates/daphne-worker/src/durable/test_state_cleaner.rs
```diff
@@ -43,14 +43,6 @@ impl TestStateCleaner {
...
                 let durable_ref: DurableReference =
                     serde_json::from_slice(&req.bytes().await?).unwrap();
-                match durable_ref.binding.as_ref() {
-                    bindings::AggregateStore::BINDING => (),
-                    s => {
-                        let message = format!("GarbageCollector: unrecognized binding: {s}");
-                        console_error!("{}", message);
-                        return Err(int_err(message));
-                    }
-                };
```
cjpatton, line 46: Why delete this?

crates/daphne/src/hpke.rs
```diff
@@ -325,7 +325,7 @@ impl HpkeDecrypter for HpkeReceiverConfig {
 pub struct HpkeReceiverConfig {
     pub config: HpkeConfig,
     #[serde(with = "HpkePrivateKeySerde")]
-    private_key: HpkePrivateKey,
+    pub private_key: HpkePrivateKey,
```
cjpatton, line 328: Why make this pub? Who needs access?

crates/daphne/src/messages/request.rs
```diff
@@ -24,15 +25,79 @@ macro_rules! impl_req_body {
...
+
+    fn hash(bytes: &[u8]) -> Self {
+        Self(
+            ring::digest::digest(&ring::digest::SHA256, bytes)
+                .as_ref()
+                .to_vec(),
+        )
+    }
+}
+
```
cjpatton, line 45:
```suggestion

/// An [`AggregationJobInitReq`] and its hash. Used by the helper to prevent the parameters of an aggregation job from changing.
```

crates/daphne/src/messages/request.rs
```diff
@@ -24,15 +25,79 @@ macro_rules! impl_req_body {
     };
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(any(test, feature = "test-utils"), derive(deepsize::DeepSizeOf))]
+pub struct AggregationJobRequestHash(Vec<u8>);
```
cjpatton, line 30: I'd stick this and `HashedAggregationJobReq` in `daphne::roles::helper`.

crates/daphne/src/messages/request.rs
```diff
@@ -24,15 +25,79 @@ macro_rules! impl_req_body {
...
-    AggregationJobInitReq | AggregationJobId
-    AggregateShareReq     | ()
-    CollectionReq         | CollectionJobId
-    CollectionPollReq     | CollectionJobId
-    ()                    | ()
+    //  body type           | id type
+    //  --------------------| ----------------
+    Report                  | ()
+    AggregationJobInitReq   | AggregationJobId
+    HashedAggregationJobReq | AggregationJobId
```
cjpatton, line 96: If you take the previous suggestion, then also update this.

crates/daphne/src/roles/helper/handle_agg_job.rs
```diff
@@ -21,7 +21,11 @@ pub struct HandleAggJob<S> {
 }
 
 /// The initial state, the aggregation request has been received.
-pub struct Init(DapRequest<AggregationJobInitReq>);
+pub struct Init(DapRequest<HashedAggregationJobReq>);
+
+/// The aggregation job is legal. Which means that it's either new or it's parameters haven't
+/// changed since the last time we've received it.
+pub struct LegalAggregationJobReq(DapRequest<AggregationJobInitReq>);
```
cjpatton, line 28: Incidentally: we were discussing whether we should use the term "agg job init req" or just "agg job req". The difference matters if we support VDAFs with more than one round of preparation, which we currently don't, and I hope we won't have to.

However, in this particular case I think the name is fine for another reason: the parameters we don't want to change pertain to an entire aggregation job, not just the initial request for that job.

crates/daphne/src/roles/helper/mod.rs
```diff
@@ -11,21 +11,34 @@ use crate::{
...
 
 /// DAP Helper functionality.
 #[async_trait]
-pub trait DapHelper: DapAggregator {}
+pub trait DapHelper: DapAggregator {
+    /// Asserts that either:
+    /// - this the first time we see this aggregation job
+    /// - this aggregation job has been seen before and it hasn't changed since the last time we
+    ///     saw it.
+    async fn assert_agg_job_is_immutable(
```
cjpatton, line 30: This is better aligned with the terminology used elsewhere.

```suggestion
    async fn assert_agg_job_req_is_legal(
```

crates/daphne/src/roles/helper/mod.rs
```diff
@@ -11,21 +11,34 @@ use crate::{
...
+        &self,
+        id: AggregationJobId,
+        version: DapVersion,
+        task_id: &TaskId,
+        req: &AggregationJobRequestHash,
```
cjpatton, line 35: Here and elsewhere
```suggestion
        req_hash: &AggregationJobRequestHash,
```

cjpatton (approved):
Looks good to me, just some superficial comments.

crates/daphne-service-utils/src/compute_offload/mod.rs
cjpatton: https://github.com/cloudflare/daphne/pull/761

crates/daphne-server/src/roles/helper.rs
mendess: I don't want to fail loudly because this step is mandatory, it's not unreachable. But it didn't exist in the server implementation. 

Which means if we fail loudly here we break the server during the transition to the async worker

crates/daphne-service-utils/src/compute_offload/compute_offload.capnp
mendess: What documentation would be useful here?

crates/daphne-service-utils/src/compute_offload/mod.rs
mendess: Whatever we use to mean here: https://github.com/cloudflare/daphne/blob/974dd0c23092809ad7a8bd1b5715902bdd61a9d9/crates/daphne/src/roles/aggregator.rs#L134

crates/daphne-service-utils/src/lib.rs
mendess: There's not point having warnings for generated code

crates/daphne-worker/src/durable/aggregation_job_store.rs
mendess: we don't really, maybe we should? I think we can wait for benchmarks before increasing complexity.

https://github.com/cloudflare/daphne/blob/974dd0c23092809ad7a8bd1b5715902bdd61a9d9/crates/daphne-service-utils/src/durable_requests/bindings/aggregation_job_store.rs#L24-L26

crates/daphne-worker/src/durable/test_state_cleaner.rs
mendess: The durable objects runtime already tells us if a durable object binding doesn't exist when we try to use it, so why check ourselves? This just makes it so we have to manually update this match statement every time we add a DO

crates/daphne/src/hpke.rs
mendess: Serialization code needs access to it. There's also no invariant the type is maintaining that requires this to be private

crates/daphne/src/messages/request.rs
mendess: What do you mean by update this? import from `roles::helper`?

crates/daphne-service-utils/src/compute_offload/compute_offload.capnp
cjpatton: Typically you'll see a comment that says something like "This file was generated by BLAH, don't edit it", e.g.:
https://github.com/cloudflare/roughtime/blob/8b34bf166fa60d69ce45fff751d4384bafe4cee0/ecosystem.json.go#L1-L2

However if capnproto doesn't generate this for you, then don't bother.

crates/daphne-service-utils/src/compute_offload/mod.rs
cjpatton: Cool, would you mind documenting this for future reference?
```suggestion
    /// Output of `DapAggregator::valid_report_time_range()` at the start of the aggregation job.
    pub valid_report_range: Range<messages::Time>,
```

crates/daphne-worker/src/durable/aggregation_job_store.rs
cjpatton: Up to you.

crates/daphne/src/messages/request.rs
cjpatton: Yes

PR #761, opened 2025-01-13
daphne: Remove 16-byte `VdafVerifyKey` variant

cjpatton:
This length was required for Prio3 by VDAF-09. However, we only support variants of Prio3 in VDAF-09 that use the custom XOF, which has a 32-byte key.

Also, once Mastic support is fully implemented, it will also use a 32-byte key.

### DIFF
--- a/crates/daphne/src/taskprov.rs
+++ b/crates/daphne/src/taskprov.rs
@@ -63,16 +63,14 @@ pub(crate) fn extract_prk_from_verify_key_init(
     Salt::new(HKDF_SHA256, &TASKPROV_SALT).extract(verify_key_init)
 }
 
-impl VdafConfig {
-    fn expand_into_taskprov_verify_key(&self, prk: &Prk, task_id: &TaskId) -> VdafVerifyKey {
-        let mut verify_key = self.uninitialized_verify_key();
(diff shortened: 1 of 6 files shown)


crates/daphne/src/taskprov.rs
```diff
@@ -63,16 +63,14 @@ pub(crate) fn extract_prk_from_verify_key_init(
...
-    }
+fn expand_into_taskprov_verify_key(prk: &Prk, task_id: &TaskId) -> VdafVerifyKey {
+    let mut verify_key = VdafVerifyKey::uninitialized();
+    let info = [task_id.as_ref()];
+    // This expand(), and the associated fill() below can only fail if the length is wrong,
+    // and it won't be, so we unwrap().
+    let okm = prk.expand(&info, verify_key.clone()).unwrap();
+    okm.fill(verify_key.as_mut()).unwrap();
+    verify_key
 }
```
mendess, line 74: Instead of having an `uninitialized` function you could instead `impl TryFrom<&[u8]> for VdafVerifyKey` and use that to create it already initialized

mendess (changes requested):

mendess (approved):

PR #764, opened 2025-01-14
Update taskprov advertisement encoding for latest draft

cjpatton:
Partially addresses #698.


crates/daphne-server/tests/e2e/e2e.rs
```diff
@@ -478,69 +478,6 @@ async fn leader_upload_taskprov() {
...
-    t.leader_request_expect_abort(
-        client,
-        None,
-        &format!("tasks/{}/reports", task_id.to_base64url()),
-        method,
-        DapMediaType::Report,
-        Some(
-            &taskprov_advertisement
-                .serialize_to_header_value(wrong_version)
-                .unwrap(),
```
cjpatton, line 532: This test was failing here, both in the latest draft and draft 09. The reason is that `TaskprovAdvertisement` now has some fields that are only expected to be set in one version or the other. Encoding would fail if a field is set unexpectedly.

We could probably refactor to make the test work, but I'm not sure it's worth the effort. What's the motivation for this @jhoyla?

mendess (approved):

crates/daphne-server/tests/e2e/e2e.rs
jhoyla: So this was testing what happens if you encode the header with the `DapVersion::Latest` and send it to a `DapVersion::Draft09` endpoint, and vice versa. I think it just fails very early on, because the `task_id` doesn't exist, so it's not doing anything very interesting.

cjpatton: Ok, do you think we can just delete this test then?

cjpatton: I'm gonna delete the test.

cjpatton:
Rebased.

PR #768, opened 2025-01-24
DAP-13: Reject reports with timestamps before task start time

cjpatton:
Partially addresses #698.
Replaces #755.

Related changes:

* Adjust the bound checks slightly: we should accept a timestamp this
  equal to either the start time or end time.

* Check the report validity window (for replay protection) before task
  validity. This isn't strictly necessary, but it simplifies some unit
  tests.

* Update the representation of the `InternalTestAddTask` struct to
  include the task duration in DAP-13. Note that the format of this
  struct is specified by draft-dcook-ppm-dap-interop-test-design, which
  has not yet been updated for DAP-13. We may need to revise the format
  at a later date.

* Update the report interval computation in the end-to-end tests to
  avoid using a timestamp that precedes the task start time.

* Add more tracing for debugging.

### DIFF
--- a/crates/daphne-server/tests/e2e/test_runner.rs
+++ b/crates/daphne-server/tests/e2e/test_runner.rs
@@ -112,12 +112,15 @@ impl TestRunner {
             default_num_agg_span_shards: NonZeroUsize::new(1).unwrap(),
         };
 
+        let not_before = now;
+        let not_after = not_before + 604_800; // one week from now
+
         let task_config = DapTaskConfig {
             version,
             leader_url: leader_url.clone(),
             helper_url: helper_url.clone(),
-            not_before: now,
-            not_after: now + 604_800, // one week from now
+            not_before,
+            not_after,
             time_precision: TIME_PRECISION,
             min_batch_size: MIN_BATCH_SIZE,
             query: query_config.clone(),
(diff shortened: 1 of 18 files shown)


mendess (approved):

crates/daphne-server/tests/e2e/test_runner.rs
```diff
@@ -327,11 +330,16 @@ impl TestRunner {
...
         // This is a portion of the interval which is guaranteed to be a valid report time
         // provided that the interval start time is valid.
-        interval.start..interval.start + REPORT_STORAGE_MAX_FUTURE_TIME_SKEW
+        let start = std::cmp::max(interval.start, self.task_config.not_before);
+        let end = std::cmp::min(
+            start + REPORT_STORAGE_MAX_FUTURE_TIME_SKEW,
+            self.task_config.not_after,
+        );
+        start..end
     }
```
cjpatton, line 343: @jhoyla FWIW I think this might have been the bug you were chasing.

jhoyla: So I did find a bug here, if you look at #755 you'll see I messed with this. The bug I was hitting is that in the e2e tests taskprov tasks were not being correctly provisioned. If the task isn't found when the leader looks it up in storage then a default task is returned, which has the current time as the `not_before`. This change makes it less likely that you'll hit an issue because the test takes less than a second to run, but doesn't actually fix the underlying.

cjpatton: Hmm, what was failing exactly?

If the Client uploads a report for task1, but the Leader resolves task2, then the report should get rejected at some point.

It would be really helpful if you could create a PR, either with the fix, a test case that exercises the bug, or both.

PR #769, opened 2025-01-27
Add support for the Mastic VDAF

cjpatton:
Mastic (https://datatracker.ietf.org/doc/draft-mouris-cfrg-mastic/) is a
VDAF that enables a richer set of functionalities than the VDAFs we
support so so far. The `daphne::vdaf::mastic` module contains a "dummy"
version of Mastic intended to exercise the DAP protocol logic we would
need in order to fully support this VDAF. The `prio` crate now
implements Mastic, so upgrade to a version of the crate that supports it
and replace the dummy VDAF with the real one.

In addition, to complete aggregation of a report, it is necessary to
know the aggregation parameter, which currently is only plumbed to
report initialization. In particular,
`DapTaskConfig::produce_agg_job_resp()` needs the aggregation parameter
from the aggregation job request message. (Likewise,
`ToInitializedReportsTransition::with_initialized_reports()` needs the
aggregation parameter.)

Finally, clean up some API cruft in `daphne::vdaf`:

1. Encapsulate variants of Mastic behind a `MasticConfig` as we've done
   for other VDAFs.

2. Modify `prep_finish_from_shares()` to not take the aggregator ID.
   This is a relic of when we supported DAP-02, when this method may
   have been called by either the Leader or the Helper. Now it's always
   called by the Helper.

3. Implement state encoding for Mastic, as required by the new async
   Helper implementation.

4. Generalize the `prep_init()` function in `daphne::prio3` to be used
   for Mastic as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,8 @@ matchit = "0.7.3"
 p256 = { version = "0.13.2", features = ["ecdsa-core", "ecdsa", "pem"] }
 paste = "1.0.15"
 prio_draft09 = { package = "prio", version = "0.16.7" }
-prio = { git = "https://github.com/divviup/libprio-rs.git", rev = "c50bb9a47b396ad6a08a3fec36b98bcc2d9217a1" }
+# TODO Point to version `0.17.0` once release. This revision is one commit ahead of `0.17.0-alpha.0`.
+prio = { git = "https://github.com/divviup/libprio-rs.git", rev = "e5e8a47ee4567f7588d0b5c8d20f75dde4061b2f" }
 prometheus = "0.13.4"
 rand = "0.8.5"
 rayon = "1.10.0"
--- a/crates/daphne-worker/src/aggregator/router/helper.rs
+++ b/crates/daphne-worker/src/aggregator/router/helper.rs
@@ -11,6 +11,7 @@ use axum::{
     routing::{post, put},
 };
 use daphne::{
+    error::DapAbort,
     fatal_error,
     hpke::HpkeProvider,
     messages::AggregateShareReq,
@@ -18,11 +19,11 @@ use daphne::{
         helper::{self, HashedAggregationJobReq},
         DapAggregator, DapHelper,
     },
-    DapError, DapResponse,
+    DapAggregationParam, DapError, DapResponse,
 };
 use daphne_service_utils::compute_offload;
 use http::StatusCode;
-use prio::codec::ParameterizedEncode;
+use prio::codec::{ParameterizedDecode, ParameterizedEncode};
 use std::{borrow::Cow, sync::Arc};
 
 pub(super) fn add_helper_routes(router: super::Router<App>) -> super::Router<App> {
@@ -60,6 +61,12 @@ async fn agg_job(
 
         let hpke_receiver_configs = app.get_hpke_receiver_configs(req.version).await?;
 
+        let agg_param = DapAggregationParam::get_decoded_with_param(
+            &transition.task_config.vdaf,
+            &req.payload.agg_param,
+        )
+        .map_err(|e| DapAbort::from_codec_error(e, req.task_id))?;
+
         let initialized_reports: compute_offload::InitializedReports = app
             .compute_offload
             .compute(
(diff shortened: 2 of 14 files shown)


crates/daphne-worker/src/aggregator/router/helper.rs
```diff
@@ -60,6 +61,12 @@ async fn agg_job(
 
         let hpke_receiver_configs = app.get_hpke_receiver_configs(req.version).await?;
 
+        let agg_param = DapAggregationParam::get_decoded_with_param(
+            &transition.task_config.vdaf,
+            &req.payload.agg_param,
+        )
+        .map_err(|e| DapAbort::from_codec_error(e, req.task_id))?;
+
```
cjpatton, line 69: @mendess is this change required in `daphne-server` as well? AFAICT it doesn't need a custom implementation of this state machine.

Cargo.toml
```diff
@@ -63,7 +63,7 @@ matchit = "0.7.3"
 p256 = { version = "0.13.2", features = ["ecdsa-core", "ecdsa", "pem"] }
 paste = "1.0.15"
 prio_draft09 = { package = "prio", version = "0.16.7" }
-prio = { git = "https://github.com/divviup/libprio-rs.git", rev = "c50bb9a47b396ad6a08a3fec36b98bcc2d9217a1" }
+prio = { git = "https://github.com/divviup/libprio-rs.git", rev = "7e479222152cb16abc2745bf651ad93a88f8088e" }
 prometheus = "0.13.4"
```
cjpatton, line 67: Before merging this PR we need to make a change upstream: https://github.com/divviup/libprio-rs/pull/1209

crates/daphne/src/vdaf/mastic.rs
cjpatton, line 1: Make sure to turn on "hide whitespace" for reviewing this diff.

crates/daphne/src/vdaf/mastic.rs
```diff
@@ -233,19 +202,22 @@ mod test {
...
-                Poplar1AggregationParam::try_from_prefixes(vec![
-                    IdpfInput::from_bytes(b"cool"),
-                    IdpfInput::from_bytes(b"trip"),
-                ])
+                MasticAggregationParam::new(
+                    vec![
+                        IdpfInput::from_bytes(b"cool"),
+                        IdpfInput::from_bytes(b"trip"),
+                    ],
+                    true,
```
mendess, line 219: what does this `true` mean?

crates/daphne/src/roles/mod.rs
```diff
@@ -2043,11 +2045,14 @@ mod test {
...
-                IdpfInput::from_bytes(&[1]),
-                IdpfInput::from_bytes(&[7]),
-            ])
+            MasticAggregationParam::new(
+                vec![
+                    IdpfInput::from_bytes(&[0]),
+                    IdpfInput::from_bytes(&[1]),
+                    IdpfInput::from_bytes(&[7]),
+                ],
+                true,
```
mendess, line 2054: Same here

crates/daphne/src/vdaf/mastic.rs
cjpatton: It's a detail of Mastic. It tells the aggreagtors to do perform a specific validity check.

This check should be mandatory, but DAP-13 doesn't enforce it. This is a bug, which we plan to fix in the next version: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/649

This is tracked by https://github.com/cloudflare/daphne/issues/770.

crates/daphne/src/roles/mod.rs
cjpatton: As above.

crates/daphne/src/vdaf/mastic.rs
mendess: so the bool parameter should disappear in the future?

mendess: ah okay. Sure

mendess (approved):

